PR #6, opened 2021-11-09
Add .proto files and generate Go Protobufs

tigrannajaryan:
This adds .proto file that matches the current spec at
https://github.com/open-telemetry/opamp-spec

I also added a makefile with gen-proto target to generate
the go Protobufs.

I suggest that we keep the .proto and generated files in
this repo for now to keep initial development fast. Kept in
"internal" for now to avoid leaking unwanted dependency.

We can move the .proto to a separate repository later if we
decide that's the right approach (similar to OTLP proto files).


CONTRIBUTING.md
```diff
@@ -0,0 +1,9 @@
+# Contributing to opamp-go library
+
+## Generate Protobuf Go Files
+
+Make sure you have [Protobuf Compiler](https://github.com/protocolbuffers/protobuf#protocol-compiler-installation)
```
djaglowski, line 5: Since `make gen-proto` depends on `protoc-gen-gofast`, we should probably include some instruction on how to install it, since this isn't covered in this link, and isn't all that intuitive.

`> go install github.com/gogo/protobuf/protoc-gen-gofast@latest`

djaglowski (approved):
Confirmed that `make gen-proto` works

makefile
```diff
@@ -0,0 +1,20 @@
+GO=$(shell which go)
```
codeboten, line 1: nit, the first letter of the filename should be capitalized

CONTRIBUTING.md
codeboten: could the build tools be used here https://github.com/open-telemetry/build-tools/tree/main/protobuf?

tigrannajaryan: I didn't mean to depend on gogo proto, let me remove that. I think it is not necessary for now.

tigrannajaryan: Changed to use https://github.com/open-telemetry/build-tools/tree/main/protobuf

makefile
tigrannajaryan: Removed. Not needed.

makefile
```diff
@@ -0,0 +1,21 @@
...
+BASELINE_PROTO_FILES := $(wildcard internal/proto/*.proto)
+
+all:
+
+# Generate Protobuf Go files.
+.PHONY: gen-proto
+gen-proto:
+	$(foreach file,$(BASELINE_PROTO_FILES),$(call exec-command,docker run --rm -v${PWD}:${PWD} \
+        -w${PWD} otel/build-protobuf:latest --proto_path=${PWD}/internal/proto/ \
+        --go_out=${PWD}/internal/proto/ -I${PWD}/internal/proto/ ${PWD}/$(file)))
```
Aneurysm9, line 18:
```suggestion
        --go_out=${PWD}/internal/protobufs/ -I${PWD}/internal/proto/ ${PWD}/$(file)))
```

Should the output go to `protobufs` instead of `proto`?

tigrannajaryan: That's just an intermediary directory (could be anything). Everything first compiles to `internal/proto/github.com/open-telemetry/opamp-go/internal/protobufs/` and then is copied to `internal/protobufs/` on the next line.

tigrannajaryan:
@codeboten @Aneurysm9 ok to merge?

jmacd (approved):

tigrannajaryan:
OK, merging, all comments resolved.

PR #7, opened 2021-11-09
Initial client interfaces and types

tigrannajaryan:
This is a possible way to design the client API. I would like
some feedback on this or possibly an alternate API suggestion
for the client.

This one uses callbacks. I considered a channel-based approach
to notify the agent about messages but it async nature of channels
makes it difficult to return responses to the server.

Any other thoughts on a better API for the client?

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -1,0 +1,97 @@
+package client
+
+import (
+	"context"
+	"crypto/tls"
+	"time"
+
+	"github.com/open-telemetry/opamp-go/client/types"
+	"github.com/open-telemetry/opamp-go/internal/protobufs"
+)
+
+type StartSettings struct {
+	// Connection parameters.
+	OpAMPServerURL      string
+	AuthorizationHeader string
+	TLSConfig           *tls.Config
+
+	// Agent information.
+	InstanceUid  string
+	AgentType    string
+	AgentVersion string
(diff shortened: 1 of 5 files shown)


tigrannajaryan:
@open-telemetry/opamp-go-approvers please have a look and feel free to suggest a different way to structure the API. I just want to start the ball rolling.

djaglowski:
I like the callbacks interface. Aside from doing what we need in an elegant way, it provides an easy point of entry for alternate implementations of the message handling. 

A question this raises for me, is whether we should provide a mechanism for additional user-defined message handling, so that the same websocket can be used for any vendor-specific control messages. Something like:
```go
type Callbacks interface {
	...
	OnGenericMessage(messageType string, messageContent interface{})
```

This doesn't need to be a concern of this proposal - just pointing out a possibility that may be useful.

tigrannajaryan:
> A question this raises for me, is whether we should provide a mechanism for additional user-defined message handling, so that the same websocket can be used for any vendor-specific control messages.

This is a very good question and something I have been thinking about possibly adding to the spec. :-)

tigrannajaryan:
@open-telemetry/opamp-go-approvers please review.

pmm-sumo (approved):

PR #30, opened 2021-12-01
Fix location of github action

tigrannajaryan:


djaglowski (approved):

djaglowski (approved):
This looks correct to me. I think we do not expect the check run on this PR because `main` controls the trigger rules.

PR #34, opened 2021-12-06
Make protobufs and client CallbacksStruct public

pmm-sumo:
This PR exports protobufs and client callbacks

Protobufs are required to use the Callbacks interface

CallbacksStruct make it easier to initialise Client (no need to implement the whole interface, only specific functions)


tigrannajaryan (commented):
Need to also fix makefile since it contains the path to protobufs dir.

pmm-sumo:
> Need to also fix makefile since it contains the path to protobufs dir.

Good point @tigrannajaryan . Updated gen-proto section and package name

PR #39, opened 2021-12-08
Add basic Agent example

tigrannajaryan:
The example demonstrates how to use OpAMPClient to create
a simple Agent that works with OpAMP Server.


pmm-sumo (approved):
That's a very helpful example

internal/examples/agent/agent/agent.go
```diff
@@ -0,0 +1,280 @@
...
+	instanceConfig := config.Config.ConfigMap[""]
+	if instanceConfig != nil {
+		orderedConfigs = append(orderedConfigs, agentConfigFileItem{
+			name: "",
+			file: instanceConfig,
+		})
+	}
+
+	// Merge received configs.
+	for _, item := range orderedConfigs {
```
pmm-sumo, line 244: Should the agent check for content type of each file?

tigrannajaryan: Good point, but the server example currently does not set the content type so this won't work right now. We can add content type later on both server and client examples in a future PR.

PR #41, opened 2021-12-08
[WIP] Draft for supervisor interface and configuration file

kart:
This changeset takes a first stab at the supervisor model for
implementing the client side of the OpAMP protocol.

There are two parts to this changeset:

1. Proposal on the supervisor configuration file based on the draft
specification. Two alternate configuration files are provided as an
example of how the supervisor might be instantiated.

2. A simple supervisor interface that all implementations should adhere
to. This allows for the implementation of a generic supervisor as well
as tailor-made ones that has an understanding of the inner workings of
the agent that it is supervising. The supervisor internally wraps the
OpAMP client that communicates with the OpAMP server for management
instructions.

### DIFF
--- a/supervisor/conf/config-1.yml
+++ b/supervisor/conf/config-1.yml
@@ -1,0 +1,85 @@
+# Connection details for the OpAMP server.
+opampServer:
+    endpoint: https://localhost:1729
+    tls:
+        # Client-side TLS configuration to connect to the OpAMP server.
+
+# Agent specifies the agent that is being managed by this supervisor. Concretely,
+# the `agent' configuration dictates how to start it and where the configuration
+# file resides. This basic information is necessary for the supervisor to
+# restart the agent on a remote configuration change.
+# `watchdog' and `systemd' config elements are mutually exclusive. So, only
+# one of the two must be specified below.
+agent:
+    # Type specifies the type of the agent that is being supervised. Internally,
+    # the `type' can be used to auto-discover several attributes if the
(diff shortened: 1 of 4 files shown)


kart:
Hello all, while working on this PR, I had a few questions and observations that I am posting here.

1. It seems to me that it is better to have several supervisor implementations running within the same supervisor executable rather than having separate supervisor executables. In a way, this is much akin to the systemd architecture. This also keeps the management easy as the user wouldn't have to run several of them.

2. If the agent is managed by a systemd facility, the supervisor will have to be running as root to be able to stop/start the agent.

3. Should we allow multiple OS facilities like systemd, Sys V init, etc? If so the command to start, stop and restart the agents must be added to the configuration file instead of presupposing a systemd facility.

4. In the case of an OT collector, there seems to be only a single executable that must be replaced with a newer version to perform an upgrade. In other agents, this may or may not be the case. Should we allow the configuration for such agents to specify a file specification (or even a copy specification) which an arbitrary supervisor can perform? If every supervisor knows exactly about the agent that it is supervising, this is not required. Otherwise, I believe some sort of a spec is required.

5. There are several deployment options for the collector (k8s, docker, VMs, etc). The configuration file seems to be more appropriate for a docker/VM deployment. I am not sure if OpAMP is targeting k8s like deployment as well.

tigrannajaryan:
Please sign the CLA.

kart:
Closing and reopening for fixing CLA issues.

PR #42, opened 2021-12-08
[WIP] Draft for supervisor interface and configuration file

kart:
This changeset takes a first stab at the supervisor model for
implementing the client side of the OpAMP protocol.

There are two parts to this changeset:

1. Proposal on the supervisor configuration file based on the draft
specification. Two alternate configuration files are provided as an
example of how the supervisor might be instantiated.

2. A simple supervisor interface that all implementations should adhere
to. This allows for the implementation of a generic supervisor as well
as tailor-made ones that has an understanding of the inner workings of
the agent that it is supervising. The supervisor internally wraps the
OpAMP client that communicates with the OpAMP server for management
instructions.

### DIFF
--- a/supervisor/conf/config-1.yml
+++ b/supervisor/conf/config-1.yml
@@ -1,0 +1,85 @@
+# Connection details for the OpAMP server.
+opampServer:
+    endpoint: https://localhost:1729
+    tls:
+        # Client-side TLS configuration to connect to the OpAMP server.
+
+# Agent specifies the agent that is being managed by this supervisor. Concretely,
+# the `agent' configuration dictates how to start it and where the configuration
+# file resides. This basic information is necessary for the supervisor to
+# restart the agent on a remote configuration change.
+# `watchdog' and `systemd' config elements are mutually exclusive. So, only
+# one of the two must be specified below.
+agent:
+    # Type specifies the type of the agent that is being supervised. Internally,
+    # the `type' can be used to auto-discover several attributes if the
+    # supervisor is tailor-made for that agent thereby relieving the user from
+    # having to specify most, if not all, of the configuration.
+    type: io.opentelemetry.collector
+    # Fully-qualified path to the agent executable required for two reasons:
+    # 1. If the supervisor is functioning as a watchdog, then the executable
+    #    is used to start/stop/restart the agent on configuration changes, etc.
+    # 2. If the supervisor is performing the upgrade, the executable indicates
+    #    the file and path in the filesystem that must be replaced with the newer
+    #    version of the executable.
+    executable: /opt/opentelemetry-collector/bin/otel
+    #
+    management:
+        # Dictates how to check if the agent is healthy.
+        healthCheck:
+            url: http://localhost:13133/
+            initialDelay: 5s
+            period: 5s
+            timeout: 20s
+        ownTelemetry:
+            url: http://localhost:8888/
+            initialDelay: 5s
+            period: 5s
+    config:
+        # Currently, only a `file' based configuration is accepted.
+        file: /opt/opentelemetry-collector/config.yaml
+        # If `autoReload` is true, then the agent doesn't need a restart on a config
+        # change as it is internally managed by the agent.
+        autoReload: false
+    # An arbitrary set of key-value pairs that is specific for this agent. This
+    # will be used by the specific supervisor that is managing this agent.
+    attrs:
+        service.name: ecommerce
+        service.namespace: shop
+    # Set of attributes that describe this agent's systemd installation. If this
+    # config element is specified, the `watchdog' config element is ignored as
+    # they are mutually exclusive. The `systemd' information specified here will
+    # be used by the supervisor to restart this agent on configuration changes,
+    # executable upgrades, etc.
+    systemd:
+        name: otel-collector
+    # Watchdog allows the supervisor to monitor the health of the agent and ensure
+    # it is always running (restarting if necessary). In this sense, it acts as
+    # an alternative to agent installations that are not managed by systemd.
+    watchdog:
+        # Set of command-line arguments required to start the `executable` specified
+        # above.
+        args:
+            - --config
+            - /etc/otel/config.yaml
+        # Additional environment variables that must be set when the agent is
+        # started.
+        env:
+            ENV_VAR1: env_value1
+            ENV_VAR2: env_value2
+        # A watchdog policy, like `systemd', albeit not as exhaustive, that
+        # the supervisor uses to control the starting and stopping of the
+        # agents that are being supervised.
+        policy:
+            # Dictates when to restart: `always' restarts any time the supervisor
+            # finds the agent not running; `failure' restarts only if the process
+            # dies as a result of a non-zero exit code (which could be a result of
+            # an internal error or an unexpected OS signal).
+            restart: always
+            # Time to wait before restarting the agent (as configured with `restart')
+            restartWait: 10s
+            # Max attempts the agent is allowed to fail when attempted to START or
+            # STOP in which case the operation will be re-attempted.
+            maxAttempts: 3
+            # Duration to wait between successive attempts of a START or STOP.
+            waitBetweenAttempts: 1s
--- a/supervisor/conf/config-2.yml
+++ b/supervisor/conf/config-2.yml
@@ -1,0 +1,89 @@
+# Connection details for the OpAMP server.
+opampServer:
+    endpoint: https://localhost:1729
+    tls:
+        # Client-side TLS configuration to connect to the OpAMP server.
+
+agents:
+    # Many collector related configuration is omitted in this section as its
+    # tailor-made supervisor can fill in with the defaults.
+    io.opentelemetry.collector:
+        executable: /opt/opentelemetry-collector/bin/otel
+        config:
+            file: /opt/opentelemetry-collector/config.yaml
+        attrs:
+            service.name: ecommerce
+            service.namespace: shop
+        # systemd section is commented -- enable this if the collector is
+        # managed by systemd.
+        # systemd:
+        #    name: otel-collector
+        watchdog:
+            env:
+                ENV_VAR1: env_value1
+                ENV_VAR2: env_value2
+            policy:
+                restart: always
+                restartWait: 10s
+                maxAttempts: 3
+                waitBetweenAttempts: 1s
+    # Many FluentBit related configuration is omitted in this section as its
+    # tailor-made supervisor can fill in with the defaults.
+    io.fluentbit:
+        executable: /opt/fluent-bit/bin/fluent-bit
+        config:
+            file: /opt/fluent-bit/fluent-bit.conf
+            autoReload: true
+        attrs:
+            attr1_key: attr1_value
+        systemd:
+            name: fluentbit
+        # watchdog section is commented -- enable this if fluentbit is NOT
+        # managed by systemd.
+        # watchdog:
+        #     env:
+        #         ENV_VAR1: env_value1
+        #         ENV_VAR2: env_value2
+        #     policy:
+        #         restart: always
+        #         restartWait: 10s
+        #         maxAttempts: 3
+        #         waitBetweenAttempts: 1s
+    # A generic agent for which there is no already implemented supervisor.
+    # In this case, all the configuration elements must be specified for the
+    # supervisor to act correctly.
+    foobar:
+        executable: /opt/foobar/bin/foobar
+        config:
+            file: /opt/foobar/config.yaml
+            autoReload: true
+        attrs:
+             attr1_key: attr1_value
+        management:
+            healthCheck:
+                url: http://localhost:12222/ping
+                initialDelay: 5s
+                period: 5s
+                timeout: 20s
+            ownTelemetry:
+                url: http://localhost:12223/metrics
+                initialDelay: 5s
+                period: 5s
+        systemd:
+            name: foobar-service
+        watchdog:
+            args:
+                - --config
+                - /opt/foobar/config.yaml
+                - --arg1
+                - arg1_value
+                - --arg2
+                - arg2_value
+            env:
+                ENV_VAR1: env_value1
+                ENV_VAR2: env_value2
+            policy:
+                restart: always
+                restartWait: 10s
+                maxAttempts: 3
+                waitBetweenAttempts: 1s
--- a/supervisor/supervisor.go
+++ b/supervisor/supervisor.go
@@ -1,0 +1,41 @@
+package supervisor
+
+import (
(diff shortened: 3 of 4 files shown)


kart:
Hello all, while working on this PR, I had a few questions and observations that I am posting here.

It seems to me that it is better to have several supervisor implementations running within the same supervisor executable rather than having separate supervisor executables. In a way, this is much akin to the systemd architecture. This also keeps the management easy as the user wouldn't have to run several of them.

If the agent is managed by a systemd facility, the supervisor will have to be running as root to be able to stop/start the agent.

Should we allow multiple OS facilities like systemd, Sys V init, etc? If so the command to start, stop and restart the agents must be added to the configuration file instead of presupposing a systemd facility.

In the case of an OT collector, there seems to be only a single executable that must be replaced with a newer version to perform an upgrade. In other agents, this may or may not be the case. Should we allow the configuration for such agents to specify a file specification (or even a copy specification) which an arbitrary supervisor can perform? If every supervisor knows exactly about the agent that it is supervising, this is not required. Otherwise, I believe some sort of a spec is required.

There are several deployment options for the collector (k8s, docker, VMs, etc). The configuration file seems to be more appropriate for a docker/VM deployment. I am not sure if OpAMP is targeting k8s like deployment as well.

pavankrish123:
cc: @annabokhan

supervisor/conf/config-1.yml
```diff
@@ -0,0 +1,85 @@
+# Connection details for the OpAMP server.
+opampServer:
```
tigrannajaryan, line 2: Let's use snake_case for config settings. That's what Otel Collector uses.

supervisor/conf/config-1.yml
```diff
@@ -0,0 +1,85 @@
...
+    endpoint: https://localhost:1729
```
tigrannajaryan, line 3: OpAMP is WebSocket, so needs `wss:` scheme:
```suggestion
    endpoint: wss://localhost:1729
```

supervisor/conf/config-1.yml
```diff
@@ -0,0 +1,85 @@
...
+# restart the agent on a remote configuration change.
+# `watchdog' and `systemd' config elements are mutually exclusive. So, only
+# one of the two must be specified below.
+agent:
+    # Type specifies the type of the agent that is being supervised. Internally,
+    # the `type' can be used to auto-discover several attributes if the
+    # supervisor is tailor-made for that agent thereby relieving the user from
+    # having to specify most, if not all, of the configuration.
+    type: io.opentelemetry.collector
+    # Fully-qualified path to the agent executable required for two reasons:
```
tigrannajaryan, line 19: Does this need to be full path? Can we allow relative too?

supervisor/conf/config-1.yml
```diff
@@ -0,0 +1,85 @@
...
+        # Dictates how to check if the agent is healthy.
+        healthCheck:
+            url: http://localhost:13133/
+            initialDelay: 5s
+            period: 5s
+            timeout: 20s
+        ownTelemetry:
+            url: http://localhost:8888/
+            initialDelay: 5s
+            period: 5s
```
tigrannajaryan, line 37: What does this do?

supervisor/conf/config-1.yml
```diff
@@ -0,0 +1,85 @@
...
+        name: otel-collector
+    # Watchdog allows the supervisor to monitor the health of the agent and ensure
+    # it is always running (restarting if necessary). In this sense, it acts as
+    # an alternative to agent installations that are not managed by systemd.
+    watchdog:
+        # Set of command-line arguments required to start the `executable` specified
+        # above.
+        args:
+            - --config
+            - /etc/otel/config.yaml
```
tigrannajaryan, line 64: This presumably should match "config.file" option above. Can we eliminate the need to specify it twice?

supervisor/conf/config-1.yml
```diff
@@ -0,0 +1,85 @@
...
+            ENV_VAR2: env_value2
+        # A watchdog policy, like `systemd', albeit not as exhaustive, that
+        # the supervisor uses to control the starting and stopping of the
+        # agents that are being supervised.
+        policy:
+            # Dictates when to restart: `always' restarts any time the supervisor
+            # finds the agent not running; `failure' restarts only if the process
+            # dies as a result of a non-zero exit code (which could be a result of
+            # an internal error or an unexpected OS signal).
+            restart: always
```
tigrannajaryan, line 78: Can we simplify further and remove this option? Do we really need "always"?

supervisor/conf/config-2.yml
```diff
@@ -0,0 +1,89 @@
+# Connection details for the OpAMP server.
+opampServer:
+    endpoint: https://localhost:1729
+    tls:
+        # Client-side TLS configuration to connect to the OpAMP server.
+
+agents:
+    # Many collector related configuration is omitted in this section as its
+    # tailor-made supervisor can fill in with the defaults.
```
tigrannajaryan, line 9: It is not clear what the intent is. Is the Supervisor supposed to supervise more than one agent simultaneously? Why is it necessary?

tigrannajaryan:
> Hello all, while working on this PR, I had a few questions and observations that I am posting here.

This is very good start, thanks for the pr.

> 
> It seems to me that it is better to have several supervisor implementations running within the same supervisor executable rather than having separate supervisor executables. In a way, this is much akin to the systemd architecture. This also keeps the management easy as the user wouldn't have to run several of them.

I think this depends on how cumbersome a generic Supervisor is compared to a specialized one. If a specialized Supervisor is much easier to use than it may warranted to have one for e.g. Collector instead of requiring the user to do complicated configuration. It is hard to tell this without seeing both approaches side-by-side. I think it is worth starting with the generic Supervisor and then see how much better we can do for Otel Collector and decide if it is worth it.

> 
> If the agent is managed by a systemd facility, the supervisor will have to be running as root to be able to stop/start the agent.

What if we make Supervisor the process that is managed by systemd and the Agent then becomes a child process which the Supervisor can start/stop independently without systemd being involved?

> 
> Should we allow multiple OS facilities like systemd, Sys V init, etc? If so the command to start, stop and restart the agents must be added to the configuration file instead of presupposing a systemd facility.

Yes, I think at least initially we should keep it simple and avoid hard-coding specialized logic for these facilities. If we can manage the Agent as a child process we may not need to do anything special.

> 
> In the case of an OT collector, there seems to be only a single executable that must be replaced with a newer version to perform an upgrade. In other agents, this may or may not be the case. Should we allow the configuration for such agents to specify a file specification (or even a copy specification) which an arbitrary supervisor can perform? If every supervisor knows exactly about the agent that it is supervising, this is not required. Otherwise, I believe some sort of a spec is required.

Yes, probably. I suggest we start with the simple case of single executable for now.

> 
> There are several deployment options for the collector (k8s, docker, VMs, etc). The configuration file seems to be more appropriate for a docker/VM deployment. I am not sure if OpAMP is targeting k8s like deployment as well.

Yes, it is likely that when deploying to k8s the users will prefer to use k8s configuration management facilities. Also, it is likely under k8s they will not want executable upgrades since k8s itself can also handle that. We need to make these capabilities optional, while retaining the rest (e.g. status and telemetry reporting).

supervisor/supervisor.go
```diff
@@ -0,0 +1,41 @@
...
+// the internal workings of the collector. Alternatively, a default supervisor
+// allows for the supervision of arbitrary daemon-like processes and provides
+// management options like remote configuration change, agent upgrades,
+// watchdog capabilities, telemetry reporting, etc.
+//
+// The supervisor wraps the client part of the OpAMP protocol and therefore
+// requires an OpAMP server to be running at some endpoint. There is a 1:1
+// mapping between a supervisor and the OpAMP client. A single supervisor
+// executable, however, can run several supervisors within it thereby managing
+// several distinct agent types.
```
tigrannajaryan, line 19: I am not sure this is necessary. What's the use case for multiple Agents per Supervisor executable?

supervisor/supervisor.go
```diff
@@ -0,0 +1,41 @@
...
+	//
+	// Start MUST return immediately (potentially with an error if the specified
+	// configuration is incorrect or it fails to parse).
+	//
+	// Start, in most cases, should spawn a Go routine to monitor the agent
+	// under supervision.
+	//
+	// Start, on a given instance, must be called exactly once. Subsequent
+	// calls have no effect.
+	Start(config *types.Configuration) error
```
tigrannajaryan, line 35: This requires a very specific way of specifying the Configuration. I am not sure what the purpose of such `Supervisor` interface is.

I was imagining a different approach when supervising capabilities are implemented as composable features which can be used to create a generic or specific Supervisor. This requires reversing the interfaces and have e.g. `SupervisableAgent` interface which can be passed to the Supervisor implementation to be managed.

annabokhan:
@tigrannajaryan 
> Yes, it is likely that when deploying to k8s the users will prefer to use k8s configuration management facilities. Also, it is likely under k8s they will not want executable upgrades since k8s itself can also handle that. We need to make these capabilities optional, while retaining the rest (e.g. status and telemetry reporting).

When you say 'users will prefer to use k8s configuration management facilities' - how would that work with OpAMP? Is the expectation that for the OpAMP-capable agents deployed in k8s the remote configuration/update capabilities are disabled, and instead these are managed by having an inbound access to the k8s cluster?

tigrannajaryan:
> When you say 'users will prefer to use k8s configuration management facilities' - how would that work with OpAMP? Is the expectation that for the OpAMP-capable agents deployed in k8s the remote configuration/update capabilities are disabled, and instead these are managed by having an inbound access to the k8s cluster?

Yes, I think that's exactly what is most likely to happen. We need to have a clear way to disable particular OpAMP capabilities while keeping the rest.

Perhaps in some rare cases the remote configuration capability may be enabled, such that the k8s ConfigMap provides only the initial "empty" configuration and OpAMP is used to fetch the real config to run, but my gut feeling is that most k8s people would prefer not to do it.

And auto-updating capabilities if I understand correctly are totally useless for k8s since any updates to executables done via OpAMP will be reverted after a routine pod restart.

annabokhan:
> Yes, I think that's exactly what is most likely to happen. We need to have a clear way to disable particular OpAMP capabilities while keeping the rest.

Considering that the ownership of a k8s cluster and OpAMP client/server can be different, that might not be a feasible option in case of a vendor-owned OpAMP client/server and customer owned k8s cluster, to which the vendor has no inbound access. IMHO vendor should still be able to remotely manage the clients in this case.

> Perhaps in some rare cases the remote configuration capability may be enabled, such that the k8s ConfigMap provides only the initial "empty" configuration and OpAMP is used to fetch the real config to run, but my gut feeling is that most k8s people would prefer not to do it.
> And auto-updating capabilities if I understand correctly are totally useless for k8s since any updates to executables done via OpAMP will be reverted after a routine pod restart.

Do you think that allowing OpAMP client to update k8s resources that comprise it would be a feasible compromise? Maybe something like a k8s-native supervisor, which is a k8s operator responsible for a specific CRD that represents this type of a k8s native agent?

tigrannajaryan:
> Maybe something like a k8s-native supervisor, which is a k8s operator responsible for a specific CRD that represents this type of a k8s native agent?

Yes, possibly, that's an interesting idea. I believe that is out of the scope of the particular Supervisor that is drafted here. 
If anyone is interested in working on a k8s operator that receives instructions via OpAMP and manages the fleet of Agents deployed on the cluster I suggest to open a separate issue and discuss there.

supervisor/conf/config-1.yml
kart: Will do.

supervisor/conf/config-1.yml
kart: My bad. Will fix this.

supervisor/conf/config-1.yml
kart: My thinking was to allow the vendor to specify how frequently the telemetry must be scraped by the supervisor. That's what the `period` specifies.

supervisor/conf/config-1.yml
kart: Probably not. I'll remove it.

supervisor/conf/config-2.yml
kart: So, here's where I wanted to add some more clarity. I am assuming that we want to the supervisor to be generic enough that it is able to manage things other than the collector (like Fluentbit). This can be accomplished in two ways:

One option, as you suggest, is to implement the supervisor as composable pieces which can be wired together. This way we can build a new supervisor executable for a collector or fluentbit. 

Another option is to implement the supervisor as composable pieces just like the above option but allow several supervisor implementations to exist within the same executable and running as separate Go routines, for example. The individual supervisor implementation will still be the same as option one except that more than one will reside in the same artifact. The advantage, in my understanding, is that if you have a system where, say, both collector and fluentbit are running, you need only one supervisor executable that manages it. Internally, that executable will spawn two Go routines one managing collector and the other managing the fluentbit by communicating with one or more OpAMP servers.

supervisor/supervisor.go
kart: This was bit of an oversight on my side. I think I was imagining that we implement the entire supervisor every time we have a new agent that requires supervision. But, many supervising capabilities might be overlapping and hence need not be re-implemented. Let me rethink this a bit and provide an alternate implementation.

supervisor/conf/config-1.yml
pmm-sumo: However, this is only metrics rather than the full telemetry? 

Also, I think the idea on the own telemetry is that agent can be given the configuration where it needs to be pushed and take care of that itself (however, I think in certain cases it could be also done by supervisor)

tigrannajaryan: This implies Supervisor scrapes the Agent for metrics. I would prefer not to do this, otherwise the Supervisor will slowly become a Collector on its own :-)

For Otel Collector I think we should go with a different approach: Supervisor to instruct the Collector where to send the metrics. This should be possible to do by configuring a Collector pipeline. This approach of course won't work for a generic Supervisor. I think it is fine to omit it as a feature in the generic Supervisor initially. We can discuss and decide if we want to make the Supervisor more complicated and be responsible for scraping in the future.

supervisor/conf/config-2.yml
tigrannajaryan: I think this complicates the Supervisor for a use case that is likely going to be rare (multiple agents on one machine). Even when you have multiple agents on one machine I don't think it is a big deal for each agent to have its own Supervisor. Let's keep it simple for now. The Supervisor if implemented correctly should be very lightweight, so we can afford to run one Supervisor instance per Agent instance.

PR #43, opened 2021-12-13
Generalize Agent description

tigrannajaryan:
This removes agent type and version fields and replaces them with generic identifying and non-identifying attributes. This will allow OpAMP to be used with other kinds of agents (e.g. Otel SDKs).

See spec issue https://github.com/open-telemetry/opamp-spec/issues/16

Implements spec PR changes https://github.com/open-telemetry/opamp-spec/pull/48

This change makes OpAMP more flexible, yet linecount/complexity of the code is almost the same (at least for the functionality that exists so far).

@open-telemetry/opamp-spec-approvers @open-telemetry/opamp-go-approvers please review/comment on the draft. If this looks reasonable I will submit a spec change and will finalize this PR.


internal/proto/opamp.proto
```diff
@@ -410,30 +408,39 @@ message AgentPackageAvailable {
...
+    //   instance uid (equal to ServerToAgent.instance_uid field) or any other value
+    //   that uniquely identifies the Agent in combination with other attributes.
+    // - any other attributes that are necessary for uniquely identifying the Agent's
+    //   own telemetry.
+    //
+    // The Agent SHOULD also include these attributes in the Resource of its own
+    // telemetry. The combination of identifying attributes SHOULD be sufficient to
+    // uniquely identify the Agent's own telemetry in the destination system to which
+    // the Agent sends its own telemetry.
+    repeated KeyValue identifying_attributes = 1;
```
tigrannajaryan, line 431: I went with key/value pair lists, similar to OTLP. We could also do a map[string]KeyValue, but we likely have to have to require uniqueness across identifyin and non-identifying attributes, so the map is not much of an improvement over a list (and is slower).

djaglowski (approved):

pmm-sumo (approved):

Aneurysm9 (approved):

client/client.go
```diff
@@ -80,13 +79,12 @@ type OpAMPClient interface {
 	// Once stopped OpAMPClient cannot be started again.
 	Stop(ctx context.Context) error
 
-	// SetAgentAttributes sets attributes of the Agent. The attributes will be included
+	// SetAgentDescription sets attributes of the Agent. The attributes will be included
 	// in the next status report sent to the server.
-	// MAY be called before Start(), in which case the attributes will be
-	// included in the first start report after  connection is established.
-	// MAY also be called after Start(), in which case the attributes will be included
+	// MUST NOT be called before Start().
```
carlosalberto, line 84: What's the reason for this?

tigrannajaryan: The initial value is provided in the `settings` passed to Start(). However, you raise a good question, perhaps we remove it from there and mandate that SetAgentDescription() is called before Start(). This way we will have only one way to set it.
Let me think about it.

tigrannajaryan: I will address this in a separate PR.

PR #46, opened 2021-12-20
APIs for implementing supervisor and its configuration file

kart:
This changeset describes several interface that an agent must implement
in order to fit itself in the supervisor model.

The basic idea is that:

1. We implement several of these interfaces as part of the reference
implementation.

2. Many of those default implementations may work with most agents but
some might be tailored fro OT Collector.

3. Custom OT collectors or other custom agents may change the
implementations. As long as the interfaces are honored, the rest of the
supervisor should work.

### DIFF
--- a/supervisor/agent/agent.go
+++ b/supervisor/agent/agent.go
@@ -1,0 +1,64 @@
+package agent
+
+import "github.com/open-telemetry/opamp-go/protobufs"
+
+type Settings struct {
+	// Capabilities are the list of capabilities that this agent supports.
+	// See `protobufs.AgentCapabilities'.
+	Capabilities []string `koanf:"capabilities"`
+
+	// Attrs are any key-value pair that describes the identifying and
+	// non-identifying attributes of the agent.
+	Attrs map[string]string `koanf:"attrs"`
+
+	// ConfigSpec describes this agent's configuration source from which
+	// its effective configuration is read and to which the remote configuration
+	// is written.
+	ConfigSpec *ConfigSpec `koanf:"config"`
+
+	// PackageSpec describes the set of files that make this agent.
+	PackageSpec *PackageSpec `koanf:"package"`
+}
+
+// Identifier provides this agent's unique identification.
+type Identifier interface {
+	// Type returns the FQDN of this agent. For example, for an OpenTelemetry
+	// Collector this should return "io.opentelemetry.collector".
+	Type() string
+
+	// Version returns this agent's build version.
+	//
+	// Version may return an error if it is unable to determine the Agent's
+	// version. This is possible, for example, if the only way to retrieve
+	// the agent's version is to make an HTTP request and the request fails.
+	Version() (string, error)
+
+	// InstanceID uniquely identifies this agent in combination with other
+	// attributes.
+	InstanceID() (string, error)
+
+	// Namespace return this agent's namespace. For example, for an OpenTelemetry
+	// Collector, this is equivalent to `service.namespace'.
+	Namespace() string
+}
+
+// Agent represents a long running process that allows for management via a
+// supervisor. OpenTelemetry Collector or a FluentBit daemon are some of the
+// examples of an Agent.
+type Agent interface {
+	// Identifier allows several identifying attributes like the Agent's
+	// type, version and namespace to be returned.
+	Identifier
+
+	// GetNonIdentifyingAttrs returns attributes that do not necessarily help
+	// identify the agent, but describe where it runs.
+	GetNonIdentifyingAttrs() ([]*protobufs.KeyValue, error)
+
+	// Commander allows the agent to be started, stopped and restarted.
+	// See `agent.Commander`
+	Commander
+
+	// HealthChecker allows the agent's health to be monitored.
+	// See `agent.HealthChecker`
+	HealthChecker
+}
--- a/supervisor/agent/commander.go
+++ b/supervisor/agent/commander.go
@@ -1,0 +1,68 @@
+package agent
+
+import "context"
+
+// CommandConfig describes a systemd-based or an exec-based way of starting,
+// stopping and restarting the agent. Only one of `ExecConfig' or `SystemdConfig'
+// must be specified.
+type CommandConfig struct {
+	// Name is the path to the Agent executable. Path can be relative
+	// or absolute. A relative path should be relative to the CWD of the
+	// supervisor.
+	Name string `koanf:"name"`
+
+	// Args are the set of command-line arguments that must be specified along
+	// with the executable to start the Agent.
+	Args []string `koanf:"args"`
+}
+
+// Process represents information related to a running agent process.
+type Process struct {
+	// PID is the process id of the agent process that is currently running.
+	//
+	// A watchdog may use this PID to stop this agent process at a later time.
+	PID int
+
+	// Done is the channel which is set when the process terminates. This allows
+	// for async start of the agent followed by a continuous monitoring of its
+	// termination.
+	//
+	// A watchdog, for instance, can wait on this channel to listen to process
+	// termination event and restart the agent if needed.
+	Done <-chan struct{}
+}
+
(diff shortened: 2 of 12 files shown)


kart:
Posting a few more observations that I made while working on this:

- On health check: the health check configuration is largely inspired by Kubernetes liveness probe. It currently supports a HTTP probe. However, much like Kubernetes, we can also support a exec-based probe. See https://v1-19.docs.kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

- On config specification: currently, the config specification is simplistic: a config file and whether or not it requires a reload-on-change. I think this idea can be expanded to allow for more complex configuration situations where the agent's configuration can be constructed from different sources --- collector, for example, allows configuration to be specified from a file, environment variables or even network. In such cases, it would be good if an endpoint on the collector is exposed from where its effective configuration can be obtained. Similarly, the new configuration can be applied as a PUT on a HTTP endpoint as well? Maybe something like this:

```
config:
    read:
        # source: file:/opt/opentelemetry-collector/config.yaml
        source: http://localhost:1010/config # GET
        content_type: application/yaml
    write:
        # dest: file:/opt/opentelemetry-collector/config.yaml
        dest: http://localhost:1010/config # PUT
        content_type: application/yaml
    auto_reload: false
```

- On the `Identifier`: this interface seems a bit awkward. An alternate way of representing this would be for the `agent.Agent` to implement `GetIdentifyingAttrs` that returns a `[]*protobufs.KeyValue`. My concern, however, is that the `protobufs.KeyValue` allows an implementation to ignore specifying `service.name` or `service.instance.id` which in-turn can result in some error cycles when communicating with the server which can be avoided. `Identifier`, on the other hand, forces the implementer to provide an implementation thereby relieving some of those concerns.

- On watcher: I wrestled for a while on allowing the watcher to be optional (as suggested in the supervisor doc by @tigrannajaryan) and if the agent is already managed by something like systemd, let it be -- but the more I thought about it, the more it seems that it is best if the supervisor always keeps the agent as its child process and let itself be managed by systemd if needed. In fact, I think we should always do that in OpAMP reference implementation and recommend custom implementations to do the same as well. The reason is that de-coupling the agent and the supervisor allows for poor control and leaves room for race conditions and other problems if the agent can be terminated without the supervisor being aware of it.

supervisor/internal/examples/config_sample.yaml
```diff
@@ -0,0 +1,78 @@
...
+watchdog:
+    # Command describes how the agent must be exec'd.
+    command:
+        # Full or relative path to the agent executable.
+        name: /opt/otcol/bin/otel
+        # Set of command-line arguments required to start the `executable`
+        # specified above.
+        args:
+            - --config
+            - /etc/otcol/config.yaml
```
tigrannajaryan, line 23: It is a pity that we need to specify the config file location twice. It is probably inevitable for a generic supervisor. For Otel Collector specific Supervisor we should avoid doing this.

supervisor/agent/commander.go
```diff
@@ -0,0 +1,68 @@
...
+	// for stopping the process depends on whether or not the agent is managed
+	// by a `systemd'-like facility.
+	//
+	// Stop MUST wait for the process to be terminated from an OS perspective.
+	Stop(ctx context.Context, process *Process) error
+
+	// Restart the agent process as defined by `process'.
+	//
+	// Restart, most likely, is implemented as a Stop followed by a Start in
+	// which case the Stop part might block until the process terminates.
```
tigrannajaryan, line 60: Why do we need Restart()? How is the behavior is expected to be different from Stop() followed by Start()?

supervisor/agent/commander.go
```diff
@@ -0,0 +1,68 @@
...
+	// Start returns the process information pertaining to the new agent process
+	// or an error if the process could not be successfully restarted.
+	Start() (*Process, error)
+
+	// Stop the agent process as defined by `process'. The actual mechanism
+	// for stopping the process depends on whether or not the agent is managed
+	// by a `systemd'-like facility.
+	//
+	// Stop MUST wait for the process to be terminated from an OS perspective.
+	Stop(ctx context.Context, process *Process) error
```
tigrannajaryan, line 55: Should Stop() be a Process method instead? This seems to be what `os` package does: https://pkg.go.dev/os#StartProcess

supervisor/agent/commander.go
```diff
@@ -0,0 +1,68 @@
...
+	// Args are the set of command-line arguments that must be specified along
+	// with the executable to start the Agent.
+	Args []string `koanf:"args"`
+}
+
+// Process represents information related to a running agent process.
+type Process struct {
+	// PID is the process id of the agent process that is currently running.
+	//
+	// A watchdog may use this PID to stop this agent process at a later time.
```
tigrannajaryan, line 23: Why is it necessary if there is already a Stop() method?

supervisor/supervisor.go
```diff
@@ -0,0 +1,124 @@
...
+
+func (l *Logger) Debugf(format string, v ...interface{}) {
+	l.Logger.Printf(format, v...)
+}
+
+func (l *Logger) Errorf(format string, v ...interface{}) {
+	l.Logger.Printf(format, v...)
+}
+
+type supervisor struct {
```
tigrannajaryan, line 30: This looks good and roughly the direction that I was thinking we should take with the supervisor.

One question that would be useful to answer: is this generic supervisor flexible enough to allow us to build a specialized Otel Collector Supervisor that knows for example that `--config` flag must be used to specify the location of the config file for the Collector processor and no longer requires the user to specify the `args` settings in the supervisor config file.

supervisor/supervisor.go
```diff
@@ -0,0 +1,124 @@
...
+	logger *Logger
+
+	configFile string
+	config     *Configuration
```
tigrannajaryan, line 34: Should we also externalize the `Configuration` as an interface that we load externally and pass to the `supervisor`? If we do it then we can have specialize implementations of the `Configuration` that have a different yaml format and for example no longer require `args` section because knowledge about command line arguments is hard-coded in that specialized implementation.

supervisor/agent/agent.go
```diff
@@ -0,0 +1,64 @@
+package agent
+
+import "github.com/open-telemetry/opamp-go/protobufs"
+
+type Settings struct {
+	// Capabilities are the list of capabilities that this agent supports.
+	// See `protobufs.AgentCapabilities'.
+	Capabilities []string `koanf:"capabilities"`
```
tigrannajaryan, line 8: Can we use "mapstructure" field tags instead of "koanf"? That's what we do in Collector codebase and would be nice to be consistent here.

supervisor/agent/commander.go
portertech: Cover other reasons for agent termination?

supervisor/internal/examples/config_sample.yaml
portertech: How can you avoid it? An assumed default path?

tigrannajaryan: I mean avoid specifying it twice. The path to the config file is also specified on line 68.

If the Supervisor knows that it deals with Otel Collector then we don't need to tell the Supervisor how to pass the path to the config file on the command line using `args` section, the Supervisor can copy the value from line 68 into the command line argument `--config` automatically.

Because this is a generic Supervisor it does not have this knowledge about Otel Collector and we have to have an `args` section as well.

tigrannajaryan:
I have been thinking about the Supervisor and it seems to me implementing a generic Supervisor such that it has great usability is going to be difficult.

I think we probably want to modify our approach a bit. Instead of building a generic Supervisor we probably want to create a small library with helpers which can be used to create Supervisors and then create an OpenTelemetry Collector-specific Supervisor. After this we will more experience and it may help us understand if we can create a nice, more generic Supervisor.

tigrannajaryan:
Closing since the author will not be able to work on this anymore.

PR #52, opened 2022-01-11
Move examples into a separate module

tigrannajaryan:
This is to avoid polluting the library with the dependencies that the examples need.

This is one of the explicitly called out cases for having multiple modules in the Go modules docs:
https://github.com/golang/go/wiki/Modules#should-i-have-multiple-modules-in-a-single-repository


internal/examples/go.mod
```diff
@@ -0,0 +1,50 @@
+module github.com/open-telemetry/opamp-go/internal/examples
+
+go 1.17
+
+require (
+	github.com/knadh/koanf v1.3.3
+	github.com/oklog/ulid/v2 v2.0.2
+	github.com/open-telemetry/opamp-go v0.1.0
```
tigrannajaryan, line 8: This is what I think I am supposed to do when splitting an existing directory into a module (reference a newer than current version).

codeboten (approved):

djaglowski (approved):

PR #53, opened 2022-01-18
Allow OpAMP server to request an agent restart or shutdown

andykellr:
* Adds ServerToAgentCommand to allow the server to request a restart or shutdown.
* Adds AcceptsRestartRequests to AgentCapabilities so that agents can indicate if they support restart.
* Adds OnCommand to the client Callbacks which is called when the server requests shutdown via the ServerToAgentCommand
* Moved client.CallbacksStruct to types.CallbacksStruct so that it could be used it in receiver_test.go without an import cycle.


internal/proto/opamp.proto
```diff
@@ -98,6 +98,9 @@ message ServerToAgent {
 
         // The server asks the agent to report addon statuses.
         ReportAddonStatus     = 0x00000002;
+
+        // The server asks the agent to restart.
+        RequestRestart        = 0x00000004;
```
tigrannajaryan, line 103: I think having this in the Flags is ambiguous. For example if both ReportEffectiveConfig and RequestRestart are set which one is the Agent supposed to do? Both? In what order?

tigrannajaryan (commented):
@andykellr thanks for the proposal.
It may useful to discuss this in the spec repo in the issue and only after agreeing to a specific approach work on the implementation.

internal/proto/opamp.proto
andykellr: It could be documented that RequestRestart should take precedence. It seems unlikely that a server implementation would have a good reason to set both flags, but I agree that the behavior should be specified.

I see 3 ways to implement this capability:

1) A new message type which would have no ambiguity but require more careful handling of the unmarshal process.

```
message ServerToAgentRestartRequest {
    string instance_uid = 1;
}
```

2) A new field on the ServerToAgent message which would still be ambiguous because flags could also be set.

```
message ServerToAgent {
    string instance_uid = 1;
    ...
    bool restart_request = 
    ...
```

3) A new flag as proposed.

I am certainly open to other ideas.

tigrannajaryan: 1. A new ServerToAgentRestartRequest is probably excessive since it requires changes to lots of existing spec/implementation which only expects one type of message ServerToAgent from the server to agent. I would prefer to avoid this.
2. A new field on the ServerToAgent message may be the way to go. It can be a sub-message, particularly if we expect to carry any additional data in it (do we?).
3. The ambiguity of the flag can be avoided by documenting the behavior but it is a single bit which makes passing any additional parameters impossible.

What if we make it a new field on the ServerToAgent message and call it something like ServerToAgentCommand? We can have
```protobuf
message ServerToAgentCommand {
  enum CommandType {
    RESTART,
    SHUTDOWN,
    // other commands?
  }
  CommandType type = 1;
  // maybe some other data for the command
}
```
We will need to document that either the command cannot be mixed with other ServerToAgent fields or it is applied in a particular specific order (which the spec already defines for all fields). This will remove the ambiguity.

andykellr: ServerToAgentCommand sounds like a good approach.

andykellr:
I was traveling and finally had a chance to refactor this PR to use a `ServerToAgentCommand`.

tigrannajaryan:
> I was traveling and finally had a chance to refactor this PR to use a `ServerToAgentCommand`.

I think this is a reasonable approach. Can you please submit a PR against https://github.com/open-telemetry/opamp-spec and also make sure the change is visible/discussed in the next workgroup meeting?

tigrannajaryan:
@andykellr Please update this PR to only implement changes accepted in https://github.com/open-telemetry/opamp-spec/pull/64 for now.

andykellr:
@tigrannajaryan I think this PR is ready to merge. Please let me know if any other changes are needed.

client/internal/receiver.go
```diff
@@ -168,3 +174,9 @@ func (r *Receiver) processErrorResponse(body *protobufs.ServerErrorResponse) {
 func (r *Receiver) rcvAddonsAvailable(addons *protobufs.AddonsAvailable) {
 	// TODO: implement this.
 }
+
+func (r *Receiver) rcvCommand(command *protobufs.ServerToAgentCommand) {
+	if command != nil {
```
pmm-sumo, line 179: It's a nit, but `command` is essentially being checked for the second time here, i.e. this function would never be called with `nil` command

client/types/callbacks.go
```diff
@@ -129,7 +129,137 @@ type Callbacks interface {
...
 
+	// OnCommand is called when the server requests that the connected agent perform a command.
+	OnCommand(command *protobufs.ServerToAgentCommand) error
+
 	// For all methods that accept a context parameter the caller may cancel the
 	// context if processing takes too long. In that case the method should return
 	// as soon as possible with an error.
 }
+
+type CallbacksStruct struct {
```
pmm-sumo, line 140: I'm not sure if moving `callbacks.go` to types is a good fit. I think this directory contained interface only until now

andykellr: I use it in `client/internal/receiver_test.go` and needed to move it out of client to avoid an import cycle.

This struct is just a convenience that has no dependencies outside of types, protobufs, and context. There is no implementation logic.

client/internal/receiver.go
andykellr: I agree that it is redundant, but I left it that way because the first check is making it clear that if there is a command that nothing else is done in processReceivedMessage. The second check is just a parameter guard following the pattern of the other functions that do nothing if called with a nil component of the message.

internal/proto/opamp.proto
```diff
@@ -111,6 +111,11 @@ message ServerToAgent {
     // be omitted in subsequent ServerToAgent messages by setting it to
     // UnspecifiedServerCapability value.
     ServerCapabilities capabilities = 8;
+
+    // Allows the Server to instruct the agent to perform a command, e.g. RESTART. This field should not be specified
+    // with fields other than instance_uid and capabilities. If specified, other fields will be ignored and the command
+    // will be performed.
+    ServerToAgentCommand command = 10;
```
tigrannajaryan, line 118: @pmm-sumo please also add the field 9 `agent_identification` in a new PR. I was confused by the gap.

client/types/callbacks.go
tigrannajaryan: I think moving is fine. IIRC, the `types` package was created to break the dependency cycle. Alternatively we could move `CallbacksStruct` to its own package, but I don't think it is worth creating a new package just for one struct.

pmm-sumo: Thanks for the explanation

tigrannajaryan (approved):
LGTM.
Let's wait for @pmm-sumo to also approve.

pmm-sumo (approved):
LGTM too :)

PR #57, opened 2022-03-14
Expose the RemoteAddr and request Header via the Connection interface

andykellr:
This was originally proposed in a slack message in the CNCF otel-agentmanwg channel.

### DIFF
--- a/server/connection.go
+++ b/server/connection.go
@@ -2,6 +2,8 @@ package server
 
 import (
 	"context"
+	"net"
+	"net/http"
 
 	"github.com/gorilla/websocket"
 	"google.golang.org/protobuf/proto"
@@ -12,9 +14,18 @@ import (
 
 type connection struct {
 	wsConn *websocket.Conn
+	header http.Header
 }
 
 var _ types.Connection = (*connection)(nil)
+
+func (c connection) RemoteAddr() net.Addr {
+	return c.RemoteAddr()
+}
+
+func (c connection) Header() http.Header {
+	return c.header
+}
 
 func (c connection) Send(ctx context.Context, message *protobufs.ServerToAgent) error {
 	bytes, err := proto.Marshal(message)
--- a/server/types/connection.go
+++ b/server/types/connection.go
@@ -2,6 +2,8 @@ package types
 
 import (
 	"context"
+	"net"
+	"net/http"
 
 	"github.com/open-telemetry/opamp-go/protobufs"
 )
@@ -9,6 +11,12 @@ import (
 // Connection represents one OpAMP WebSocket connections.
 // The implementation MUST be a comparable type so that it can be used as a map key.
 type Connection interface {
+	// RemoteAddr returns the remote network address of the connection.
+	RemoteAddr() net.Addr
+
+	// Header returns the request header fields of the http request that opened the connection.
+	Header() http.Header
+
 	// Send a message. Should not be called concurrently for the same Connection instance.
 	// Blocks until the message is sent.
 	// Should return as soon as possible if the ctx is cancelled.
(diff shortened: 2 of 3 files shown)


server/types/connection.go
```diff
@@ -2,13 +2,21 @@ package types
...
 )
 
 // Connection represents one OpAMP WebSocket connections.
 // The implementation MUST be a comparable type so that it can be used as a map key.
 type Connection interface {
+	// RemoteAddr returns the remote network address of the connection.
+	RemoteAddr() net.Addr
+
+	// Header returns the request header fields of the http request that opened the connection.
+	Header() http.Header
```
tigrannajaryan, line 18: Do we need the Header here? The header can be already accessed in the `OnConnecting` callback, so this seems redundant.

andykellr: `OnConnecting` provides the request but at that point there is no agent information available, e.g. `InstanceUid` or `AgentDescription`, which would identify which agent is actually connecting. This information will come later in a message after the connection is established. There is also no `Connection` because the connection isn't established. Providing the Header on the connection will make this information available in the other callbacks.

andykellr: I see. HTTP headers are generally not accessible in gRPC implementations. This presents an issue with the existing `OnConnecting` callback because http.Request is also not available in gRPC implementations. We can take this out if you prefer.

tigrannajaryan: Let's add `RemoteAddr()` in this PR and we can think about what we do with `Header()` separately.

andykellr: Done, thanks.

pmm-sumo: For the headers, this is how we handle them for gRPC and HTTP in OpenTelemetry Collector:

https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configgrpc/configgrpc.go#L401-L409

https://github.com/open-telemetry/opentelemetry-collector/blame/main/config/confighttp/clientinfohandler.go#L52-L59

tigrannajaryan: So, looks like it is called Metadata and defined as:
```go
// Metadata is an immutable map, meant to contain request metadata.
type Metadata struct {
	data map[string][]string
}
```

server/types/connection.go
```diff
@@ -2,13 +2,17 @@ package types
...
 import (
 	"context"
+	"net"
 
 	"github.com/open-telemetry/opamp-go/protobufs"
 )
 
 // Connection represents one OpAMP WebSocket connections.
 // The implementation MUST be a comparable type so that it can be used as a map key.
 type Connection interface {
```
tigrannajaryan, line 12: @andykellr to support the use case that you have I think we also need to add `net.Conn` as a field. This will allow to access the TLS certificate. I believe `net.Conn` is accessible for WebSocket. If you can confirm it is also accessible for gRPC connections and for plain HTTP connections then it should be OK to add it to this interface.

tigrannajaryan: I will go ahead and merge this PR. Feel free to submit a new PR that exposes net.Conn.

server/connection.go
```diff
@@ -16,6 +17,10 @@ type connection struct {
 
 var _ types.Connection = (*connection)(nil)
 
+func (c connection) RemoteAddr() net.Addr {
+	return c.RemoteAddr()
```
tigrannajaryan, line 21: Is this recursively calling itself or I am misreading the code?

tigrannajaryan (commented):
Please add a text to verify the new functonality.

server/connection.go
andykellr: You didn't read that wrong. I put this PR together quickly to begin discussion and it was an oversight. It clearly needs a test case as well.

server/types/connection.go
tigrannajaryan: @andykellr this PR looks good to me now. I will merge as is and you can address this comment in a separate PR if you need the `Conn`.

PR #58, opened 2022-03-22
AgentIdentification protobuf and callback

pmm-sumo:
Implements OpAMP Spec change https://github.com/open-telemetry/opamp-spec/pull/63


pmm-sumo:
@tigrannajaryan do you think it would be worthwhile to implement this in the examples? I was wondering about that but then wasn't fully sure if it wouldn't make things too complex for the sake of the examples

tigrannajaryan:
> @tigrannajaryan do you think it would be worthwhile to implement this in the example agent/server? I was wondering about that but then wasn't fully sure if it wouldn't make things too complex for the sake of the examples (e.g. agent would need to persist the instance id)

I think it is worth showing in the example. Doesn't necessarily have to persist the instance id, can just keep it in memory during the execution session.

client/internal/receiver.go
```diff
@@ -168,3 +169,14 @@ func (r *Receiver) processErrorResponse(body *protobufs.ServerErrorResponse) {
 func (r *Receiver) rcvAddonsAvailable(addons *protobufs.AddonsAvailable) {
 	// TODO: implement this.
 }
+
+func (r *Receiver) rcvAgentIdentification(ctx context.Context, agentId *protobufs.AgentIdentification) {
+	if agentId == nil {
+		return
+	}
+
+	err := r.callbacks.OnAgentIdentification(ctx, agentId)
```
tigrannajaryan, line 178: I think in addition to calling the callback we need to update the instanceUid in the Sender (if the callback doesn't return an error).

client/callbacks.go
```diff
@@ -124,3 +130,13 @@ func (c CallbacksStruct) OnAgentPackageAvailable(
...
 }
+
+func (c CallbacksStruct) OnAgentIdentification(
+	ctx context.Context,
+	agentId *protobufs.AgentIdentification,
+) (string, error) {
+	if c.OnAgentIdentificationFunc != nil {
+		return c.OnAgentIdentificationFunc(ctx, agentId)
+	}
+	return "", errCallbackNotSet
```
pmm-sumo, line 141: I wasn't sure what's the most elegant way to handle the case when no function is set. Alternatives are using `*string` as the return value (and then just assuming that if the function did not extract the instance_uid, it will return nil) or having a bool flag with information if new instance uid was received or not. If you have any preference here, I will gladly accept it

tigrannajaryan: What's the intent here? Do we want to allow the callback function to override the suggested new id? I am not sure that's a good idea. Is there a use-case for this?

pmm-sumo: My thought was consistency with the other callback functions + perhaps ability to reset it to the original id (when agent doesn't support changing it). I will do another iteration over that tomorrow

tigrannajaryan: When agent doesn't support changing the id I would expect the callback to return an error. That can be an indication to the OpAMP client that the new id is rejected.

pmm-sumo: @tigrannajaryan did a modest change and removed `errCallbackNotSet`

tigrannajaryan: I am still confused about the purpose of returning the string from OnAgentIdentification callback. I cannot see the use case where the callback should be allowed to override the offered id.

pmm-sumo: > I am still confused about the purpose of returning the string from OnAgentIdentification callback. I cannot see the use case where the callback should be allowed to override the offered id.

We need `OnAgentIdentifaction` to return `string` as we want to use for [updating the client id](
https://github.com/open-telemetry/opamp-go/pull/58/files#diff-ea4da29b17874d95bdf2949cd7d6bac9b86180e6be366d0308ff3b3240346472R188) later

We might get rid of `OnAgentIdentificationFunc` completely though and just return the value stored in the message (however by omitting it we lose the ability for client to easily implement some additional functions on agent identification (e.g. it might require ULID and validate it, etc.)

tigrannajaryan: Why do we need OnAgentIdentifaction to return that string? The string value is already available as NewInstanceUid in the AgentIdentification message.

>We might get rid of OnAgentIdentificationFunc completely though and just return the value stored in the message (however by omitting it we lose the ability for client to easily implement some additional functions on agent identification (e.g. it might require ULID and validate it, etc.)

I don't suggest this. I think OnAgentIdentificationFunc is necessary so that it can return an error if for some reason the agent is unable to accept the new instance id. However, I do not see why OnAgentIdentificationFunc needs to return a string id. If it is the same id as it is passed as an argument then it is pointless. If it is a different id than what is passed then I don't understand why OnAgentIdentificationFunc needs to be able to return a different id and override the id offered by the server.

I may be missing something.

pmm-sumo: Ahh, I see your point now. One extra reason for OnAgentIdentification to return string was that it could potentially do some agent-specific normalization of the id, if needed. But then it's a weak argument to include extra complexity.

In either case, I updated the code just now. Let me know if this looks better

client/internal/receiver.go
```diff
@@ -168,3 +169,23 @@ func (r *Receiver) processErrorResponse(body *protobufs.ServerErrorResponse) {
...
+	if err != nil {
+		r.logger.Errorf("Error while retrieving agent identification: %v", err)
```
tigrannajaryan, line 180:
```suggestion
		r.logger.Errorf("Error while updating agent identification: %v", err)
```

client/internal/receiver.go
```diff
@@ -168,3 +169,23 @@ func (r *Receiver) processErrorResponse(body *protobufs.ServerErrorResponse) {
...
+		return
+	}
+
+	instanceUid := agentId.NewInstanceUid
+	if instanceUid == "" {
+		r.logger.Debugf("Empty instance uid is not allowed")
+		return
+	}
```
tigrannajaryan, line 188: Should this check happen before we call the callback? Otherwise the callback may update its internal id to empty string but we reject it here.

tigrannajaryan (commented):
Mostly LGTM now, just a couple small comments.

tigrannajaryan:
Also needs a rebase.

client/internal/receiver.go
pmm-sumo: I was thinking mostly about making sure we pass a valid id to `SetInstanceUid` but perhaps it should have its own check and we could move this one up

tigrannajaryan: Yes, I agree. I suggest this order:
1. Check non-empty, exit if empty.
2. Callback, exit if error.
3. SetInstanceUid()

pmm-sumo: Just updated the code (including rebase). Now SetInstanceUid might return an error, which is logged here

tigrannajaryan (approved):

Issue #55, opened 2022-03-04
Client should send status report on reconnect


andykellr:
Currently, the status report message is prepared here in client.Start:

https://github.com/open-telemetry/opamp-go/blob/5255de573f20af81acfb8ab864bfea8d986bbeef/client/clientimpl.go#L109-L113

It then begins runUntilStopped and in runOneCycle it ensures it is connected before entering the ReceiverLoop to receive massages from the server:

https://github.com/open-telemetry/opamp-go/blob/5255de573f20af81acfb8ab864bfea8d986bbeef/client/clientimpl.go#L313-L318

If the connection is closed, it will repeat runOneCycle and reconnect, but after reconnection no status message is sent because no message has been prepared.

In our own agent, the workaround was to call SetAgentDescription in the OnConnect callback, but I think the implementation should be doing this automatically.

andykellr:
Here is the section of the spec regarding the agent sending a status report upon connection. I doesn't specifically address reconnecting, but I think this is implied.

https://github.com/open-telemetry/opamp-spec/blob/410dc86792d6109050e767a1d76de12a5b9f0129/specification.md?plain=1#L485-L491

dsvanlani:
@tigrannajaryan I'm working on a PR for this, but will not have anything to show by this afternoons work group.

PR #63, opened 2022-04-13
Ensure client includes AgentDescription on Reconnect

dsvanlani:
Resolves: https://github.com/open-telemetry/opamp-go/issues/55

This PR ensures that the client includes AgentDescription every time it connects to the server, not just after `client.Start`.

Current behavior:
- Clients connection to server is interrupted.
- Client reconnects.
- Client does not prepare a "first status message" with an `AgentDescription`.

Expected behavior:
- Clients connection to server is interrupted.
- Client reconnects.
- Client prepares and sends a status including AgentDescription.

### DIFF
--- a/client/clientimpl.go
+++ b/client/clientimpl.go
@@ -152,6 +152,9 @@ func (w *client) SetAgentDescription(descr *protobufs.AgentDescription) error {
 		return errAgentDescriptionMissing
 	}
 
+	// store the agent description to send on reconnect
+	w.settings.AgentDescription = descr
+
 	w.sender.UpdateNextStatus(func(statusReport *protobufs.StatusReport) {
 		statusReport.AgentDescription = descr
 	})
--- a/client/clientimpl.go
+++ b/client/clientimpl.go
@@ -106,22 +106,6 @@ func (w *client) Start(settings StartSettings) error {
 		w.requestHeader["Authorization"] = []string{w.settings.AuthorizationHeader}
 	}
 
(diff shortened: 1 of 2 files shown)


andykellr (approved):
I collaborated with @dsvanlani on this fix so I would appreciate another review.

tigrannajaryan (approved):
LGTM

tigrannajaryan:
Thanks for the fix.

tigrannajaryan:
I may have been too quick to merge this.

Re-reading the spec, it says:
```
The Agent MUST send a status report:
* First time immediately after connecting to the Server. The status report MUST
  be the first message sent by the Agent.
* Subsequently every time the status of the Agent changes.
```

Nowhere does it says that the AgentDescription must be also included in the status report. On the contrary, [the spec says](https://github.com/open-telemetry/opamp-spec/blob/410dc86792d6109050e767a1d76de12a5b9f0129/specification.md?plain=1#L624):
```
<h4 id="agent_description">agent_description</h4>

The description of the agent, its type, where it runs, etc. See
[AgentDescription](#agentdescription-message) message for details.

This field SHOULD be unset if no Agent description fields have changed since the
last StatusReport was sent.
```

So, this fix it correct that it sends the status report after reconnect (it was a bug that we didn't), but we now unnecessarily include the AgentDescription in these subsequent status reports.

@dsvanlani Can you do a follow-up PR that corrects this behavior?

PR #65, opened 2022-04-28
Unify addons and agent packages

tigrannajaryan:
Applies spec change https://github.com/open-telemetry/opamp-spec/pull/77


internal/proto/opamp.proto
```diff
@@ -594,61 +579,66 @@ message RemoteConfigStatus {
...
+    // Package name. MUST be always set and MUST match the key in the packages field
+    // of PackageStatuses message.
     string name = 1;
 
-    // The hash of the addon that the agent has.
-    // MUST be set if the agent has this addon.
-    // MUST be empty if the agent does not have this addon. This may be the case for
-    // example if the addon was offered by server but failed to install and the agent
-    // did not have this addon previously.
+    // The version of the package that the Agent has.
```
pmm-sumo, line 604: This is nit again, I see that sometimes it's "the Agent" and sometimes "the agent"

pmm-sumo (approved):
Few nits in some comments, otherwise looks good!

client/types/callbacks.go
```diff
@@ -118,16 +118,11 @@ type Callbacks interface {
 		certificate *protobufs.ConnectionSettings,
 	) error
 
-	// OnAddonsAvailable is called when the server has addons available which are
+	// OnPackagesAvailable is called when the server has addons available which are
```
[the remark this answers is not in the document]
tigrannajaryan: Fixed.

internal/proto/opamp.proto
tigrannajaryan: This inconsistency is widespread. I will do a separate PR to fix all to avoid adding noise to this PR.

tigrannajaryan:
@pmm-sumo I missed quite a few other places where I still had "addons". All should be fixed now. PTAL.

client/types/callbacks.go
```diff
@@ -118,16 +118,11 @@ type Callbacks interface {
...
 	// different from what the agent indicated it has via
 	// LastServerProvidedAllAddonsHash.
```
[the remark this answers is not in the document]
tigrannajaryan: Fixed.

andykellr (changes requested):
There is some inconsistency around version and hash.

client/types/packagessyncer.go
```diff
@@ -0,0 +1,56 @@
...
+	// of the file must be replaced by the data. The data must be read until
+	// it returns an EOF. If reading from data fails UpdateContent must abort and return
+	// an error.
+	// Content hash must be updated if the data is updated without failure.
+	// The function must cancel and return an error if the context is cancelled.
+	UpdateContent(ctx context.Context, packageName string, data io.Reader, contentHash []byte) error
+
+	// SetPackageHash must remember the hash for the specified package. Must be returned
+	// later when PackageHash is called. SetPackageHash is called after all UpsertFile
+	// and DeleteFile calls complete successfully.
```
pmm-sumo, line 46: I don't see the referenced functions (`UpsertFile` and `DeleteFile`). Should it refer to `UpdateContent` and `DeletePackage` instead?

tigrannajaryan: Good catch, this was remaining from the time when addons had multiple files. It should be "after Update Content".

internal/proto/opamp.proto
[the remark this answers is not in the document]
tigrannajaryan: Thanks for catching. I failed to copy/paste from the spec properly :-(
The spec has both the "hash" and "version" fields: https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#packagestatus-message

Should be fixed now. Please have a look again.

I am also not sure if there is a good way avoid the duplication of the content in the comments that describe the fields (the text is almost, but not quite the same for version and hash fields).

andykellr (approved):

pmm-sumo (approved):

PR #66, opened 2022-05-06
Implement synchronization of AgentDescription and EffectiveConfig messages

tigrannajaryan:
This implements synchronization of AgentDescription and EffectiveConfig messages
according to the OpAMP spec. The hash fields are now correctly used and Report* flags
received from the Server are correctly handled.

Also added a new "expect" mode to MockServer to make writing client-server
interaction tests easier. Some existing tests may also be converted to this
new style later.

TODO:
- Implement synchronization of RemoteConfigStatus and PackageStatuses. The PR was getting too
  large, so I decided to do these 2 in a separate PR.
- Refactor WSClient and HTTPClient to reduce duplication of client logic. I moved some of it to
  a new clientcommon.go file, but more refactoring is needed - we still have to much unnecessary
  code duplication.


tigrannajaryan:
@open-telemetry/opamp-go-approvers PTAL.

pmm-sumo (approved):
That's a large PR. The first part looks good. I especially like the new tests added (with expect semantics)

PR #67, opened 2022-05-11
Refactor to deduplicate common client functionality

tigrannajaryan:
- There are no changes in functionality. Tests are not touched (except package
  renaming).
- The OpAMP client logic that is common between WebSocket and plain HTTP
  transports is now implemented in one place: ClientCommon struct.
- Renamed HTTPLooper to HTTPSender to have more consistent naming (we have WSSender).
- StartSettings moved to types package (forced change to avoid dependency loop).

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -2,47 +2,10 @@ package client
 
 import (
 	"context"
-	"crypto/tls"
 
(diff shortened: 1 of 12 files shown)


andykellr (approved):
Nice refactor

PR #68, opened 2022-05-12
Implement hash-based remote config status

tigrannajaryan:
- Used Hash field correctly.
- Added a new SaveRemoteConfigStatus callback to ask the Agent to save
  the remote config status for the future.
- Updated opamp.proto to match the OpAMP spec.


tigrannajaryan:
@open-telemetry/opamp-go-approvers please review.

client/internal/clientstate.go
```diff
@@ -67,3 +68,26 @@ func (s *ClientSyncedState) SetAgentDescription(descr *protobufs.AgentDescriptio
...
+	h := sha256.New()
+	h.Write(status.LastRemoteConfigHash)
+	h.Write([]byte(status.Status.String()))
+	h.Write([]byte(status.ErrorMessage))
+	status.Hash = h.Sum(nil)
+}
+
+func (s *ClientSyncedState) SetRemoteConfigStatus(status *protobufs.RemoteConfigStatus) error {
+	if len(status.Hash) == 0 {
+		return errors.New("hash field must be set, use CalcHashRemoteConfigStatus")
```
andykellr, line 83: If we can determine that the hash is not set and we are able to compute the hash, then should this code just compute the hash?

We could do the same above in SetAgentDescription.

Also, comparing to SetAgentDescription, should this produce an errRemoteConfigStatusMissing if status is nil?

andykellr (approved):

Issue #70, opened 2022-05-13
CalcHashEffectiveConfig not deterministic


cpheps:
It looks like in [CalcHashEffectiveConfig](https://github.com/open-telemetry/opamp-go/blob/main/client/helpers.go#L50-L60) that the hashing algorithm is iterating over the `msg.ConfigMap.ConfigMap` which is of type `map[string]*AgentConfigFile`. Iteration of a map in go is not deterministic so the hash computed may be different using the same set of data depending on the key order during iteration.

Should probably first sort the keyspace then iterate over the map.

PR #73, opened 2022-05-13
Make Hash field calculations internal responsibility

tigrannajaryan:
The Hash is now calculated internally by the client instead of requiring
the caller to use the helper function to calculate the Hash before
passing the message to the client.


pmm-sumo (approved):

andykellr:
This looks good, but should we keep helpers or provide something similar for the server to use? Particularly with the json encode and the issue mentioned in #70, it may not be obvious how the server should compute a hash to compare with the client hash.

andykellr:
It should be possible to internalize the hash computation on the server side as well, but I haven't had time to think though the changes required.

tigrannajaryan:
> This looks good, but should we keep helpers or provide something similar for the server to use? Particularly with the json encode and the issue mentioned in #70, it may not be obvious how the server should compute a hash to compare with the client hash.

The server should not need to computer the hash. It only needs to store and compare. See example implementation: https://github.com/open-telemetry/opamp-go/blob/dfbe616ffb2211c2e75870f461e4521f1419de94/internal/examples/server/data/agent.go#L113

However if we find that for some other cases it is needed we can refactor again to make them publicly available helpers.

tigrannajaryan:
Rebased and conflicts resolved.

Also need to fix https://github.com/open-telemetry/opamp-go/issues/70 but will do in a separate PR.

andykellr (approved):

andykellr:
> > This looks good, but should we keep helpers or provide something similar for the server to use? Particularly with the json encode and the issue mentioned in #70, it may not be obvious how the server should compute a hash to compare with the client hash.
> 
> The server should not need to computer the hash. It only needs to store and compare. See example implementation:
> 
> https://github.com/open-telemetry/opamp-go/blob/dfbe616ffb2211c2e75870f461e4521f1419de94/internal/examples/server/data/agent.go#L113
> 
> However if we find that for some other cases it is needed we can refactor again to make them publicly available helpers.

I was thinking of EffectiveConfig, not AgentDescription, but it's a similar concern.

https://github.com/open-telemetry/opamp-go/blob/dfbe616ffb2211c2e75870f461e4521f1419de94/internal/examples/server/data/agent.go#L254

tigrannajaryan:
> I was thinking of EffectiveConfig, not AgentDescription, but it's a similar concern.

Right. I think the server doesn't need to compute the hashes.

PR #75, opened 2022-05-16
Make calcHashEffectiveConfig deterministic

### LINKED ISSUE #70: CalcHashEffectiveConfig not deterministic

cpheps:
Resolves #70 

Made `calcHashEffectiveConfig` by sorting the keyspace before computing the hash. Also, added checks for nil or empty cfg maps and returns an empty hash for those cases. Added unit tests to function.


client/internal/clientcommon.go
```diff
@@ -191,9 +192,27 @@ func (c *ClientCommon) SetAgentDescription(descr *protobufs.AgentDescription) er
 // calcHashEffectiveConfig calculates and sets the Hash field from the rest of the
 // fields in the message.
 func calcHashEffectiveConfig(msg *protobufs.EffectiveConfig) {
+	cfgMap := msg.GetConfigMap().GetConfigMap()
+
+	// If no cfgMap return empty hash
+	if cfgMap == nil || len(cfgMap) == 0 {
```
tigrannajaryan, line 198: The spec does not prohibit empty configurations. Empty map is a valid (empty) configuration. We need to calc the hash in that case.

tigrannajaryan (commented):
Thanks for the PR @cpheps

client/internal/clientcommon.go
cpheps: So the hash would have to just be `msg.Hash = h.Sum(nil)` since there's nothing to add to the hash. I avoided that since it isn't determined by the config at all. I wasn't sure if `h.Sum(nil)` is better than just empty byte array.

I have no problem switching it to an "empty" hash but I just wanted to throw out my thought process.

tigrannajaryan: The spec [requires](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agent-to-server-state-synchronization) that the Hash field is always set:
>The compression is done by omitting all fields in the message, except the hash field which MUST always be present (see below for how the hash field is used).

`h.Sum(nil)` is a fine value to indicate an empty config.

cpheps: @tigrannajaryan updated. I still had it shortcut around to not cause a panic in the case of a nil configMap

client/internal/clientcommon.go
```diff
@@ -191,12 +192,28 @@ func (c *ClientCommon) SetAgentDescription(descr *protobufs.AgentDescription) er
...
+	// Construct hash
 	h := sha256.New()
-	if msg.ConfigMap != nil {
-		for k, v := range msg.ConfigMap.ConfigMap {
-			h.Write([]byte(k))
-			h.Write(v.Body)
-			h.Write([]byte(v.ContentType))
+
+	// If the config is empty don't attemp to add more to the hash
+	if cfgMap != nil || len(cfgMap) > 0 {
```
pmm-sumo, line 201: So if I see correctly, `cfgMap` is `map[string]*protobufs.AgentConfigFile{}`, which means you don't need to check for nil, i.e.

```suggestion
	if len(cfgMap) > 0 {
```

client/internal/clientcommon.go
```diff
@@ -191,12 +192,28 @@ func (c *ClientCommon) SetAgentDescription(descr *protobufs.AgentDescription) er
 // calcHashEffectiveConfig calculates and sets the Hash field from the rest of the
 // fields in the message.
 func calcHashEffectiveConfig(msg *protobufs.EffectiveConfig) {
+	cfgMap := msg.GetConfigMap().GetConfigMap()
```
pmm-sumo, line 195: I think `msg.GetConfigMap()` can return nil and this needs to be checked before calling the other `GetConfigMap()`

andykellr: Because of the way protobuf code is generated, this chaining is safe.

```go
func (x *AgentConfigMap) GetConfigMap() map[string]*AgentConfigFile {
	if x != nil {
		return x.ConfigMap
	}
	return nil
}
```

pmm-sumo: Oh, you're right. TBH, I wasn't aware this can be pulled with GoLang. Thanks!

client/internal/clientcommon.go
cpheps: Yeah you're right. I forgot the `len` returns 0 for nil or empty.

client/internal/clientcommon.go
cpheps: The generate proto code actually does a nil check on the pointer receiver so it'll just return nil and not panic.

https://github.com/open-telemetry/opamp-go/blob/main/protobufs/opamp.pb.go#L2463-L2468

cpheps: My first test case actually has a nil `AgentConfigMap` so it tests this case.

tigrannajaryan (approved):
Thanks you @cpheps

PR #76, opened 2022-05-17
Implement package syncing

tigrannajaryan:
- This implements "Packages" spec section https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#packages-1
- TODO: increase test coverage to handle many more edge cases.
- TODO: add package usage to the example Agent/Server.

### DIFF
--- a/client/clientimpl_test.go
+++ b/client/clientimpl_test.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"math/rand"
 	"net/http"
+	"net/http/httptest"
 	"net/url"
 	"sync/atomic"
 	"testing"
@@ -391,6 +392,7 @@ func TestIncludesDetailsOnReconnect(t *testing.T) {
 	eventually(t, func() bool { return atomic.LoadInt64(&receivedDetails) == 1 })
 
 	// close the Agent connection. expect it to reconnect and send details again.
+	require.NotNil(t, client.conn)
 	err := client.conn.Close()
 	assert.NoError(t, err)
 
@@ -438,7 +440,7 @@ func TestSetEffectiveConfig(t *testing.T) {
 		settings.OpAMPServerURL = "ws://" + srv.Endpoint
 		prepareClient(t, &settings, client)
 
-		assert.NoError(t, client.Start(context.Background(), settings))
+		require.NoError(t, client.Start(context.Background(), settings))
 
 		// Verify config is delivered.
 		eventually(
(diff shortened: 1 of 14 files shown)


client/internal/packagessyncer.go
```diff
@@ -0,0 +1,327 @@
...
+			if err != nil {
+				lastErr = err
+			}
+		}
+	}
+
+	// Also remove packages that were not offered from the statuses.
+	for name := range s.statuses.Packages {
+		if _, offered := s.available.Packages[name]; !offered {
+			delete(s.available.Packages, name)
```
andykellr, line 299:
```suggestion
			delete(s.statuses.Packages, name)
```

andykellr:
Aside from the small issue I mentioned, overall this looks good. However, unless I missed something, it appears to me that it is possible to have multiple sync operations running at the same time which could be confusing.

tigrannajaryan:
>However, unless I missed something, it appears to me that it is possible to have multiple sync operations running at the same time which could be confusing.

Good point. I will create an issue to fix this in a future PR.

client/internal/packagessyncer.go
tigrannajaryan: @andykellr fixed. PTAL.

client/clientimpl_test.go
```diff
@@ -966,3 +968,253 @@ func TestRemoteConfigUpdate(t *testing.T) {
...
+		srv := internal.StartMockServer(t)
+		srv.EnableExpectMode()
+
+		localPackageState := internal.NewInMemPackagesStore()
+
+		var syncerDoneCh <-chan struct{}
+
+		// Prepare a callback that returns either success or failure.
+		onPackagesAvailable := func(ctx context.Context, packages *protobufs.PackagesAvailable, syncer types.PackagesSyncer) error {
+			if testCase.errorOnCallback {
```
pmm-sumo, line 995: Why the error is returned when the flag is false rather than true?

tigrannajaryan: Oops. Fixed.

client/internal/httpsender.go
```diff
@@ -97,6 +103,10 @@ func (h *HTTPSender) makeOneRequestRoundtrip(ctx context.Context) {
 	if err != nil {
 		return
```
pmm-sumo, line 104: not scope of this PR but just noticed this - shouldn't we at least log the error?

pmm-sumo (approved):

kpratyus:
In the file packagesyncer.go defer func() closes the channel s.doneCh but it doesn't necessarily wait for the subroutine go 
s.doSync to get completed so wouldn't it be giving a wrong sync status?

```
func (s *packagesSyncer) Sync(ctx context.Context) error {	
         defer func() {
		close(s.doneCh)
	}()'

	// Prepare package statuses.
	if err := s.initStatuses(); err != nil {
		return err
	}

	if err := s.clientSyncedState.SetPackageStatuses(s.statuses); err != nil {
		return err
	}

	// Now do the actual syncing in the background.
	go s.doSync(ctx, s.packageSyncComplete)

	return nil
}
```

PR #81, opened 2022-05-20
Add Header and remove AuthorizationHeader to allow any HTTP header to be specified

### LINKED ISSUE #79: Allow the agent User-Agent header to be configurable
I would like to be able to control the User-Agent header of the agent when communicating with the management server. Currently this defaults to `Go-http-client/1.1`. It can be useful to identify traffic for logging.

I would propose adding a `UserAgent string` field to `StartSettings` and setting the User-Agent header when we set the Authorization header.

andykellr:
Resolves #79


tigrannajaryan (approved):
LGTM

PR #93, opened 2022-06-29
Simplify status compression

tigrannajaryan:
- Eliminate all "hash" fields.
- Allow omitting the sub-messages of AgentToServer message. When omitted it is implied that previously reported value of the sub-message is current (unchanged).
- To detect lost messages have one auto-incremented sequence_num field AgentToServer message. Server can easily detect losses by just keeping the last sequence_num (as opposed to keeping 4 different hashes).


andykellr:
Actually, doing a search for `Hash` I found a couple more places that should be updated. I'll add suggestions.

client/internal/clientcommon.go
```diff
@@ -310,15 +270,14 @@ func (c *ClientCommon) SetPackageStatuses(statuses *protobufs.PackageStatuses) e
...
 
-	// Get the hash of the status before we update it.
-	prevHash := c.ClientSyncedState.PackageStatuses().GetHash()
+	statusChanged := !proto.Equal(c.ClientSyncedState.PackageStatuses(), statuses)
 
 	if err := c.ClientSyncedState.SetPackageStatuses(statuses); err != nil {
 		return err
 	}
 
 	// Check if the new status is different from the previous by comparing the hashes.
```
andykellr, line 279:
```suggestion
	// Check if the new status is different from the previous.
```
```suggestion
	// Check if the new status is different from the previous by comparing the hashes.
```

andykellr:
The following have comments that refer to the Hash field:
* client/types/startsettings.go
* client/types/callbacks.go
* client/client.go

tigrannajaryan:
> The following have comments that refer to the Hash field:
> 
> * client/types/startsettings.go
> * client/types/callbacks.go
> * client/client.go

Thanks for the catch. I did a pass over this files. Please check again.

client/internal/clientcommon.go
tigrannajaryan: Fixed.

PR #103, opened 2022-07-15
Add "fully set" flag to AgentToServer message

tigrannajaryan:
This is a draft PR to demonstrate how https://github.com/open-telemetry/opamp-spec/issues/109
can be resolved.

The new StatusIsFullSet is set by the agent the status is fully set the message.
The server checks the flag to know if it needs to request the full status.

This is a backward compatible change. New servers that work with old agents may
make an extra request for the full status because old agents don't set this flag,
but this was already a problem in the past even before this change.


tigrannajaryan:

tigrannajaryan:
Closing in favour of https://github.com/open-telemetry/opamp-go/pull/105

PR #105, opened 2022-07-15
Correctly handle status compression on the server side

tigrannajaryan:
We check the omitted field and the corresponding capability bit to understand
if the compression was used.

Related to spec issue https://github.com/open-telemetry/opamp-spec/issues/109
The issue was discarded in favour of the implementation that does not
change the spec, but instead uses already existing information.

### DIFF
--- a/client/internal/clientcommon.go
+++ b/client/internal/clientcommon.go
@@ -206,12 +206,7 @@ func (c *ClientCommon) PrepareFirstMessage(ctx context.Context) error {
 			msg.EffectiveConfig = cfg
 			msg.RemoteConfigStatus = c.ClientSyncedState.RemoteConfigStatus()
 			msg.PackageStatuses = c.ClientSyncedState.PackageStatuses()
-
-			if c.PackagesStateProvider != nil {
-				// We have a state provider, so package related capabilities can work.
-				msg.Capabilities |= protobufs.AgentCapabilities_AcceptsPackages
-				msg.Capabilities |= protobufs.AgentCapabilities_ReportsPackageStatuses
-			}
+			msg.Capabilities = c.Capabilities
 		},
 	)
 	return nil
--- a/internal/examples/server/data/agent.go
+++ b/internal/examples/server/data/agent.go
@@ -194,16 +194,34 @@ func (agent *Agent) updateEffectiveConfig(
 	}
 }
 
+func (agent *Agent) hasCapability(capability protobufs.AgentCapabilities) bool {
+	return agent.Status.Capabilities&capability != 0
+}
+
 func (agent *Agent) processStatusUpdate(
 	newStatus *protobufs.AgentToServer,
 	response *protobufs.ServerToAgent,
 ) {
-	if agent.Status != nil && agent.Status.SequenceNum+1 != newStatus.SequenceNum {
-		// We lost the previous status update. Request full status update from the agent.
-		response.Flags |= protobufs.ServerToAgent_ReportFullState
-	}
+	// We don't have any status for this Agent, or we lost the previous status update from the Agent, so our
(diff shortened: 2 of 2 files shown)


internal/proto/opamp.proto
```diff
@@ -538,6 +538,10 @@ enum AgentCapabilities {
     AcceptsOtherConnectionSettings = 0x00000200;
     // The Agent can accept restart requests.
     AcceptsRestartCommand          = 0x00000400;
+    // The Agent reports Health
+    ReportsHealth                  = 0x00000800;
+    //    // The Agent will report RemoteConfig in AgentToServer.
+    //    ReportsRemoteConfig            = 0x00001000;
```
tigrannajaryan, line 544: We should consider adding this to be consistent in Accepts/Reports pairing of capabilities.

tigrannajaryan:
This needs to be rebased after https://github.com/open-telemetry/opamp-go/pull/110 is merged and then it can be reviewed.

client/internal/clientcommon.go
```diff
@@ -206,12 +206,7 @@ func (c *ClientCommon) PrepareFirstMessage(ctx context.Context) error {
 			msg.EffectiveConfig = cfg
 			msg.RemoteConfigStatus = c.ClientSyncedState.RemoteConfigStatus()
 			msg.PackageStatuses = c.ClientSyncedState.PackageStatuses()
-
-			if c.PackagesStateProvider != nil {
-				// We have a state provider, so package related capabilities can work.
-				msg.Capabilities |= protobufs.AgentCapabilities_AcceptsPackages
-				msg.Capabilities |= protobufs.AgentCapabilities_ReportsPackageStatuses
-			}
+			msg.Capabilities = c.Capabilities
```
tigrannajaryan, line 209: This ensures that the agent correctly reports the capabilities. I can extract this fix into a separate PR, but probably not worth it since it only became visible in this PR.

tigrannajaryan:
@andykellr this is now ready for review.

andykellr (approved):

internal/examples/server/data/agent.go
```diff
@@ -194,17 +194,35 @@ func (agent *Agent) updateEffectiveConfig(
...
+	// Check if any fields were omitted in the status report.
+	effectiveConfigOmitted := agent.hasCapability(protobufs.AgentCapabilities_ReportsEffectiveConfig) && newStatus.EffectiveConfig == nil
+	packageStatusesOmitted := agent.hasCapability(protobufs.AgentCapabilities_ReportsPackageStatuses) && newStatus.PackageStatuses == nil
+	remoteConfigStatusOmitted := agent.hasCapability(protobufs.AgentCapabilities_AcceptsRemoteConfig) && newStatus.RemoteConfigStatus == nil
+	healthOmitted := agent.hasCapability(protobufs.AgentCapabilities_ReportsHealth) && newStatus.Health == nil
+
+	// True if the status was not fully reported.
+	statusIsCompressed := effectiveConfigOmitted || packageStatusesOmitted || remoteConfigStatusOmitted || healthOmitted
+
+	if statusIsCompressed && lostPreviousUpdate {
```
PeterF778, line 220: Should it be OR?

tigrannajaryan: No, "AND" is correct here. We only ned to ask for the full state if both of these conditions are true:
- The full state was not reported in the message we have just received. If this is not true, it means we have already just received the full state, there is no need to ask for it again.
- We detected that we lost a previous update. If this is not true, it means we already have the correct state on our end, there is no need to ask the client to send it again.

PeterF778: Got it, thanks!

Issue #100, opened 2022-07-14
Make sure all exporter funcs and types have docs


tigrannajaryan:
Many are currently missing doc comments, e.g. NewHTTP.

mhausenblas:
@tigrannajaryan I’d be interested to take a stab at this, seems like a good way to familiarise oneself with the code base.

tigrannajaryan:
@mhausenblas great, assigned to you.

PR #107, opened 2022-07-16
Update client package docs

srikanthccv:
Partially addresses #100. Also added to some internal funcs which I thought would be helpful to get a quick summary.

### DIFF
--- a/client/httpclient.go
+++ b/client/httpclient.go
@@ -34,7 +34,7 @@ func NewHTTP(logger types.Logger) *httpClient {
 	return w
 }
 
-// Start starts the client and runs until Stop is called.
+// Start implements OpAMPClient.Start.
 func (c *httpClient) Start(ctx context.Context, settings types.StartSettings) error {
 	if err := c.common.PrepareStart(ctx, settings); err != nil {
 		return err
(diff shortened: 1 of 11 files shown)


client/httpclient.go
```diff
@@ -33,6 +34,7 @@ func NewHTTP(logger types.Logger) *httpClient {
 	return w
 }
 
+// Start starts the client and runs until Stop is called.
```
tigrannajaryan, line 37: `Start()` is explained in more detail in the `OpAMPClient` interface description. We can either copy the comment from there or reference from here.

tigrannajaryan (commented):
Great comments, thanks a lot. I have just one minor suggestion.

client/types/callbacks.go
```diff
@@ -39,6 +39,8 @@ type MessageData struct {
 }
 
 type Callbacks interface {
+	// OnConnect is called when the connection is successfully established to the Server.
+	// May be called after Start() is called and every time connected to the Server.
```
andykellr, line 43:
```suggestion
	// May be called after Start() is called and every time a connection is established to the Server.
```

andykellr: Should we add more description here for websocket and http?

For example:

WebSocket clients will call OnConnect when a new WebSocket connection is established.

HTTP clients will call OnConnect after any request that results in a 200 response.

client/httpclient.go
srikanthccv: Right, I wasn't really sure about this. What do you think about remaining ones `SetAgentDescription`, `SetHealth`? Should we copy for all of them?

client/types/callbacks.go
srikanthccv: Good suggestion. I will update the description.

client/httpclient.go
tigrannajaryan: If you can add a reference in a way that it becomes a link in godoc [here](https://pkg.go.dev/github.com/open-telemetry/opamp-go) that would be best, so that we don't have to duplicate the content and risk diverging.

srikanthccv: Looks like there is no way to achieve this. I looked up how the packages in stdlib cross reference the functions from other modules. It just uses plain text like `os.Exit` or `json.Marshal` for the same reason.

tigrannajaryan: OK, let's just put plain text to reference `OpAMPClient.Start`.

tigrannajaryan (approved):
Thanks @srikanthccv

Issue #99, opened 2022-07-14
Improve test coverage


tigrannajaryan:
The test coverage is inadequate. We need to increase it, particularly add coverage for edge cases and erroneous situations.

srikanthccv:
@tigrannajaryan I can pick this up

tigrannajaryan:
@srikanthccv great, assigned to you.

azimmsd:
Hi @tigrannajaryan ,
Is the issue still open?

I can pick it up.

PR #113, opened 2022-07-27
Don't call OnConnectFailed when failed to connect during shutdown

### LINKED ISSUE #111: Websocket Race Condition when closing
There appears to be a race condition when calling `Stop()` on a websocket client. 

The [runUntilStopped](https://github.com/open-telemetry/opamp-go/blob/main/client/wsclient.go#L256-L265) function checks `IsStopping()` before `runOneCycle`. Inside `runOneCycle` we attempt to connect via [ensureConnected](https://github.com/open-telemetry/opamp-go/blob/main/client/wsclient.go#L200) then check `IsStopping()` again. The race condition lies in if `Stop()` is called on the client after the first `IsStopping()` check in `runUntilStopped` but before the second one in `runOneCycle`. 

If this happens `ensureConnected` could possible fail due to a closed connection. Which is looks like we account for in [this](https://github.com/open-telemetry/opamp-go/blob/main/client/wsclient.go#L201-L203) comment. The problem is on a failed connection the `OnConnectFailedHandler` is called and could execute any code associated with that on the implementors side. This is not ideal in a graceful shutdown.

I'm wondering if it's a better idea to swap the `IsStopping()` check and `ensureConnected` inside `runOnceCycle`. This way we don't try to connect if we're currently stopping.

```go
	if c.common.IsStopping() {
		_ = c.conn.Close()
		return
	}

	if err := c.ensureConnected(ctx); err != nil {
                // Maybe log error here
		return
	}
```

cpheps:
Fixes #111

Once `Stop()` is called on the websocket client we don't call `OnConnectFailed` handler if a dial fails.

Tested this fix on a system that was repeatedly showing the `OnConnectFailed` handler called during shutdown.

### DIFF
--- a/client/wsclient.go
+++ b/client/wsclient.go
@@ -121,7 +121,7 @@ func (c *wsClient) tryConnectOnce(ctx context.Context) (err error, retryAfter sh
 	var resp *http.Response
 	conn, resp, err := c.dialer.DialContext(ctx, c.url.String(), c.requestHeader)
 	if err != nil {
-		if c.common.Callbacks != nil {
+		if c.common.Callbacks != nil && !c.common.IsStopping() {
 			c.common.Callbacks.OnConnectFailed(err)
 		}
 		if resp != nil {


tigrannajaryan (commented):
@cpheps can you add a test that creates the race condition and fails without this fix, but passes after the fix?

cpheps:
@tigrannajaryan I can try. I'm not sure how possible it is since you need to Stop the opamp client at certain point. I don't know if it's possible right now to force the client to block so we can hit that point.

tigrannajaryan:
> @tigrannajaryan I can try. I'm not sure how possible it is since you need to Stop the opamp client at certain point. I don't know if it's possible right now to force the client to block so we can hit that point.

OK, please try. Ideally we want to clearly demonstrate that the PR really fixes the problem and an automated test helps a lot.

If you find that it is impossible we can merge as is.

cpheps:
@tigrannajaryan I've been messing around with this and I can't seem to get it consistent. I was able to write a test that triggered the race condition like 1 in 5 runs. The race conditions lies inside the `runUntilStopped` which runs concurrently on start. We have no injection point currently to have it hang. 

The test I have was starting the client, waiting a few milliseconds, stopping it, the checking if `OnConnectFailedFunc` was called. This is probably a good test just in general but it doesn't isolate/prove the race condition was solved.

There's only one other test for the websocket implementation that shows `OnConnectFailedFunc` is called when the sever closes the connection. So maybe a test showing we don't call `OnConnectFailedFunc` on a client connection close is good, but like I said doesn't highly the race condition fix.

tigrannajaryan:
OK, let's merge this as is and maybe we can add more tests as part of https://github.com/open-telemetry/opamp-go/issues/99

tigrannajaryan (approved):

PR #114, opened 2022-07-28
Use ReportsRemoteConfig capability

tigrannajaryan:
Implements spec change https://github.com/open-telemetry/opamp-spec/pull/115


client/internal/clientcommon.go
```diff
@@ -293,6 +294,10 @@ func (c *ClientCommon) UpdateEffectiveConfig(ctx context.Context) error {
 // It also remembers the new RemoteConfigStatus in the client state so that it can be
 // sent to the Server when the Server asks for it.
 func (c *ClientCommon) SetRemoteConfigStatus(status *protobufs.RemoteConfigStatus) error {
+	if c.Capabilities&protobufs.AgentCapabilities_ReportsRemoteConfig == 0 {
```
srikanthccv, line 297: Probably a naive question, Does this mean client start fails if the `ReportsRemoteConfig` capability not enabled? 

https://github.com/open-telemetry/opamp-go/blob/3d4cbdc975d341c1765050a9a59882ef89dda006/client/internal/clientcommon.go#L75-L98

tigrannajaryan: No, the start does not fail. It fails when SetRemoteConfigStatus is called, which is supposed to be called after Start:
```
	// SetRemoteConfigStatus sets the current RemoteConfigStatus.
	// LastRemoteConfigHash field must be non-nil.
	// May be called anytime after Start(), including from OnMessage handler.
```

srikanthccv: Ah thanks, I am confusing `ClientSyncedState.SetRemoteConfigStatus` with  `ClientCommon.SetRemoteConfigStatus`.

srikanthccv (approved):

andykellr (approved):

PR #115, opened 2022-07-29
Consistently declare Protobuf enums

tigrannajaryan:
Enums were not consistently declared. Some were at the global scope,
some others wire nested in messages. This moves all enums to the global
scope and uses most of the official style guide rules here:
https://developers.google.com/protocol-buffers/docs/style#enums

Unfortunately this results in long and stuttering names of enum values in
the code but this change is necessary to future-proof the proto and ensure
no enum names conflict when we add new enums and values (all enum value
names are in one namespace).

I don't like the end result but not doing it appears to be worse.
We have seen this first-hand in OTLP proto where enum names are
inconsistent and are now impossible to fix due to backward compatibility
concerns.

Note: I ignored the rule to CAPITALIZE names and use underscores
because if I do that the names in the code become unreadable walls
of text.

If anyone has any ideas how to fix the enum names issue in a nicer
way I am happy to discard this PR.

One alternate is to remove the prefixes from enum value names and
just require that all enum value names introduced in the future are
also unique, but I think it can result in weird names in the future.


internal/proto/opamp.proto
```diff
@@ -450,21 +440,38 @@ message ServerErrorResponse {
...
+    // The AgentToServer message was malformed. The Agent SHOULD NOT retry
+    // the message.
+    ServerErrorResponse_BadRequest = 1;
+
+    // The Server is overloaded and unable to process the request. The Agent
+    // should retry the message later. retry_info field may be optionally
+    // set with additional information about retrying.
+    ServerErrorResponse_Unavailable = 2;
+}
+
```
andykellr, line 458: Should these include Type, e.g. `ServerErrorResponseType_Unknown`, to match the name of the enum?

tigrannajaryan: Yes, let me fix.

tigrannajaryan: Fixed.

internal/proto/opamp.proto
```diff
@@ -660,32 +668,33 @@ message PackageStatus {
...
-    }
-    Status status = 6;
+    PackageStatusEnum status = 6;
 
     // Error message if the status is erroneous.
     string error_message = 7;
 }
 
+// The status of this package.
+enum PackageStatusEnum {
```
andykellr, line 678: nit: Should we call this PackageStatuses to match other similar enum names? This is the only one with an Enum suffix.

tigrannajaryan: Unfortunately I can't because we already have a message named `PackageStatuses`. This was a forced deviation.

andykellr (approved):

tigrannajaryan:
Rebased and fixed merge conflicts. @andykellr if you have time please take another look (just want to make sure I didn't miss anything), otherwise I will merge soon.

PR #118, opened 2022-07-29
Package Syncing

kpratyus:
Making sure that only one package syncing operation happens at a time.

### DIFF
--- a/client/internal/clientstate.go
+++ b/client/internal/clientstate.go
@@ -32,6 +32,7 @@ var (
 // It is safe to call methods of this struct concurrently.
 type ClientSyncedState struct {
 	mutex sync.Mutex
+	packageSyncMutex sync.Mutex
 
 	agentDescription   *protobufs.AgentDescription
 	health             *protobufs.AgentHealth
--- a/client/internal/packagessyncer.go
+++ b/client/internal/packagessyncer.go
@@ -58,7 +58,7 @@ func (s *packagesSyncer) Sync(ctx context.Context) error {
 	}
 
 	// Now do the actual syncing in the background.
-	go s.doSync(ctx)
+	go s.doSync(ctx, &s.clientSyncedState.packageSyncMutex)
 
 	return nil
 }
@@ -98,7 +98,7 @@ func (s *packagesSyncer) initStatuses() error {
 }
 
 // doSync performs the actual syncing process.
-func (s *packagesSyncer) doSync(ctx context.Context) {
+func (s *packagesSyncer) doSync(ctx context.Context, mutex *sync.Mutex) {
 	hash, err := s.localState.AllPackagesHash()
 	if err != nil {
 		s.logger.Errorf("Package syncing failed: %V", err)
@@ -108,6 +108,9 @@ func (s *packagesSyncer) doSync(ctx context.Context) {
 		s.logger.Debugf("All packages are already up to date.")
 		return
 	}
+
+	(*mutex).Lock()
+	defer (*mutex).Unlock()
 
 	failed := false
 	if err := s.deleteUnneededLocalPackages(); err != nil {


client/internal/clientstate.go
```diff
@@ -32,6 +32,7 @@ var (
 // It is safe to call methods of this struct concurrently.
 type ClientSyncedState struct {
 	mutex sync.Mutex
+	packageSyncMutex sync.Mutex
```
tigrannajaryan, line 35: I don't think this is the right place for this mutex. The ClientSyncedState struct has a specific purpose that is described above and adding a mutex for packager syncer does not fit that purpose.

`receivedProcessor` may be a better place for this mutex.

client/internal/packagessyncer.go
```diff
@@ -98,7 +99,7 @@ func (s *packagesSyncer) initStatuses() error {
 }
 
 // doSync performs the actual syncing process.
-func (s *packagesSyncer) doSync(ctx context.Context) {
+func (s *packagesSyncer) doSync(ctx context.Context, mutex *sync.Mutex) {
 	hash, err := s.localState.AllPackagesHash()
```
tigrannajaryan, line 103: Why isn't the mutex locked right from the top of doSync? `AllPackagesHash` is also susceptible to the race. It reads data that is written by `SetAllPackagesHash` below.

tigrannajaryan (commented):
@kpratyus Thanks for the PR!
Is it possible to add a test that tries to race (e.g. quickly return 2 successive PackagesAvailable offers from the Server to trigger 2 concurrent sync operations) and demonstrates that this fix works correctly?

client/internal/packagessyncer.go
kpratyus: Ah, Missed that. I actually thought that it only handles logging. Will change the mutex. Thanks for pointing it out

client/internal/clientstate.go
kpratyus: The  `receivedProcessor` is constructed in the constructor of `wsReceiver`. As the mutex has to be unique for each client and the pointer has to be passed to `receivedProcessor'. So mutex has to be included in the constructor of client.common, wsReceiver and receivedProcessor. To avoid these many changes I added the mutex in ClientSyncedState. What's your opinion on this?

kpratyus:
> @kpratyus Thanks for the PR!
> Is it possible to add a test that tries to race (e.g. quickly return 2 successive PackagesAvailable offers from the Server to trigger 2 concurrent sync operations) and demonstrates that this fix works correctly?

I will try to add this test

client/internal/clientstate.go
tigrannajaryan: There is one-to-one relationship between Client, ClientSyncedState and receivedProcessor, so I don't understand where do you see a problem. A mutex declared in receivedProcessor struct will be unique for each client, just like it is unique when it is in ClientSyncedState. There is no need to pass the mutex to receivedProcessor, receivedProcessor can contain the mutex.

kpratyus: Correct me if I am wrong here. For every client The function `runOneCycle` runs in a loop. Everytime the function runs it creates a new receiver object.
`r := internal.NewWSReceiver(
		c.common.Logger,
		c.common.Callbacks,
		c.conn,
		c.sender,
		&c.common.ClientSyncedState,
		c.common.PackagesStateProvider,
		c.common.Capabilities,
	)`
This in turn creates a new `receivedProcessor' object everytime. 
`processor: newReceivedProcessor(logger, callbacks, sender, clientSyncedState, packagesStateProvider, capabilities)`
So how is there a one to one relationship between client and receivedProcessor?

tigrannajaryan: You are absolutely right and I was wrong! Due to reconnections receivedProcessor may be re-created while a previous sync is still happening. So, receivedProcessor is the wrong owner for the mutex, you are right.

I still don't like putting it in ClientSyncedState, it doesn't fit the purpose.

Here are a couple options:
- Extend the purpose of ClientSyncedState to contain all of the client state. In that case the mutex can be a good fit there. It probably needs to be renamed to ClientState. I am not entirely sure this is a good idea though, since we are not really putting the entire client state in it, so it may become misleading.
- Another good owner for the mutex is ClientCommon. However, this will require passing the mutex where receivedProcessor is created and store the pointer to the mutex in receivedProcessor. Probably a bit awkward but doable.

Maybe there is another option that I haven't thought of, please feel free to suggest.

kpratyus: While raising the PR I also thought of these 2 options. Code clarity wise adding all the mutexes in a single structure(ClientSyncedState here) makes sense but I totally get the point that we want to keep ClientSyncedState as lightweight as possible. 

Although awkward I think keeping it in ClientCommon is a better option. In future if there arises need of any other locks then it would be totally extendable as well.

tigrannajaryan: >Although awkward I think keeping it in ClientCommon is a better option. In future if there arises need of any other locks then it would be totally extendable as well.

Sounds good, let's go with this approach.

PR #125, opened 2022-08-07
Implement gzip compression for HTTP client

### LINKED ISSUE #112: Implement gzip compression for HTTP transport
The [spec says](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#plain-http-transport):

>The Agent MAY compress the request body using gzip method and MUST specify "Content-Encoding: gzip" in that case. Server implementations MUST honour the "Content-Encoding" header and MUST support gzipped or uncompressed request bodies.
>The Server MAY compress the response if the Agent indicated it can accept compressed response via the "Accept-Encoding" header.

srikanthccv:
Resolves #112

### DIFF
--- a/client/httpclient.go
+++ b/client/httpclient.go
@@ -42,10 +42,12 @@ func (c *httpClient) Start(ctx context.Context, settings types.StartSettings) er
 
 	c.opAMPServerURL = settings.OpAMPServerURL
 
-	// TODO: implement compression based on settings.EnableCompression value.
-
 	// Prepare Server connection settings.
 	c.sender.SetRequestHeader(settings.Header)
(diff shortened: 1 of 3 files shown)


tigrannajaryan:
@srikanthccv thanks for the PR. Can you please implement the server-side gzip support first? Ideally we want to merge that first before merging this PR, otherwise someone may try to enable client-side compression and it won't work because server doesn't support it yet.

srikanthccv:
@tigrannajaryan PTAL https://github.com/open-telemetry/opamp-go/pull/126

tigrannajaryan (approved):

PR #127, opened 2022-09-05
fix(supervisor-example): fix agent initial configurations

nemoshlag:
A suggested fix to issue #128


tigrannajaryan:
The issue is closed as non-repro: https://github.com/open-telemetry/opamp-go/issues/128

PR #129, opened 2022-09-13
Add Disconnect() method to client connections

### LINKED ISSUE #91: Add Close() to types.Connection to allow the server to close a connection
It can be useful to close an agent connection to the server and I propose that a Close() method be added to the [types.Connection](https://github.com/open-telemetry/opamp-go/blob/main/server/types/connection.go) interface.

The agent may quickly reconnect, but in a server cluster with multiple nodes, periodically closing connections can help with rebalancing the cluster.

Consider the case of 100 agents connected to a 2-node cluster and assume a load balancer splits them 50/50 between the 2 nodes. If one node is terminated, 50 agents will reconnect and now all 100 agents will be connected to a single node. If a node is added, there will still be 100 agents connected to a single node and the new node will have 0 agents.

nemoshlag:
Resolves #91 

This enables OpAMP to close the network connection to an agent from the server side

### DIFF
--- a/server/httpconnection.go
+++ b/server/httpconnection.go
@@ -9,7 +9,7 @@ import (
 	"github.com/open-telemetry/opamp-go/server/types"
 )
 
-var errInvalidHTTPConnection = errors.New("cannot Send() over HTTP connection")
+var errInvalidHTTPConnection = errors.New("cannot operate over HTTP connection")
 
 // httpConnection represents an OpAMP connection over a plain HTTP connection.
 // Only one response is possible to send when using plain HTTP connection
@@ -31,6 +31,7 @@ func (c httpConnection) Send(_ context.Context, _ *protobufs.ServerToAgent) erro
 	return errInvalidHTTPConnection
 }
 
-func (c httpConnection) Close() error {
-	return c.conn.Close()
+func (c httpConnection) Disconnect() error {
+	// Disconnect() should not be called for plain HTTP connection.
+	return errInvalidHTTPConnection
 }
--- a/server/httpconnection.go
+++ b/server/httpconnection.go
@@ -9,7 +9,11 @@ import (
 	"github.com/open-telemetry/opamp-go/server/types"
 )
 
-var errInvalidHTTPConnection = errors.New("cannot operate over HTTP connection")
+// ErrInvalidHTTPConnection represents an event of misuse function for plain HTTP
+// connection, such as httpConnection.Send() or httpConnection.Disconnect().
+// Usage will not result with change but return this error to indicate current state
+// might not be as expected.
+var ErrInvalidHTTPConnection = errors.New("cannot operate over HTTP connection")
 
 // httpConnection represents an OpAMP connection over a plain HTTP connection.
 // Only one response is possible to send when using plain HTTP connection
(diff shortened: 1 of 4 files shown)


tigrannajaryan (commented):
Thank you for the PR @nemoshlag

server/serverimpl_test.go
```diff
@@ -126,6 +126,36 @@ func TestServerStartAcceptConnection(t *testing.T) {
...
+
+	// Close connection from server side
+	srvConn := wsConnection{wsConn: conn}
+	err = srvConn.Close()
+
+	assert.NoError(t, err)
+
+	// Verify that the Server receives the closing notification.
+	eventually(t, func() bool { return atomic.LoadInt32(&connectionCloseCalled) == 1 })
+}
```
tigrannajaryan, line 157: Can we also verify that the connection is closed from the Client's perspective?

server/serverimpl_test.go
```diff
@@ -126,6 +126,36 @@ func TestServerStartAcceptConnection(t *testing.T) {
...
+	// Connect to the Server.
+	conn, _, err := dialClient(settings)
+
+	// Verify that the connection is successful.
+	assert.NoError(t, err)
+	assert.True(t, atomic.LoadInt32(&connectionCloseCalled) == 0)
+
+	// Close connection from server side
+	srvConn := wsConnection{wsConn: conn}
+	err = srvConn.Close()
```
tigrannajaryan, line 151: Since we have an implementation of Close() for `httpConnection`, is it possible to test that too?

server/serverimpl_test.go
nemoshlag: Added a check from the connection itself on lines 155-157 which I'm debating on, do you think it's ok?

tigrannajaryan:
I think Disconnect() is a better name and it is fine to make it an error to call Disconnect() for plain http connections.

andykellr:
> I think Disconnect() is a better name and it is fine to make it an error to call Disconnect() for plain http connections.

I was going to suggest that it be a no-op, but an error sound good. I would suggest creating an error like `ErrDisconnectHttpConnection` or something similar that can be easily checked and ignored by the caller in the event that the calling code is unaware of the type of connection be used.

tigrannajaryan:
> > I think Disconnect() is a better name and it is fine to make it an error to call Disconnect() for plain http connections.
> 
> I was going to suggest that it be a no-op, but an error sound good. I would suggest creating an error like `ErrDisconnectHttpConnection` or something similar that can be easily checked and ignored by the caller in the event that the calling code is unaware of the type of connection be used.

Yes, sorry, I want's clear. I meant it is good to return an error from Disconnect(), wasn't suggesting that it panics or anything.

tigrannajaryan (approved):

tigrannajaryan:
@andykellr please take another look and if it looks good please merge.

server/httpconnection.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/open-telemetry/opamp-go/server/types"
 )
 
-var errInvalidHTTPConnection = errors.New("cannot Send() over HTTP connection")
+var errInvalidHTTPConnection = errors.New("cannot operate over HTTP connection")
```
andykellr, line 12: Can we export this error so that callers of Disconnect() can check for this?

tigrannajaryan:
Merged. Thanks @nemoshlag!

PR #130, opened 2022-09-17
Add CustomMessage to AgentToServer and ServerToAgent

andykellr:
See https://github.com/open-telemetry/opamp-spec/issues/49

It was easiest for me to start here. If we agree on the direction of this, I plan to do the following: 

* update the opamp-spec
* add OnCustomMessage method to the client Callbacks interfaces
* add SendCustomMessage to the OpAMPClient
* add implementation to the client and server
* add tests for that implementation

### DIFF
--- a/client/types/callbacks.go
+++ b/client/types/callbacks.go
@@ -114,6 +114,12 @@ type Callbacks interface {
 
 	// OnCommand is called when the Server requests that the connected Agent perform a command.
 	OnCommand(command *protobufs.ServerToAgentCommand) error
+
+	// OnCustomMessage is called when either a CustomMessageResponse is sent from the server or a new CustomMessage
+	// message is sent from the server, or both. Either the response or the message will be non-nil. A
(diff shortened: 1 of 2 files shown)


andykellr:
@tigrannajaryan Let me know what you think of this direction.

andykellr:
I realized overnight that we should add a hash so it is clear that the response is appropriately associated with the last message.

I also stubbed the client callback, which I don't think is perfect, so feedback is welcome. Obviously none if this compiles because I added the method and didn't implement it anywhere. I also didn't generate the protobuf code.

tigrannajaryan:
Discussed 3 topics today:
- Describe uses cases to make the motivation clear
- Explain possible communication patterns (request/response only or can stream unidirectionally or bidirectionally?)
- Consider using sequence ids instead of hashes

andykellr:
Closing while spec is being finalized and discussion is happening in the spec repo.

PR #131, opened 2022-09-19
Add client-side support for datetime value of Retry-After header

### LINKED ISSUE #119: retry policy: parse datetime value for Retry-After header?
I started working on writing tests as part of #99. I noticed that the retry header only works with delay [seconds](https://github.com/open-telemetry/opamp-go/blob/main/internal/retryafter.go#L26). The header value can either be seconds or datatime. Should the datetime case be accounted for or assume the server always sends the seconds?

nemoshlag:
According to specifications, Retry-After header should be either seconds-delay or HTTP-date
https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.3

Resolves #119

### DIFF
--- a/internal/retryafter.go
+++ b/internal/retryafter.go
@@ -29,11 +29,11 @@ func ExtractRetryAfterHeader(resp *http.Response) OptionalDuration {
 				retryInterval := time.Duration(retryIntervalSec) * time.Second
 				return OptionalDuration{Defined: true, Duration: retryInterval}
 			}
-			// Try to parse retryAfterHTTPHeader as HTTP-date
+			// Try parse retryAfterHTTPHeader as HTTP-date
 			// See https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.3
-			t, err := time.Parse(time.RFC1123, retryAfter)
+			t, err := http.ParseTime(retryAfter)
 			if err == nil {
-				retryInterval := time.Until(t) * time.Second
+				retryInterval := time.Until(t)
 				return OptionalDuration{Defined: true, Duration: retryInterval}
 			}
 		}
--- a/internal/retryafter.go
+++ b/internal/retryafter.go
@@ -1,18 +1,43 @@
 package internal
 
 import (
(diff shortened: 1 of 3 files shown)


internal/retryafter.go
```diff
@@ -23,11 +23,19 @@ func ExtractRetryAfterHeader(resp *http.Response) OptionalDuration {
...
 		if retryAfter != "" {
+			// Try to parse retryAfterHTTPHeader as delay-seconds
 			retryIntervalSec, err := strconv.Atoi(retryAfter)
 			if err == nil {
 				retryInterval := time.Duration(retryIntervalSec) * time.Second
 				return OptionalDuration{Defined: true, Duration: retryInterval}
 			}
+			// Try to parse retryAfterHTTPHeader as HTTP-date
+			// See https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.3
+			t, err := time.Parse(time.RFC1123, retryAfter)
```
tigrannajaryan, line 34: The HTTP RFC7231 defines 3 different formats of HTTP-date that receivers MUST support. One of the formats is RFC5322. This uses RFC1123 which does not seem to be what RFC7231 requires.

internal/retryafter_test.go
```diff
@@ -0,0 +1,100 @@
...
+	resp := response503()
+	resp.StatusCode = http.StatusTooManyRequests
+	resp.Header.Add(retryAfterHTTPHeader, strconv.Itoa(retryIntervalSec))
+
+	expectedDuration := OptionalDuration{
+		Defined:  true,
+		Duration: time.Second * time.Duration(retryIntervalSec),
+	}
+	assert.Equal(t, expectedDuration, ExtractRetryAfterHeader(resp))
+}
```
tigrannajaryan, line 50: These 2 tests are almost identical. It would be nice to make them one test with 2 sub-tests. It will eliminate a lot of lines.

tigrannajaryan (approved):
Thank you @nemoshlag

PR #132, opened 2022-09-20
Change bitfields from enum to uint64

tigrannajaryan:
Implements spec change https://github.com/open-telemetry/opamp-spec/pull/125

In some languages enum fields are strongly typed and it is impossible to assigned OR-ed values to the enum field. This makes impossible to compose the bit fields properly.

This changes all bit field declarations from enum to uint64. The enum declarations stay, so that bit definitions are clear.


tigrannajaryan:
@andykellr can you please review so that we can merge this?

We want to [move the protos](https://github.com/open-telemetry/opamp-go/issues/133) and I want to make sure we don't have outstanding proto changes in this repo.

Issue #133, opened 2022-09-22
Move the protos to another repository


jlegoff:
As per our [discussion](https://cloud-native.slack.com/archives/C02J58HR58R/p1663781021774309) in slack, the protos should be moved to another repository so that the java implementation can use them.

tigrannajaryan:
I think we can move the proto files to a sub-directory in opamp-spec and consume opamp-spec as a submodule in this repo.

PR #134, opened 2022-09-22
Add opamp-spec as submodule to consume protos

tigrannajaryan:
Proto files were moved to opamp-spec. They are no longer necessary here. We now consume protos from opamp-spec repo.

Resolves https://github.com/open-telemetry/opamp-go/issues/133


andykellr (approved):
Looks good. I was able to update the submodule but I can't test gen-proto because of M1.

PR #138, opened 2022-10-20
Set telemetry resource attributes for Otel Collector Supervisor

tigrannajaryan:
The OpAMP spec requires that certain attributes are set in own telemetry, see here: https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agentdescriptionidentifying_attributes


tigrannajaryan:

PR #139, opened 2022-10-27
Implement health checks in Supervisor

tigrannajaryan:
The Supervisor now periodically polls Collector's health check endpoint and reports the health status to the OpAMP Server.

### DIFF
--- a/internal/examples/supervisor/supervisor/supervisor.go
+++ b/internal/examples/supervisor/supervisor/supervisor.go
@@ -2,6 +2,7 @@ package supervisor
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math/rand"
 	"os"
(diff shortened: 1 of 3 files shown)


internal/examples/supervisor/supervisor/supervisor.go
```diff
@@ -437,16 +497,18 @@ func (s *Supervisor) runAgentProcess() {
...
+
+		case <-s.healthCheckTicker.C:
+			s.healthCheck()
 		}
 	}
 }
 
-func (s *Supervisor) applyConfigWithAgentRestart() {
+func (s *Supervisor) stopAgentApplyConfig() {
 	s.logger.Debugf("Restarting the agent with the new config.")
```
andykellr, line 508: If this no longer performs a restart, we should change the Debug statement.
```suggestion
	s.logger.Debugf("Stopping the agent to apply new config.")
```

andykellr (approved):
One small suggestion in the debug comment, but looks good otherwise.

internal/examples/supervisor/supervisor/supervisor.go
tigrannajaryan: Fixed.

PR #142, opened 2022-10-28
Implement health checks in Supervisor

tigrannajaryan:
The Supervisor now periodically polls Collector's
health check endpoint and reports the health status to the OpAMP Server.


tigrannajaryan:
@andykellr please take a look again. Not sure how I managed to mess up with the main branch. Should be good now.

andykellr (approved):

PR #143, opened 2022-11-01
Add State interface{} to types.Connection to share state between callbacks

andykellr:
It can be useful to share state between opamp server callbacks. This PR adds `State interface{}` to types.ConnectionResponse and makes it available on the `type.Connection` via `State()`.

A specific use-case of this is to share information from the http.Request in OnConnecting with the OnMessage callback.

It could also be used to share connection state such as:
* client certificate
* duration
* message count

It may be possible to store some of this state in a map managed by the implementer of the server, but the existing serverimpl makes it trivial to share state between callbacks. Also, there is currently no overlapping fields between the OnConnecting method and the other callbacks making it impossible to share information from the request.

I also considered separating the callbacks into two groups and having OnConnecting return an implementation of the other 3 callbacks. This would allow any struct with any state to be returned by OnConnecting and the implementation of the other 3 callbacks could access that state as needed. This would require a bit more refactoring to implement.

We also discussed having the server generate a connection ID and passing it to all of the callbacks, but I prefer this approach.

### DIFF
--- a/server/httpconnection.go
+++ b/server/httpconnection.go
@@ -20,7 +20,8 @@ var ErrInvalidHTTPConnection = errors.New("cannot operate over HTTP connection")
 // and that response will be sent by OpAMP Server's HTTP request handler after the
 // onMessage callback returns.
 type httpConnection struct {
-	conn net.Conn
+	conn  net.Conn
+	state interface{}
 }
 
 func (c httpConnection) RemoteAddr() net.Addr {
@@ -38,4 +39,8 @@ func (c httpConnection) Send(_ context.Context, _ *protobufs.ServerToAgent) erro
 func (c httpConnection) Disconnect() error {
 	// Disconnect() should not be called for plain HTTP connection.
 	return ErrInvalidHTTPConnection
+}
+
+func (c httpConnection) State() interface{} {
+	return c.state
 }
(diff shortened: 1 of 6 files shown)


andykellr:
> Question: carrying the state is only necessary between OnConnecting and OnConnected callbacks, right? Once OnConnected is called we have the Connection, which will be carried to all other callbacks and _is_ already a state.
> 
> So, we essentially keeping unnecessary data in the connection struct. This is negligible for plain http, but adds memory for WS case.
> 
> Can we avoid storing the state in the connection struct and redesign this to only carry the state from OnConnecting to OnConnected callback, and then starting from OnConnected you can use the Connection itself (and if you need to keep extra data use a map with Connection being the key).

What I don't like about that approach is that the implementer needs to use 2 mechanisms to keep connection state, one to make a bridge between OnConnecting and OnConnected and another self-managed data structure to preserve state between OnConnected, OnMessage, etc.

I have another approach that I considered that you might prefer. I'll put another PR together with that approach.

andykellr:
@tigrannajaryan Please see the alternative implementation in https://github.com/open-telemetry/opamp-go/pull/148

andykellr:
Closing in favor of approach in #148

PR #144, opened 2022-11-07
Disable scheduleSend when onMessage callback is running

### LINKED ISSUE #88: Delay ScheduleSend if OnMessage is in progress
OpAMPClient functions which update the state normally call ScheduleSend to deliver the changed state to the Server. When this happens while OnMessage callback is active we want to delay calling ScheduleSend and only call it once when OnMessage handler returns. This will avoid unnecessary messages going to the server.

nemoshlag:
Resolves #88


client/internal/receivedprocessor.go
```diff
@@ -32,6 +35,7 @@ func newReceivedProcessor(
 	logger types.Logger,
 	callbacks types.Callbacks,
 	sender Sender,
+	senderCommon *SenderCommon,
```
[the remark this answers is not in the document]
nemoshlag: I agree it's confusing but since Sender is an interface I thought SenderCommon is the relevant struct for adding logic the user won't need to implement. Perhaps have Sender hold a SenderCommon instance?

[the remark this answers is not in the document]
tigrannajaryan: >Perhaps have Sender hold a SenderCommon instance?

This doesn't sound right. SenderCommon is an implementation, Sender is an interface.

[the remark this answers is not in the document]
nemoshlag: Ok thanks, please see the latest commit if I understood correctly

client/internal/receivedprocessor.go
```diff
@@ -19,6 +19,9 @@ type receivedProcessor struct {
 	// what will be sent later.
 	sender Sender
 
+	// A senderCommon to handle the scheduling of sending.
+	senderCommon *SenderCommon
```
tigrannajaryan, line 23: Unused?

client/internal/receivedprocessor.go
```diff
@@ -127,7 +139,6 @@ func (r *receivedProcessor) ProcessReceivedMessage(ctx context.Context, msg *pro
 				msgData.AgentIdentification = msg.AgentIdentification
 			}
 		}
-
 		r.callbacks.OnMessage(ctx, msgData)
 
 		r.rcvOpampConnectionSettings(ctx, msg.ConnectionSettings)
```
tigrannajaryan, line 144: I think we need to enable sending before `rcvOpampConnectionSettings` is called. rcvOpampConnectionSettings may trigger reconnection in the future and blocking sending will prevent that reconnection from working.

client/internal/sender.go
```diff
@@ -31,6 +38,12 @@ type SenderCommon struct {
 	// Indicates that there is a pending message to send.
 	hasPendingMessage chan struct{}
 
+	// Indicates onMessage callback is running
+	onMessageRunning int32
```
tigrannajaryan, line 42: Maybe name this a bit differently to match the DisableScheduleSend/DisableScheduleSend names.
```suggestion
	isSendingDisabled int32
```

client/internal/sender.go
```diff
@@ -62,6 +87,27 @@ func (h *SenderCommon) NextMessage() *NextMessage {
 	return &h.nextMessage
 }
 
+// IsOnMessageRunning returns true if onMessage callback is running
+func (h *SenderCommon) IsOnMessageRunning() bool {
```
tigrannajaryan, line 91:
```suggestion
func (h *SenderCommon) IsSendingDisabled() bool {
```

client/internal/receivedprocessor.go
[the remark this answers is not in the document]
tigrannajaryan: Yes, that's what I was envisioning. Thank you.

client/internal/receivedprocessor.go
```diff
@@ -127,9 +130,9 @@ func (r *receivedProcessor) ProcessReceivedMessage(ctx context.Context, msg *pro
 				msgData.AgentIdentification = msg.AgentIdentification
 			}
 		}
-
 		r.callbacks.OnMessage(ctx, msgData)
 
+		r.sender.EnableScheduleSend()
```
tigrannajaryan, line 135: Can we add a comment to EnableScheduleSend definition that calling EnableScheduleSend when it is already enabled is allowed? (since we do it twice here).

client/internal/sender.go
```diff
@@ -31,6 +38,12 @@ type SenderCommon struct {
 	// Indicates that there is a pending message to send.
 	hasPendingMessage chan struct{}
 
+	// Indicates onMessage callback is running
```
tigrannajaryan, line 41: Nit: the comment doesn't match what the field name implies. It would be best to reword it to something like "Set to non-zero to indicate that the message sending is disabled"

client/internal/sender.go
```diff
@@ -39,15 +52,27 @@ type SenderCommon struct {
...
+		isSendingDisabled:    0,
 	}
 }
 
 // ScheduleSend signals to HTTPSender that the message in NextMessage struct
 // is now ready to be sent. If there is no pending message (e.g. the NextMessage was
 // already sent and "pending" flag is reset) then no message will be sent.
 func (h *SenderCommon) ScheduleSend() {
+	if h.IsSendingDisabled() {
+		// onMessage callback is running, ScheduleSend() will rerun after it is done
```
tigrannajaryan, line 67: Same here, comment doesn't match what the code and func names imply.

client/internal/sender.go
```diff
@@ -62,6 +87,28 @@ func (h *SenderCommon) NextMessage() *NextMessage {
...
+	return atomic.LoadInt32(&h.isSendingDisabled) != 0
+}
+
+// DisableScheduleSend temporary preventing ScheduleSend from writing to channel
+func (h *SenderCommon) DisableScheduleSend() {
+
+	atomic.StoreInt32(&h.isSendingDisabled, 1)
+}
+
+// EnableScheduleSend re-enables ScheduleSend and checks if it was called during onMessage callback
```
tigrannajaryan, line 101: Nit: I would refrain from mentioning onMessage here. It is not a concern of SenderCommon.

tigrannajaryan (commented):
Code LGTM, I just left a few comments about improving the comments (pun not intended).

tigrannajaryan:
Technically the change in this PR can be the cause of the test failure:
```
go test -race ./...
--- FAIL: TestUpdatePackages (5.10s)
    --- FAIL: TestUpdatePackages/error_on_callback (5.03s)
        --- FAIL: TestUpdatePackages/error_on_callback/ws (5.01s)
            mockserver.go:250: Time out expecting a message from the client: compressed PackageStatuses
sent 40405, received 401[9](https://github.com/open-telemetry/opamp-go/actions/runs/3539567104/jobs/5941628042#step:7:10)0
```
This happened because the server did not receive a message client was supposed to send. I haven't seen this error before and it is not reproducible on the latest `main`.

I can't see the problem by reading the code but maybe there is some race that I am not seeing that causes a ScheduleSend to missfire?

Can you try `go test -run TestUpdatePackages/error_on_callback/ws -race -count 1000` on this change and see if it is possible to reproduce?

client/internal/sender.go
```diff
@@ -62,6 +87,30 @@ func (h *SenderCommon) NextMessage() *NextMessage {
...
+	atomic.StoreInt32(&h.isSendingDisabled, 1)
+}
+
+// EnableScheduleSend re-enables message sending, won't block on reading from channel.
+func (h *SenderCommon) EnableScheduleSend() {
+	atomic.StoreInt32(&h.isSendingDisabled, 0)
+	select {
+	case <-h.registerScheduleSend:
+		h.ScheduleSend()
+	case <-time.Tick(100 * time.Millisecond):
```
tigrannajaryan, line 107: Unclear why this is necessary.

nemoshlag: This fixed my sync issue. Several tests failed when testing with -race -count 1000 flags. Honestly, I don't fully understand why but it seems that the default case was selected over reading from the registerScheduleSend channel even when it (should have) held an unread message.

tigrannajaryan (changes requested):
Blocking until we have clarity around failing tests.

client/internal/sender.go
nemoshlag: I've removed this bypass and tests are looking stable at the moment with the updated branch. Still, I don't have a smoking gun regarding the sync issue, but it could have been resolved due to recent commits. Would appreciate your opinion on whether or not this should stay blocked or more changes are needed

srikanthccv:
@nemoshlag, any chance you can rebase and resolve the conflicts?

tigrannajaryan:
The PR now has conflicting merged and unresolved comments. I am closing. Please recreate if you would like to continue.

PR #148, opened 2022-11-11
Permit per-connection Callbacks to allow per-connection state

andykellr:
This adds a ConnectionHandler field to the ConnectionResponse struct. If nil, the Callbacks provided with StartSettings would be used.

I don't have time at the moment to provide a working demonstration of maintaining per-connection state, but hopefully this is obvious.

The implementation would look something like this:

```go
// connectionState contains any state the implementer wishes to preserve per-connection.
// It also implements types.ConnectionCallbacks
type connectionState struct {
	connectedAt time.Time
}

func (c *connectionState) OnConnected(conn types.Connection) {
	c.connectedAt = time.Now()
}

func (c *connectionState) OnMessage(conn types.Connection, message *protobufs.AgentToServer) *protobufs.ServerToAgent {
	// TODO: real message handling
	return &protobufs.ServerToAgent{}
}

func (c *connectionState) OnConnectionClose(conn types.Connection) {
}
```

The callbacks passed to StartSettings.Settings would construct a new connectionState in OnConnecting.

```go
callbacks := CallbacksStruct{
	OnConnectingFunc: func(request *http.Request) types.ConnectionResponse {
		return types.ConnectionResponse{
			Accept:             true,
			ConnectionHandler:  &connectionState{},
		}
	},
}
```


tigrannajaryan:
I like this due to stronger typing.

Earlier I said for WebSocket we may want to avoid keeping this but I think in real world scenarios you inevitably need to keep carry this state for any transport. You typically authenticate in OnConnecting and need to save the tenant id that you get as a result of authenticating somewhere since it is necessary later in OnMessage.

Should we get rid of the callback in Settings and replace it with this per connection Callbacks?

andykellr:
> I like this due to stronger typing.

I agree.

> Earlier I said for WebSocket we may want to avoid keeping this but I think in real world scenarios you inevitably need to keep carry this state for any transport. You typically authenticate in OnConnecting and need to save the tenant id that you get as a result of authenticating somewhere since it is necessary later in OnMessage.

Good example. I think it is critical that we at least figure out a way to connect OnConnecting and OnMessage and this approach feels right.

> Should we get rid of the callback in Settings and replace it with this per connection Callbacks?

I considered that, but wanted to make sure this PR was small and easy to understand. Changing the callbacks requires changes to the `CallbacksStruct`, etc.

One benefit of this implementation is compatibility with existing configurations, but there probably aren't many out there and if the same implementation is used for both callbacks, it's easy enough to return it in the ConnectionResponse.

andykellr:
@tigrannajaryan I finally had a chance to refactor this PR to separate the Callbacks and ConnectionCallbacks and update the tests.

server/types/callbacks.go
```diff
@@ -28,6 +26,11 @@ type Callbacks interface {
 	//   non-zero value to indicate the rejection reason (typically 401, 429 or 503).
 	//   HTTPResponseHeader may be optionally set (e.g. "Retry-After: 30").
 	OnConnecting(request *http.Request) ConnectionResponse
```
tigrannajaryan, line 28: Please add some comments to explain how the ConnectionHandler field in the returned struct is used, can it be nil or must be set, etc. Or perhaps add comments to the ConnectionResponse struct to explain it there.

server/types/callbacks.go
```diff
@@ -11,12 +11,10 @@ type ConnectionResponse struct {
 	Accept             bool
 	HTTPStatusCode     int
 	HTTPResponseHeader map[string]string
+	ConnectionHandler  ConnectionCallbacks
```
tigrannajaryan, line 14: Nit: should the field be also named ConnectionCallbacks? Or rename ConnectionCallbacks struct to ConnectionHandler?
The names being different somehow is not helping me to read.

server/serverimpl.go
```diff
@@ -292,22 +296,22 @@ func (s *server) handlePlainHTTPRequest(req *http.Request, w http.ResponseWriter
 		conn: connFromRequest(req),
 	}
 
-	if s.settings.Callbacks == nil {
+	if connectionHandler == nil {
```
tigrannajaryan, line 299: We probably need to document somewhere that the connection callback is required for plain http connections.

tigrannajaryan (commented):
Generally I like this direction, just left some minor comments.

server/types/callbacks.go
andykellr: Added that it MUST be set if Accept=true and added a comment to the ConnectionCallbacks interface.

tigrannajaryan (approved):

tigrannajaryan:
@andykellr I think this can be merged.

PR #156, opened 2023-02-21
Explain in CONTRIBUTING.md that opamp-spec submodule is needed

tigrannajaryan:
The submodule is necessary for `make gen-proto` target.

Also added a check in makefile to verify that the submodule is cloned.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,18 @@
 
 ## Generate Protobuf Go Files
 
-Make sure you have Docker installed.
+1. Make sure `opamp-go/internal/opamp-spec` submodule files exist. You can do this
+either by cloning the `opamp-go` repo with submodules the first time, e.g. 
+`git clone --recurse-submodules git@github.com:open-telemetry/opamp-go.git`.
+Alternatively, if you already cloned `opamp-go` without the submodule then
+execute `git submodule update --init` to fetch all files for `opamp-spec` submodule.
+Note that `opamp-spec` submodule requires ssh git cloning access to github and won't
+work if you only have https access.
 
-Run `make gen-proto`. This should compile `internal/proto/*.proto` files to 
+2. Make sure you have Docker installed and running.
+
+3. Run `make gen-proto`. This should compile `internal/proto/*.proto` files to 
 `internal/protobufs/*.pb.go` files.
+
+Note that this is tested on linux/amd64 and darwin/amd64 and is known to not work
+on darwin/arm64 (M1/M2 Macs). Fixes are welcome.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -2,7 +2,13 @@
 
 ## Generate Protobuf Go Files
 
-Make sure you have Docker installed.
+1. Make sure `opamp-go/internal/opamp-spec` submodule files exist. You can do this
```
mhausenblas, line 5: Shall we point out the `ssh` requirement? see also https://stackoverflow.com/questions/6031494/git-submodules-and-ssh-access

CONTRIBUTING.md
```diff
@@ -2,7 +2,13 @@
...
+either by cloning the `opamp-go` repo with submodules the first time, e.g. 
+`git clone --recurse-submodules git@github.com:open-telemetry/opamp-go.git`.
+Alternatively, if you already cloned `opamp-go` without the submodule then
+execute `git submodule update --init` to fetch all files for `opamp-spec` submodule.
 
-Run `make gen-proto`. This should compile `internal/proto/*.proto` files to 
+2. Make sure you have Docker installed.
```
mhausenblas, line 11: Technically correct, you need to have it installed … and running! ;)

CONTRIBUTING.md
```diff
@@ -2,7 +2,13 @@
...
+
+3. Run `make gen-proto`. This should compile `internal/proto/*.proto` files to
```
mhausenblas, line 13: We could point out that the current setup is tested for `linux/amd64` and is known to NOT work on M1 chips etc. Just to safe folks from wasting time?

CONTRIBUTING.md
tigrannajaryan: Done.

CONTRIBUTING.md
tigrannajaryan: Added.

CONTRIBUTING.md
tigrannajaryan: Done.

mhausenblas (approved):

andykellr (approved):

CONTRIBUTING.md
```diff
@@ -2,7 +2,18 @@
...
+work if you only have https access.
 
-Run `make gen-proto`. This should compile `internal/proto/*.proto` files to 
+2. Make sure you have Docker installed and running.
+
+3. Run `make gen-proto`. This should compile `internal/proto/*.proto` files to 
 `internal/protobufs/*.pb.go` files.
+
+Note that this is tested on linux/amd64 and mac/intel and is known to not work
+on M1 Macs (fixes are welcome).
```
andykellr, line 19: Pretty minor, but I would suggest using the GOOS name. Also M2 is out.

```suggestion
Note that this is tested on linux/amd64 and darwin/amd64 and is known to not work
on darwin/arm64 (M1/M2 Macs). Fixes are welcome.
```

PR #160, opened 2023-03-14
Adding missing error log statement when HTTP sending fails

kailash-bhanushali:
1. Client emits logs when receiving any error while trying to perform an HTTP connection to the server

### DIFF
--- a/client/internal/httpsender.go
+++ b/client/internal/httpsender.go
@@ -168,9 +168,6 @@ func (h *HTTPSender) sendRequestWithRetries(ctx context.Context) (*http.Response
 						interval = recalculateInterval(interval, resp)
 						err = fmt.Errorf("server response code=%d", resp.StatusCode)
 
-					case http.StatusNotFound:
-						return nil, fmt.Errorf("failed to connect to the server, server not found: %d", resp.StatusCode)
-
 					default:
 						return nil, fmt.Errorf("invalid response from server: %d", resp.StatusCode)
 					}
--- a/client/internal/httpsender.go
+++ b/client/internal/httpsender.go
@@ -117,6 +117,7 @@ func (h *HTTPSender) SetRequestHeader(header http.Header) {
 func (h *HTTPSender) makeOneRequestRoundtrip(ctx context.Context) {
 	resp, err := h.sendRequestWithRetries(ctx)
 	if err != nil {
+		h.logger.Errorf("%v", err)
 		return
 	}
 	if resp == nil {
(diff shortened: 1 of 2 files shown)


andykellr (changes requested):
This looks good. Please add a test case verifying the new error response.

kailash-bhanushali:
> This looks good. Please add a test case verifying the new error response.

I have added it, pls have a look and let me know if this looks good.

andykellr (changes requested):
Instead of using an external dependency (https://google.com/v1/opamp) which could potentially change and requires an actual response from google.com (and also a working network connection), could you spin up an http test server that returns a 404 to verify these 2 cases?

andykellr (approved):
Looks good, thanks!

@tigrannajaryan Can you merge if you're ok with these changes?

client/internal/httpsender_test.go
```diff
@@ -14,6 +14,36 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func TestHTTPSender_makeOneRequestRoundtrip(t *testing.T) {
```
tigrannajaryan, line 17: What does this test verify? I can't see any assertions.

client/internal/httpsender_test.go
```diff
@@ -57,6 +87,33 @@ func TestHTTPSenderRetryForStatusTooManyRequests(t *testing.T) {
 	srv.Close()
 }
 
+func TestHTTPSenderRetryForStatusNotFound(t *testing.T) {
```
tigrannajaryan, line 90: I am not user I understand the purpose of the test. The name suggest that it verifies that sender tries on a status not found, but I don't see the code in this test that actually does the verification. I see a single sendRequestWithRetries() call and verification that it returns an error. Where is the retry part?

client/internal/httpsender_test.go
kailash-bhanushali: not sure how to test, as I see that function does not return anything. any suggestions on how it can be done?

client/internal/httpsender_test.go
kailash-bhanushali: I got this understanding based on your msg.
1. function name should be `TestHTTPSenderStatusNotFound` as this is exactly what it is testing and `retryfor` should be removed as it is not required.
2. Currently in `httpsender.go` I am sending response as `nil` and `error` for case `StatusNotFound`, so if exactly wants to check for StatusNotFound, then we have to send `HTTPresponse` too instead of nil, if that makes sense?

client/internal/httpsender.go
```diff
@@ -167,6 +168,9 @@ func (h *HTTPSender) sendRequestWithRetries(ctx context.Context) (*http.Response
 						interval = recalculateInterval(interval, resp)
 						err = fmt.Errorf("server response code=%d", resp.StatusCode)
 
+					case http.StatusNotFound:
```
tigrannajaryan, line 171: Looks at this again, I am not sure I understand why this case is needed. We have a `default` later which returns an error for all other cases including StatusNotFound. Can you clarify what the purpose of this separate case is?

tigrannajaryan (changes requested):

client/internal/httpsender.go
kailash-bhanushali: @tigrannajaryan  got it. So, I removed the extra case and test cases and only kept `missing logging statement` which returns if any error is found. This sounds good?

tigrannajaryan:
Merged, thank you @kailash-bhanushali

PR #166, opened 2023-04-29
fix nil pointer dereference in supervisor example

haoqixu:
`Supervisor.healthCheckTicker` is prepared in `(*Supervisor).startAgent`. If there is no effective config file, we will skip the call of `startAgent` causing an nil pointer dereference in the loop.

### DIFF
--- a/internal/examples/supervisor/supervisor/supervisor_test.go
+++ b/internal/examples/supervisor/supervisor/supervisor_test.go
@@ -1,0 +1,63 @@
+package supervisor
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/open-telemetry/opamp-go/internal"
+	"github.com/open-telemetry/opamp-go/internal/examples/server/data"
+	"github.com/open-telemetry/opamp-go/internal/examples/server/opampsrv"
+	"github.com/stretchr/testify/assert"
+)
+
+func changeCurrentDir(t *testing.T) string {
+	t.Helper()
+
+	tmp := t.TempDir()
+
+	oldCWD, err := os.Getwd()
+	if err != nil {
+		t.Fatalf("getting working directory: %v", err)
+	}
+
+	if err := os.Chdir(tmp); err != nil {
+		t.Fatalf("changing working directory: %v", err)
+	}
+	t.Cleanup(func() {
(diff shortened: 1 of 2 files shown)


tigrannajaryan:
Thank you for the PR @haoqixu

Can you provide a test that reproduces this bug?

tigrannajaryan:
@haoqixu thanks for the link. Sorry I was not clear. To accept a bug fix ideally we need an automated test that shows that the bug exists and verifies that the PR fixes it. Will you be able to write a test like that and include in this PR?

haoqixu:
PTAL

internal/examples/supervisor/supervisor/supervisor_test.go
```diff
@@ -0,0 +1,63 @@
...
+  endpoint: ws://127.0.0.1:4320/v1/opamp
+agent:
+  executable: %s/dummy_agent.sh`, tmpDir)), 0644)
+
+	os.WriteFile("dummy_agent.sh", []byte("#!/bin/sh\nsleep 9999\n"), 0755)
+
+	startOpampServer(t)
+
+	supervisor, err := NewSupervisor(&internal.NopLogger{})
+	assert.Nil(t, err)
```
tigrannajaryan, line 60: Nit:
```suggestion
	assert.NoError(t, err)
```

internal/examples/supervisor/supervisor/supervisor_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/open-telemetry/opamp-go/internal"
+	"github.com/open-telemetry/opamp-go/internal/examples/server/data"
+	"github.com/open-telemetry/opamp-go/internal/examples/server/opampsrv"
+	"github.com/stretchr/testify/assert"
```
tigrannajaryan, line 11:
```suggestion
	"testing"
	
	"github.com/stretchr/testify/assert"

	"github.com/open-telemetry/opamp-go/internal"
	"github.com/open-telemetry/opamp-go/internal/examples/server/data"
	"github.com/open-telemetry/opamp-go/internal/examples/server/opampsrv"
```

tigrannajaryan (approved):

PR #170, opened 2023-06-16
Fix deadlock from Stop() called from callbacks

tigrannajaryan:
Calling Stop() from callbacks may result in a deadlock if the callback is called because a message is received. Stop() will wait for receiver to stop, but receiver can't stop until the callback is returned.

This change allow Stop() to initiate stopping and return without waiting for Stop() to finish.

This ability is important for implementing use cases where connection settings offer is received and the callback needs to Stop() the current OpAMP client and reconnect again using the new settings.

### DIFF
--- a/client/clientimpl_test.go
+++ b/client/clientimpl_test.go
@@ -465,6 +465,85 @@ func TestIncludesDetailsOnReconnect(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+func TestStopFromCallback(t *testing.T) {
+	// This test verifies calling Stop() from a callback. We had a bug previously
+	// where Stop() would hang if called from a callback.
+
+	callbacksToTest := []string{"connect", "opamp", "message"}
+	for _, callbackToTest := range callbacksToTest {
+		t.Run(
+			callbackToTest, func(t *testing.T) {
+
+				testClients(
+					t, func(t *testing.T, client OpAMPClient) {
+						var called int64
+
+						hash := []byte{1, 2, 3}
+						opampSettings := &protobufs.OpAMPConnectionSettings{DestinationEndpoint: "http://opamp.com"}
+
+						// Start a Server.
+						srv := internal.StartMockServer(t)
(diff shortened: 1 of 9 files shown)


client/clientimpl_test.go
```diff
@@ -465,6 +465,85 @@ func TestIncludesDetailsOnReconnect(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+func TestStopFromCallback(t *testing.T) {
```
tigrannajaryan, line 468: Note: before this PR's fix this test hangs in Stop(). This test verifies that the fix works correctly.

client/internal/clientcommon.go
```diff
@@ -186,6 +288,7 @@ func (c *ClientCommon) StartConnectAndRun(runner func(ctx context.Context)) {
 		return
 	}
 	c.runCancel = runCancel
+	c.isStarted = true
```
tigrannajaryan, line 291: This was a bug that resulted in racing with the goroutine started below. If the goroutine was quick enough we would get a callback while isStarted was still false.

client/internal/clientcommon.go
```diff
@@ -25,11 +27,105 @@ var (
...
+	}
+
+	return &CallbacksWrapper{
+		wrapped:    wrapped,
+		inCallback: &zero,
+	}
+}
+
+func (cc *CallbacksWrapper) EnterCallback() {
+	atomic.AddInt64(cc.inCallback, 1)
```
tigrannajaryan, line 113: At the moment we don't really have concurrent execution of callbacks, so we could just use plain non-atomic increment/decrement, but it is just trivial to use atomics here so that in case in the future we start calling callbacks concurrently we don't end up forgetting to fix this code. There is virtually no performance impact since callback are called very rarely.

tigrannajaryan:

andykellr (approved):
Looks good and confirmed the test fails without the fix.

PR #171, opened 2023-06-16
Replace Connection.RemoteAddr() by Connection()

tigrannajaryan:
I am implementing TLS in the example server and found that I need access to the underlying net.Conn. I added Connection() method to access it. Now that we have access to net.Conn, RemoteAddr() is no longer needed, it can be substituted by Connection().RemoteAddr().

Note: this is a breaking change.

### DIFF
--- a/server/httpconnection.go
+++ b/server/httpconnection.go
@@ -23,8 +23,8 @@ type httpConnection struct {
 	conn net.Conn
 }
 
-func (c httpConnection) RemoteAddr() net.Addr {
-	return c.conn.RemoteAddr()
+func (c httpConnection) Connection() net.Conn {
+	return c.conn
 }
 
 var _ types.Connection = (*httpConnection)(nil)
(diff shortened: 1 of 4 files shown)


srikanthccv (approved):

andykellr (approved):
I think the change is good. I would prefer we use the name "Connection" or "NetworkConnection" (seems redundant), rather than match the package/struct name of the connection. I will let you decide.

tigrannajaryan:
> I think the change is good. I would prefer we use the name "Connection" or "NetworkConnection" (seems redundant), rather than match the package/struct name of the connection. I will let you decide.

Just to be clear, you suggest `Connection()` as the method name instead of `NetConn()`?

andykellr:
> > I think the change is good. I would prefer we use the name "Connection" or "NetworkConnection" (seems redundant), rather than match the package/struct name of the connection. I will let you decide.
> 
> 
> 
> Just to be clear, you suggest `Connection()` as the method name instead of `NetConn()`?

Yes

Issue #163, opened 2023-03-21
wsclient: Unexpected error while receiving:read ... use of closed network connection 


srikanthccv:
When ws client is stopped, it closes the connection here  https://github.com/open-telemetry/opamp-go/blob/06011e4b754e9d41cbba7b5c166c6d2eec7b82d8/client/wsclient.go#L84-L90 and cancels the context https://github.com/open-telemetry/opamp-go/blob/06011e4b754e9d41cbba7b5c166c6d2eec7b82d8/client/internal/clientcommon.go#L151-L156 And the [ReceiverLoop](https://github.com/open-telemetry/opamp-go/blob/06011e4b754e9d41cbba7b5c166c6d2eec7b82d8/client/internal/wsreceiver.go#L45) tries to read a message from the closed connection resulting in an error.
```
Unexpected error while receiving: read tcp 127.0.0.1:53775->127.0.0.1:53774: use of closed network connection
```

portertech:
Running into this one consistently:
```
2023-11-22T22:18:00.217-0800    ERROR   internal/wsreceiver.go:53       Unexpected error while receiving: read tcp 127.0.0.1:33992->127.0.0.1:4320: use of closed network connection
github.com/open-telemetry/opamp-go/client/internal.(*wsReceiver).ReceiverLoop
        /home/portertech/go/pkg/mod/github.com/open-telemetry/opamp-go@v0.10.0/client/internal/wsreceiver.go:53
github.com/open-telemetry/opamp-go/client.(*wsClient).runOneCycle
        /home/portertech/go/pkg/mod/github.com/open-telemetry/opamp-go@v0.10.0/client/wsclient.go:244
github.com/open-telemetry/opamp-go/client.(*wsClient).runUntilStopped
        /home/portertech/go/pkg/mod/github.com/open-telemetry/opamp-go@v0.10.0/client/wsclient.go:266
github.com/open-telemetry/opamp-go/client/internal.(*ClientCommon).StartConnectAndRun.func1
        /home/portertech/go/pkg/mod/github.com/open-telemetry/opamp-go@v0.10.0/client/internal/clientcommon.go:199
```

PR #173, opened 2023-06-21
changes to check ws connection close error

phanidevavarapu:
This PR is to fix the issue https://github.com/open-telemetry/opamp-go/issues/163.
When we try to ReadMessage on a closed connection, will through net.ErrClosed error.

Tested this adding the following code
https://github.com/open-telemetry/opamp-go/blob/main/client/wsclient.go#L243
go func() { time.Sleep(5 * time.Second) c.Stop(ctx) }()

[opamp-go/client/internal/wsreceiver.go](https://github.com/open-telemetry/opamp-go/blob/06011e4b754e9d41cbba7b5c166c6d2eec7b82d8/client/internal/wsreceiver.go#L65)

Line 65 in [06011e4](https://github.com/open-telemetry/opamp-go/commit/06011e4b754e9d41cbba7b5c166c6d2eec7b82d8)
 _, bytes, err := r.conn.ReadMessage() 

		time.Sleep(5 * time.Second)

This allowed to close connection while the receiveMessage is in progress.

### DIFF
--- a/client/internal/wsreceiver_test.go
+++ b/client/internal/wsreceiver_test.go
@@ -3,8 +3,10 @@ package internal
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"testing"
 
+	"github.com/gorilla/websocket"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/proto"
@@ -20,6 +22,11 @@ type TestLogger struct {
 
 func (logger TestLogger) Debugf(format string, v ...interface{}) {
 	logger.Logf(format, v...)
+}
+
+func (logger TestLogger) Errorf(format string, v ...interface{}) {
+	err := fmt.Sprintf(format, v...)
+	logger.T.Errorf("unexpected error found: %s", err)
 }
 
 type commandAction int
@@ -141,4 +148,19 @@ func TestDecodeMessage(t *testing.T) {
 			assert.True(t, proto.Equal(msg, &decoded))
 		}
 	}
+}
+
+func TestWSReceiverReceiverLoop(t *testing.T) {
+	ctx := context.Background()
+	// Start a Server.
+	srv := StartMockServer(t)
+	defer srv.Close()
+	dialer := *websocket.DefaultDialer
+	conn, _, err := dialer.DialContext(ctx, "ws://"+srv.Endpoint, make(http.Header))
+	assert.NotNil(t, conn)
+	assert.NoError(t, err)
+	err = conn.Close()
+	assert.NoError(t, err)
+	receiver := NewWSReceiver(TestLogger{t}, nil, conn, nil, nil, nil, 0)
+	receiver.ReceiverLoop(ctx)
 }
--- a/client/internal/wsreceiver_test.go
+++ b/client/internal/wsreceiver_test.go
@@ -3,10 +3,8 @@ package internal
 import (
 	"context"
 	"fmt"
-	"net/http"
 	"testing"
 
-	"github.com/gorilla/websocket"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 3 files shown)


tigrannajaryan:
Thanks @phanidevavarapu 

Would you be able to add a unit test that verifies the fix? We should expect the unit test to fail before the fix is applied and pass after the fix.

phanidevavarapu:
> 



> Thanks @phanidevavarapu
> 
> Would you be able to add a unit test that verifies the fix? We should expect the unit test to fail before the fix is applied and pass after the fix.

Added unit test

client/internal/wsreceiver_test.go
```diff
@@ -142,3 +149,18 @@ func TestDecodeMessage(t *testing.T) {
...
+func TestWSReceiverReceiverLoop(t *testing.T) {
+	ctx := context.Background()
+	// Start a Server.
+	srv := StartMockServer(t)
+	defer srv.Close()
+	dialer := *websocket.DefaultDialer
+	conn, _, err := dialer.DialContext(ctx, "ws://"+srv.Endpoint, make(http.Header))
+	assert.NotNil(t, conn)
+	assert.NoError(t, err)
+	err = conn.Close()
```
tigrannajaryan, line 162: Is there any way that we can use wsClient here instead of create a websocket on our own? The [bug](https://github.com/open-telemetry/opamp-go/issues/163) is about when the wsClient is stopped, so this tests technically doesn't verified that, it tests something else that creates a similar situation but if wsClient implementation in the future changes this test will no longer be useful..

phanidevavarapu:
@tigrannajaryan Can you please review it now, changed the unit test to run wsclient.

client/wsclient_test.go
```diff
@@ -59,6 +73,23 @@ func TestDisconnectWSByServer(t *testing.T) {
...
+	defer srv.Close()
+
+	var settings types.StartSettings
+	settings.OpAMPServerURL = "ws://" + srv.Endpoint
+	client := NewWebSocket(TestLogger{t})
+	go func() {
+		time.Sleep(1 * time.Second)
+		startClient(t, settings, client)
+	}()
+	time.Sleep(2 * time.Second)
```
tigrannajaryan, line 89: Thanks @phanidevavarapu 

Is there any way we can wait for the specific event instead of sleeping for 1,2 seconds? This is fragile and also slows down tests.

client/wsclient_test.go
```diff
@@ -17,6 +17,19 @@ import (
 	"github.com/open-telemetry/opamp-go/protobufs"
 )
 
+type TestLogger struct {
```
tigrannajaryan, line 20:
```suggestion
type testLogger struct {
```
We don't want to export this, it is just for local usage.

tigrannajaryan (commented):
LGTM, thanks for updating the test.

Just one minor comment left. We can merge once fixed.

client/wsclient_test.go
phanidevavarapu: Fixed it. Made the change

tigrannajaryan:
The test failed while running for coverage: https://github.com/open-telemetry/opamp-go/actions/runs/5413784756/jobs/9839874168

I will rerun to see if it is unstable.

tigrannajaryan:
Failed again: https://github.com/open-telemetry/opamp-go/actions/runs/5413784756/jobs/9840358490?pr=173

tigrannajaryan:
It keeps failing: https://github.com/open-telemetry/opamp-go/actions/runs/5413784756/jobs/9840663479

Something appears to be off with the test. Can you take a look?

phanidevavarapu:
The test is working fine locally, let me tweak and test again.

tigrannajaryan:
I run the tests myself locally and they work fine for me too. It may be a timing issue since github actions are often on slower machines. I haven't looked into why it fail, but we need to make sure it is reliable before merging.

srikanthccv:
Good that it consistently failed. I looked into it. When the context is cancelled, the returned error from `c.dialer.DialContext` isn't `context.Canceled`.   https://github.com/open-telemetry/opamp-go/blob/13a6fd17b28bb7dbbdb9f119db0a0f9acd24994c/client/wsclient.go#L124 

and the error will be logged here with the message "Connection failed (dial tcp 127.0.0.1:41349: operation was canceled), will retry" https://github.com/open-telemetry/opamp-go/blob/13a6fd17b28bb7dbbdb9f119db0a0f9acd24994c/client/wsclient.go#L163-L168

We could perhaps follow what is described here https://github.com/golang/go/issues/28529#issuecomment-435883505

PR #174, opened 2023-06-21
Add OpAMP connection settings capability to the example

tigrannajaryan:
This change demonstrates OpAMP Connection Setting Offer Flow from OpAMP spec (see https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#opamp-connection-setting-offer-flow)

- The example agent and server now use TLS for OpAMP connection.
- The server's certificate is signed by a CA authority. The CA authority's certificate is self-generated (can be re-generated using examples/certs/generate.sh).
- The server may generate and offer a client-side certificate to the agent. The generated certificate is signed by the same CA.
- The client can accept the certificate and use it for future connections.
- Server UI now allows the user to generate a certificate to offer to the agent (for the first time or to rotate). The flows are described in the OpAMP spec, see https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#trust-on-first-use

 WARNING: all included certificate files are examples only and MUST NOT
 be used in production.

 How to test:
 1. Run the server.
 2. Run the agent.
 3. Go to http://localhost:4321/
 4. Find the connected agent in the list and click it
 5. Scroll down and click "Accept and Offer Client Certificate" button.
image
6. The certificate should be generated, sent to agent, accepted and agent will reconnect and the details should be shown:
image


internal/examples/agent/agent/agent.go
```diff
@@ -354,3 +394,60 @@ func (agent *Agent) onMessage(ctx context.Context, msg *types.MessageData) {
...
+		return err
+	}
+
+	agent.logger.Debugf("Reconnecting to verify offered client certificate.\n")
+
+	agent.disconnect()
+	agent.opampClientCert = cert
+	if err := agent.connect(); err != nil {
+		agent.logger.Errorf("Cannot connect using offered certificate: %s. Ignoring the offer\n", err)
+		agent.opampClientCert = nil
```
srikanthccv, line 414: Adding a TODO to take a copy of the settings might be a good idea and restore here.

internal/examples/agent/agent/agent.go
```diff
@@ -354,3 +394,60 @@ func (agent *Agent) onMessage(ctx context.Context, msg *types.MessageData) {
...
+	cert, err := agent.getCertFromSettings(settings.Certificate)
+	if err != nil {
+		return err
+	}
+
+	agent.logger.Debugf("Reconnecting to verify offered client certificate.\n")
+
+	agent.disconnect()
+	agent.opampClientCert = cert
+	if err := agent.connect(); err != nil {
```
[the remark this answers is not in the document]
tigrannajaryan: So, yes, correct, Stop()/disconnect() will return immediately, a new client will be created by connect() and will connect while the previous client is still shutting down. I think it is fine, I don't see a problem with this.

Alternately, we can wait until stopping is done and then attempt the new connection, but in this example I think it doesn't matter.

>It could be possible that SetRemoteConfigStatus is sent to the new server instead of the prev server which offered the updated config.

In this example we don't really change the destination server, so I don't think this is an issue (should not cause any problems on the server side of our example). In a production implementation I think you are right you would go with complete stopping (waiting for it) and the initiating the new connection. 

This would make the code more complicated. It is doable, perhaps can be done in a separate PR, I don't want to make already large PR even larger.

internal/examples/agent/agent/agent.go
tigrannajaryan: What exactly do you want to prevent by that? Are you worried about the new server pushing connection settings too?

tigrannajaryan (commented):
@srikanthccv thanks for reviewing.

internal/examples/agent/agent/agent.go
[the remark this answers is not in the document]
srikanthccv: Right, I was trying to understand how would this work in practice but this is fine for example

internal/examples/agent/agent/agent.go
srikanthccv: To not lose the old connection certs, if any. Say the server offered certificates two times with different settings, the first one was successful, and when the second offer attempt fails, do we go back to the last step, which had working settings or reset to nil? I think it is fine for this example.

srikanthccv (approved):
Thanks for addressing comments

internal/examples/agent/agent/agent.go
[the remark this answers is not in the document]
tigrannajaryan: Thinking more about this I believe you are right, we need to do better in the example. I added an issue to address this in a future PR: https://github.com/open-telemetry/opamp-go/issues/178

evan-bradley (approved):
Looks good to me.

internal/examples/agent/agent/agent.go
tigrannajaryan: Can be addressed in issue https://github.com/open-telemetry/opamp-go/issues/178

tigrannajaryan:
@andykellr OK to merge this or you would like to review?

Issue #175, opened 2023-06-24
Provide more flexibility for PackagesStateProvider for error 


phanidevavarapu:
During the package UpdateContent in the package syncer, the error is wrapped, instead please provide more flexibility for the PackagesStateProvider implementer for error to throw. 
This will help code the service to determine the types of errors internally.   

Note: The error message itself is wrong in this case. 
https://github.com/open-telemetry/opamp-go/blob/main/client/internal/packagessyncer.go#L279-L282

tigrannajaryan:
This needs more details. It is unclear what exactly the problem is and the goal is.

PR #182, opened 2023-06-29
PackagesStateProvider error message flexibility on UpdateContent

phanidevavarapu:
This PR provides more flexibly for the PackagesStateProvider interface implementer to format and handle the error message on UpdateContent call. 

PR for Issue: https://github.com/open-telemetry/opamp-go/issues/175

### DIFF
--- a/client/internal/packagessyncer.go
+++ b/client/internal/packagessyncer.go
@@ -276,11 +276,7 @@ func (s *packagesSyncer) downloadFile(ctx context.Context, pkgName string, file 
 	// TODO: either add a callback to verify file.Signature or pass the Signature
 	// as a parameter to UpdateContent.
 
-	err = s.localState.UpdateContent(ctx, pkgName, resp.Body, file.ContentHash)
-	if err != nil {
-		return fmt.Errorf("cannot download file from %s: %v", file.DownloadUrl, err)
-	}
-	return nil
+	return s.localState.UpdateContent(ctx, pkgName, resp.Body, file.ContentHash)
 }
 
 // deleteUnneededLocalPackages deletes local packages that are not


client/internal/packagessyncer.go
```diff
@@ -276,11 +276,7 @@ func (s *packagesSyncer) downloadFile(ctx context.Context, pkgName string, file
 	// TODO: either add a callback to verify file.Signature or pass the Signature
 	// as a parameter to UpdateContent.
 
-	err = s.localState.UpdateContent(ctx, pkgName, resp.Body, file.ContentHash)
-	if err != nil {
-		return fmt.Errorf("cannot download file from %s: %v", file.DownloadUrl, err)
```
tigrannajaryan, line 281: This error message is indeed incorrect. This should read something like "cannot store the file %s locally".

I am not sure I understand what we gain by returning the error from UpdateContent directly without wrapping it with additional context.

Can you explain the reasoning for this change?

srikanthccv: >send a custom error object with status codes for effective communication between service and client.

How does returning an error directly without wrapping it help in effective communication?

phanidevavarapu: As per the code, the error is directly set to ErrorMessage, https://github.com/open-telemetry/opamp-go/blob/main/client/internal/packagessyncer.go#L214. 

If we see this example, https://go.dev/play/p/spNCFgpESUv

By not wrapping the error, we are allowing the PackagesStateProvider interface implemented to define and send those custom errors intact.

tigrannajaryan: packagesSyncer wraps errors and adds context. You should not rely on this internal behavior of packagesSyncer and assume that it will set the error you return from PackagesStateProvider directly as ErrorMessage. That's very fragile and can easily break in a future refactoring.

It would be useful to explain more detail what exactly you are trying to do in https://github.com/open-telemetry/opamp-go/issues/175

I advise to avoid opening PRs on issues that are not discussed and accepted as actionable by maintainers/approvers.

Issue #188, opened 2023-07-06
Command should be ignored when agent does not have AcceptsCommands capability


srikanthccv:
There should be `hasCapability` check for command https://github.com/open-telemetry/opamp-go/blob/715146e48697c93ea6678ed9687b9d67ed1b07bf/client/internal/receivedprocessor.go#L54-L58

PR #189, opened 2023-07-06
OnCommand is ignored for when AcceptsCommands not set

srikanthccv:
Fixes https://github.com/open-telemetry/opamp-go/issues/188


client/internal/receivedprocessor.go
```diff
@@ -52,9 +52,13 @@ func newReceivedProcessor(
 func (r *receivedProcessor) ProcessReceivedMessage(ctx context.Context, msg *protobufs.ServerToAgent) {
 	if r.callbacks != nil {
 		if msg.Command != nil {
-			r.rcvCommand(msg.Command)
-			// If a command message exists, other messages will be ignored
-			return
+			if r.hasCapability(protobufs.AgentCapabilities_AgentCapabilities_AcceptsRestartCommand) {
```
tigrannajaryan, line 55: Perhaps it would be useful to add a comment that says when any new command capabilities are added they need to be checked here. I am not sure what's the best way to ensure we don't forget to do it, but a comment may help.

srikanthccv: Added note

tigrannajaryan (approved):

PR #194, opened 2023-08-20
Rewind the request body for retries

### LINKED ISSUE #193: httpsender will always fail to do retry request
## Description

Using httpsender to do opamp client request, when get one err by `http.Do`, then no more request can be sent successfully.

## Logs

```
# first fail
github.com/open-telemetry/opamp-go@v0.8.0/client/internal/httpsender.go:179	
Failed to do HTTP request (Post "http://example.com/v1/opamp": read tcp 10.114.31.202:11942->10.112.18.29:80: read: connection reset by peer), will retry

# retry
github.com/open-telemetry/opamp-go@v0.8.0/client/internal/httpsender.go:179	
Failed to do HTTP request (Post "http://example.com/v1/opamp": http: ContentLength=34 with Body length 0), will retry

# retry again
github.com/open-telemetry/opamp-go@v0.8.0/client/internal/httpsender.go:179	
Failed to do HTTP request (Post "http://example.com/v1/opamp": http: ContentLength=34 with Body length 0), will retry

# more retry and send no more heartbeat
...
```
## Problems

In `client/internal/httpsender.go` httpsender will prepare to send a new request by timer, but `http.Request` can not to be send twice or more, because of `http.Body` is `io.Reader`.

When the first http request failed, the retry code will always return err at L159 because of `http.Body` is nothing.

https://github.com/open-telemetry/opamp-go/blob/eb2f28dd08150252e85378e6d02d0fda8365a438/client/internal/httpsender.go#L214-L231

https://github.com/open-telemetry/opamp-go/blob/eb2f28dd08150252e85378e6d02d0fda8365a438/client/internal/httpsender.go#L152-L182

srikanthccv:
Fixes #193 

`prepareRequest` in HTTP Sender returns a `http.Request` https://github.com/open-telemetry/opamp-go/blob/deb33889ff1aac0d542a9fd08f946461cccfc1c7/client/internal/httpsender.go#L201 
We have the backoff retry strategy for failed requests; however, underlying transport would close the request body and should be rewound for any subsequent attempt. This creates a thin wrapper around the `http.Request` that helps to rewind the body before making a request. We hijack and close the underlying connection to simulate the connection error case.

### DIFF
--- a/client/internal/httpsender_test.go
+++ b/client/internal/httpsender_test.go
@@ -3,13 +3,18 @@ package internal
 import (
 	"context"
 	"crypto/tls"
+	"io"
+	"net"
 	"net/http"
+	"net/http/httptest"
+	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
 
 	"github.com/open-telemetry/opamp-go/client/types"
 	sharedinternal "github.com/open-telemetry/opamp-go/internal"
+	"github.com/open-telemetry/opamp-go/internal/testhelpers"
 	"github.com/open-telemetry/opamp-go/protobufs"
 	"github.com/stretchr/testify/assert"
 )
@@ -97,4 +102,74 @@ EKTcWGekdmdDPsHloRNtsiCa697B2O9IFA==
 	}
 
 	return cert, nil
+}
+
+func TestHTTPSenderRetryForFailedRequests(t *testing.T) {
+
+	srv, m := newMockServer(t)
+	address := testhelpers.GetAvailableLocalAddress()
+	var connectionAttempts int64
+
+	var buf []byte
+	srv.OnRequest = func(w http.ResponseWriter, r *http.Request) {
+		attempt := atomic.AddInt64(&connectionAttempts, 1)
+		if attempt == 1 {
+			hj, ok := w.(http.Hijacker)
+			if !ok {
+				t.Error("server doesn't support hijacking")
+				return
+			}
+			conn, _, err := hj.Hijack()
+			if err != nil {
+				t.Error(err)
+				return
+			}
+			conn.Close()
+		} else {
+			buf, _ = io.ReadAll(r.Body)
+			w.WriteHeader(http.StatusOK)
+		}
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	url := "http://" + address
+	sender := NewHTTPSender(&sharedinternal.NopLogger{})
+	sender.NextMessage().Update(func(msg *protobufs.AgentToServer) {
+		msg.AgentDescription = &protobufs.AgentDescription{
+			IdentifyingAttributes: []*protobufs.KeyValue{{
+				Key: "service.name",
+				Value: &protobufs.AnyValue{
+					Value: &protobufs.AnyValue_StringValue{StringValue: "test-service"},
+				},
+			}},
+		}
+	})
+	sender.callbacks = types.CallbacksStruct{
+		OnConnectFunc: func() {
+		},
+		OnConnectFailedFunc: func(_ error) {
+		},
+	}
+	sender.url = url
+	var wg sync.WaitGroup
+	wg.Add(2)
+	go func() {
+		sender.sendRequestWithRetries(ctx)
+		wg.Done()
+	}()
+	go func() {
+		time.Sleep(200 * time.Millisecond)
+		l, err := net.Listen("tcp", address)
+		assert.NoError(t, err)
+		ts := httptest.NewUnstartedServer(m)
+		ts.Listener.Close()
+		ts.Listener = l
+		ts.Start()
+		srv.srv = ts
+		wg.Done()
+	}()
+	wg.Wait()
+	assert.True(t, len(buf) > 0)
+	assert.Contains(t, string(buf), "test-service")
+	cancel()
+	srv.Close()
 }
--- a/client/internal/httpsender_test.go
+++ b/client/internal/httpsender_test.go
@@ -157,7 +157,6 @@ func TestHTTPSenderRetryForFailedRequests(t *testing.T) {
 		wg.Done()
 	}()
 	go func() {
-		time.Sleep(200 * time.Millisecond)
 		l, err := net.Listen("tcp", address)
 		assert.NoError(t, err)
 		ts := httptest.NewUnstartedServer(m)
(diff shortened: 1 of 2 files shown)


client/internal/httpsender_test.go
```diff
@@ -98,3 +100,50 @@ EKTcWGekdmdDPsHloRNtsiCa697B2O9IFA==
 
 	return cert, nil
 }
+
+func TestHTTPSenderRetryForFailedRequests(t *testing.T) {
+
+	srv := StartMockServer(t)
+	srv.srv.Config.ReadTimeout = 1 * time.Nanosecond
```
tigrannajaryan, line 107: I wonder how reliable this is. There is still a small chance that this may succeed.

Instead can we simulate an error by attempting to connect to a non-existing server? So make the first request attempt before starting the mock server. That should fail with an error. Then start them mock server.

client/internal/httpsender_test.go
```diff
@@ -98,3 +100,50 @@ EKTcWGekdmdDPsHloRNtsiCa697B2O9IFA==
...
+	sender.url = url
+	var wg sync.WaitGroup
+	wg.Add(2)
+	go func() {
+		sender.sendRequestWithRetries(ctx)
+		wg.Done()
+	}()
+	go func() {
+		time.Sleep(200 * time.Millisecond)
+		srv.srv.Config.ReadTimeout = 10 * time.Second
```
tigrannajaryan, line 141: This seems like a possibility for data race. Something is reading this field during the request.

client/internal/httpsender_test.go
srikanthccv: >I wonder how reliable this is. There is still a small chance that this may succeed.

It was pretty reliable  (observation from running it on a local machine a few dozen times).

> Instead can we simulate an error by attempting to connect to a non-existing server? So make the first request attempt before starting the mock server. That should fail with an error. Then start them mock server

I thought about this. Once the request gets prepared and goes into the retry loop, I couldn't think of a clean way to change the url. Any ideas?

client/internal/httpsender_test.go
srikanthccv: line at 107 also.

```
==================
WARNING: DATA RACE
Write at 0x00c0002dc030 by goroutine 20:
  github.com/open-telemetry/opamp-go/client/internal.TestHTTPSenderRetryForFailedRequests()
      /Users/srikanthccv/Github/opamp-go/client/internal/httpsender_test.go:107 +0x78
  testing.tRunner()
      /opt/homebrew/Cellar/go/1.20.2/libexec/src/testing/testing.go:1576 +0x180
  testing.(*T).Run.func1()
      /opt/homebrew/Cellar/go/1.20.2/libexec/src/testing/testing.go:1629 +0x40

Previous read at 0x00c0002dc030 by goroutine 21:
  net/http.http2ConfigureServer()
      /opt/homebrew/Cellar/go/1.20.2/libexec/src/net/http/h2_bundle.go:3999 +0x174
  net/http.(*Server).onceSetNextProtoDefaults()
      /opt/homebrew/Cellar/go/1.20.2/libexec/src/net/http/server.go:3331 +0x12c
  net/http.(*Server).onceSetNextProtoDefaults_Serve()
      /opt/homebrew/Cellar/go/1.20.2/libexec/src/net/http/server.go:3312 +0x100
  net/http.(*Server).onceSetNextProtoDefaults_Serve-fm()
      <autogenerated>:1 +0x34
  sync.(*Once).doSlow()
      /opt/homebrew/Cellar/go/1.20.2/libexec/src/sync/once.go:74 +0xb0
  sync.(*Once).Do()
      /opt/homebrew/Cellar/go/1.20.2/libexec/src/sync/once.go:65 +0x40
  net/http.(*Server).setupHTTP2_Serve()
      /opt/homebrew/Cellar/go/1.20.2/libexec/src/net/http/server.go:3306 +0x230
  net/http.(*Server).Serve()
      /opt/homebrew/Cellar/go/1.20.2/libexec/src/net/http/server.go:3038 +0x24c
  net/http/httptest.(*Server).goServe.func1()
      /opt/homebrew/Cellar/go/1.20.2/libexec/src/net/http/httptest/server.go:310 +0xa0
```

client/internal/httpsender_test.go
tigrannajaryan: Is it necessary to change the URL? You can find a free port number nnnn, hit http://localhost:nnnn with a request and provided that the mock server is not started the request should fail with an error. Then start the server and try the same URL again, it should pass.

srikanthccv: Let me check that.

srikanthccv: Connection to a non-existing server wasn't enough. I updated the test to take control of the underlying conn and simulate the error. Here is the run with a failed test without the fix: https://github.com/open-telemetry/opamp-go/actions/runs/6085619221/job/16510155045 and run that passes with fix https://github.com/open-telemetry/opamp-go/actions/runs/6085624937/job/16510179749. Let me know what do you think.

tigrannajaryan:
Since the code has changed, do you mind squashing into one commit and updating the PR description to reflect what the code does? I will then do another review pass.

client/internal/httpsender_test.go
```diff
@@ -98,3 +103,73 @@ EKTcWGekdmdDPsHloRNtsiCa697B2O9IFA==
...
+	}
+	sender.url = url
+	var wg sync.WaitGroup
+	wg.Add(2)
+	go func() {
+		sender.sendRequestWithRetries(ctx)
+		wg.Done()
+	}()
+	go func() {
+		time.Sleep(200 * time.Millisecond)
```
evan-bradley, line 160: I'm concerned this won't be completely reliable. My understanding is that this is intended to wait for the request to fail before starting a server, is there a way we can wait for the `OnConnectFailed` callback to fire before starting it?

tigrannajaryan: +1

srikanthccv: I added this when trying to create a failed test based on this discussion https://github.com/open-telemetry/opamp-go/pull/194#discussion_r1300231060. This is no longer needed because the connection refused error doesn't reproduce the original issue so I updated the test to create a connection reset issue. Hope this is clear.

evan-bradley (approved):

phanidevavarapu:
@tigrannajaryan @srikanthccv are we good to merge this? I also started facing the similar issue.

tigrannajaryan (approved):

PR #195, opened 2023-08-20
(coverage): Add a test case to cover that command is ignored when AcceptsCommand capability is not set

srikanthccv:


tigrannajaryan:
@srikanthccv please clarify what the change is. Was the test incorrect or did you just add more cases to it?

srikanthccv:
Ah, sorry, I added one more case. I was going through the coverage report and noticed that the command ignore part is not covered.

tigrannajaryan (approved):

PR #197, opened 2023-08-31
Added appropriate error message on package installation failure

phanidevavarapu:
Added appropriate error message on failure during the installation/upgrade of a package.

### DIFF
--- a/client/internal/packagessyncer.go
+++ b/client/internal/packagessyncer.go
@@ -278,7 +278,7 @@ func (s *packagesSyncer) downloadFile(ctx context.Context, pkgName string, file 
 
 	err = s.localState.UpdateContent(ctx, pkgName, resp.Body, file.ContentHash)
 	if err != nil {
-		return fmt.Errorf("failed to install/update the package %s: %v", pkgName, err)
+		return fmt.Errorf("failed to install/update the package %s downloaded from %s: %v", pkgName, file.DownloadUrl, err)
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


client/internal/packagessyncer.go
```diff
@@ -278,7 +278,7 @@ func (s *packagesSyncer) downloadFile(ctx context.Context, pkgName string, file
 
 	err = s.localState.UpdateContent(ctx, pkgName, resp.Body, file.ContentHash)
 	if err != nil {
-		return fmt.Errorf("cannot download file from %s: %v", file.DownloadUrl, err)
+		return fmt.Errorf("failed to install/update the package %s: %v", pkgName, err)
```
tigrannajaryan, line 281: I don't think the error message here should be changed like this. The error is in downloading the file and that's what it should say. If you want to have additional information like the package name it is fine to include it. Alternatively you can enrich the error message at the call site and add the package name.

phanidevavarapu: But if we look at the actual code flow, we are past the downloading the package.

tigrannajaryan: You are right. Can we include both package name and URL we downloaded from in the message? Something like this: `failed to install/update the package %s downloaded from %v`?

tigrannajaryan (approved):

tigrannajaryan:
Thank you @phanidevavarapu

PR #200, opened 2023-09-15
Embed mutex in the server's websocket client

evan-bradley:
The Gorilla websocket package states that [only one concurrent write](https://pkg.go.dev/github.com/gorilla/websocket#hdr-Concurrency) can be made to a connection at a time. The library will try to detect concurrent writes and will panic if it detects that one is happening.

We might see concurrent writes in two places:
1. If the server application doesn't guard writes when writing to an agent connection.
2. If the OpAMP server library is [responding to a message](https://github.com/open-telemetry/opamp-go/blob/146b79221c76637368e7937b50a82689733f46a2/server/serverimpl.go#L233) from the agent while the server application is also sending a message using its reference to the agent connection obtained from `OnConnectedFunc`.

In my opinion it would make sense to prevent concurrent writes within the library to relieve the application from having to worry about this detail. This does necessitate a mutex per connection, but I don't see any significant performance impact from the mutexes.

Without the mutex in `wsConnection`:
```
go test -bench=. -run=^# -count=5
goos: linux
goarch: amd64
pkg: github.com/open-telemetry/opamp-go/server
cpu: 12th Gen Intel(R) Core(TM) i9-12900
BenchmarkSendToClient-24           14610            118877 ns/op
BenchmarkSendToClient-24           17510            326918 ns/op
BenchmarkSendToClient-24           16641            267099 ns/op
BenchmarkSendToClient-24           16287            239085 ns/op
BenchmarkSendToClient-24           16243            238157 ns/op
PASS
ok      github.com/open-telemetry/opamp-go/server       23.877s
```

With the mutex:
```
go test -bench=. -run=^# -count=5
goos: linux
goarch: amd64
pkg: github.com/open-telemetry/opamp-go/server
cpu: 12th Gen Intel(R) Core(TM) i9-12900
BenchmarkSendToClient-24           15170            175727 ns/op
BenchmarkSendToClient-24           16916            307997 ns/op
BenchmarkSendToClient-24           16446            249592 ns/op
BenchmarkSendToClient-24           16563            260181 ns/op
BenchmarkSendToClient-24           17136            315825 ns/op
PASS
ok      github.com/open-telemetry/opamp-go/server       25.690s
```

This is a fairly rudimentary solution, but it does solve the issue I was seeing and is straightforward. I'm open to more sophisticated ways of managing the write concurrency if there's a deficiency with this approach.

### DIFF
--- a/server/serverimpl_test.go
+++ b/server/serverimpl_test.go
@@ -755,9 +755,9 @@ func TestConnectionAllowsConcurrentWrites(t *testing.T) {
 
 	select {
 	case <-timeout.Done():
-		t.Error("Connections failed to establish in time")
+		t.Error("Client failed to connect before timeout")
 	default:
-		if srvConnVal.Load() != nil {
+		if _, ok := srvConnVal.Load().(types.Connection); ok == true {
 			break
 		}
 	}
--- a/server/serverimpl_test.go
+++ b/server/serverimpl_test.go
@@ -726,12 +726,12 @@ func TestDecodeMessage(t *testing.T) {
 }
 
 func TestConnectionAllowsConcurrentWrites(t *testing.T) {
-	var srvConn types.Connection
+	srvConnVal := atomic.Value{}
 	callbacks := CallbacksStruct{
 		OnConnectingFunc: func(request *http.Request) types.ConnectionResponse {
 			return types.ConnectionResponse{Accept: true, ConnectionCallbacks: ConnectionCallbacksStruct{
 				OnConnectedFunc: func(conn types.Connection) {
-					srvConn = conn
+					srvConnVal.Store(conn)
 				},
 			}}
 		},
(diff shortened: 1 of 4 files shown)


server/serverimpl_test.go
```diff
@@ -723,3 +724,124 @@ func TestDecodeMessage(t *testing.T) {
...
+
+	defer conn.Close()
+
+	timeout, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+
+	select {
+	case <-timeout.Done():
+		t.Error("Client failed to connect before timeout")
+	default:
+		if _, ok := srvConnVal.Load().(types.Connection); ok == true {
```
evan-bradley, line 760: I get a linter warning that `ok` is unused if I just have `ok` as the if-statement condition. `ok == true` fixes the warning, not sure how else to address this.

tigrannajaryan (approved):
LGTM. Thanks for fixing.

tigrannajaryan:
Hint: https://pkg.go.dev/golang.org/x/perf/cmd/benchstat is very useful for showing before/after benchmark comparison.

Here is what it produces for your benchmark results:
```
name             old time/op  new time/op  delta
SendToClient-24   238µs ±50%   262µs ±33%   ~     (p=0.690 n=5+5)
```
Looks like it is close, but could use a more stable execution environment for benchmarking. :-)

andykellr (approved):

tigrannajaryan:
@evan-bradley opamp-go v0.9.0 released: https://github.com/open-telemetry/opamp-go/releases/tag/v0.9.0

PR #201, opened 2023-09-16
Move certs & helper code from /internal/examples to /internal

### LINKED ISSUE #199: Move certs helper code from /internal/examples to /internal or /internal/testhelper?
In an effort to improve the test coverage for [serverimpl](https://app.codecov.io/gh/open-telemetry/opamp-go/blob/main/server%2Fserverimpl.go), I started to add some tests with TLS enabled. I noticed we already have some useful code for generating and reading certs in `internal/examples/...`, and it seems like a good idea to use for tests. I was thinking of moving them to `/internal` or maybe `internal/testhelper` and use in both serverimpl tests and examples.

@tigrannajaryan do you see any issues with this? If not, I can send a PR for this along with the new serverimpl TLS tests.

srikanthccv:
Fixes #199

I verified that examples run with the changes.


internal/certs.go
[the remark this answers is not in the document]
srikanthccv: I initially thought of moving them to `/internal/testhelpers`, but they will be used by both tests and examples, so  `/internal` seemed a good choice. I don't have a strong opinion here. How about `/internal/certhelpers`?

internal/certs.go
```diff
@@ -0,0 +1,160 @@
...
+	"os"
+	"path"
+	"time"
+
+	"github.com/open-telemetry/opamp-go/protobufs"
+)
+
+func CreateClientTLSConfig(clientCert *tls.Certificate, certsDir string) (*tls.Config, error) {
+	// Read the CA's public key. This is the CA that signs the server's certificate.
+	caCertBytes, err := os.ReadFile(path.Join(certsDir, "certs/ca.cert.pem"))
```
tigrannajaryan, line 25: Since this is now a more general-purpose utility let's avoid making assumptions about directory structure here and instead just accept as a function parameter the full path to the pem file instead of accepting the `certsDir` as a parameter.
The caller should be responsible for knowing in which directory the file resides.

internal/certs.go
```diff
@@ -0,0 +1,160 @@
...
+	"math/big"
+	"net"
+	"os"
+	"path"
+	"time"
+
+	"github.com/open-telemetry/opamp-go/protobufs"
+)
+
+func CreateClientTLSConfig(clientCert *tls.Certificate, certsDir string) (*tls.Config, error) {
```
tigrannajaryan, line 23:
```suggestion
func CreateClientTLSConfig(clientCert *tls.Certificate, caPublicKeyPath string) (*tls.Config, error) {
```

internal/certs.go
```diff
@@ -0,0 +1,160 @@
...
+		RootCAs: caCertPool,
+	}
+	if clientCert != nil {
+		// If there is a client-side certificate use it for connection too.
+		cfg.Certificates = []tls.Certificate{*clientCert}
+	}
+	return cfg, nil
+}
+
+func CreateServerTLSConfig(certsDir string) (*tls.Config, error) {
```
tigrannajaryan, line 46: Same here. I think I would prefer to accept the 3 file paths directly (instead of accepting the directory name): the CA public key path, the server's private and public ket file paths.

evan-bradley (approved):

internal/examples/server/uisrv/ui.go
```diff
@@ -130,7 +121,7 @@ func rotateInstanceClientCert(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Create a new certificate for the agent.
-	certificate, err := createTLSCert()
+	certificate, err := internal.CreateTLSCert("../../certs/certs/ca.cert.pem", "../../certs/certs/ca.key.pem")
```
tigrannajaryan, line 124: Same here.

tigrannajaryan (commented):
Code change LGTM, just need to confirm that the paths are correct.

internal/examples/agent/agent/agent.go
```diff
@@ -89,9 +89,17 @@ func NewAgent(logger types.Logger, agentType string, agentVersion string) *Agent
 func (agent *Agent) connect() error {
 	agent.opampClient = client.NewWebSocket(agent.logger)
 
+	tlsConfig, err := internal.CreateClientTLSConfig(
+		agent.opampClientCert,
+		"../../certs/certs/ca.cert.pem",
```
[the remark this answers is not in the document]
srikanthccv: Yes, the path changed directory from `/internal/examples/` to `/internal`. I verified that examples run.

tigrannajaryan (approved):

PR #203, opened 2023-09-16
Avoid spurious error log for Client.Stop closed connections

### LINKED ISSUE #163: wsclient: Unexpected error while receiving:read ... use of closed network connection

srikanthccv:
Co-authored-by: phanidevavarapu

This is a continuation of closed PR https://github.com/open-telemetry/opamp-go/pull/173. Fixes #163

### DIFF
--- a/client/internal/wsreceiver.go
+++ b/client/internal/wsreceiver.go
@@ -2,7 +2,9 @@ package internal
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net"
 
 	"github.com/gorilla/websocket"
 	"github.com/open-telemetry/opamp-go/client/types"
@@ -49,7 +51,7 @@ out:
 	for {
 		var message protobufs.ServerToAgent
 		if err := r.receiveMessage(&message); err != nil {
-			if ctx.Err() == nil && !websocket.IsCloseError(err, websocket.CloseNormalClosure) {
+			if ctx.Err() == nil && !websocket.IsCloseError(err, websocket.CloseNormalClosure) && !errors.Is(err, net.ErrClosed) {
 				r.logger.Errorf("Unexpected error while receiving: %v", err)
 			}
 			break out
--- a/client/wsclient.go
+++ b/client/wsclient.go
@@ -123,12 +123,6 @@ func (c *wsClient) tryConnectOnce(ctx context.Context) (retryAfter sharedinterna
 	var resp *http.Response
 	conn, resp, err := c.dialer.DialContext(ctx, c.url.String(), c.requestHeader)
 	if err != nil {
-		// `DailContext` returns a mapped error for any context error, we remap it
-		// to the original context error.
-		// See https://github.com/golang/go/blob/561a5079057e3a660ab638e1ba957a96c4ff3fd1/src/net/net.go#L424-L435
-		if ctx.Err() == context.Canceled || ctx.Err() == context.DeadlineExceeded {
-			err = ctx.Err()
-		}
 		if c.common.Callbacks != nil && !c.common.IsStopping() {
 			c.common.Callbacks.OnConnectFailed(err)
 		}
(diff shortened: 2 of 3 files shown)


client/wsclient.go
```diff
@@ -119,19 +119,25 @@ func (c *wsClient) SetPackageStatuses(statuses *protobufs.PackageStatuses) error
...
+func (c *wsClient) tryConnectOnce(ctx context.Context) (retryAfter sharedinternal.OptionalDuration, err error) {
 	var resp *http.Response
 	conn, resp, err := c.dialer.DialContext(ctx, c.url.String(), c.requestHeader)
 	if err != nil {
+		// `DailContext` returns a mapped error for any context error, we remap it
+		// to the original context error.
+		// See https://github.com/golang/go/blob/561a5079057e3a660ab638e1ba957a96c4ff3fd1/src/net/net.go#L424-L435
+		if ctx.Err() == context.Canceled || ctx.Err() == context.DeadlineExceeded {
+			err = ctx.Err()
+		}
```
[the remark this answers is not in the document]
srikanthccv: I checked again, and this should be supported with the newer go release as they implemented `errors.Is` for these un-exported errors. I will create a separate PR to bump the version and then remove this code.

client/internal/wsreceiver.go
```diff
@@ -49,7 +51,7 @@ out:
 	for {
 		var message protobufs.ServerToAgent
 		if err := r.receiveMessage(&message); err != nil {
-			if ctx.Err() == nil && !websocket.IsCloseError(err, websocket.CloseNormalClosure) {
+			if ctx.Err() == nil && !websocket.IsCloseError(err, websocket.CloseNormalClosure) && !errors.Is(err, net.ErrClosed) {
```
evan-bradley, line 54: Were you able to determine why this error occurs? My first impression is that it indicates we're not properly managing the connection. It should be fine to ignore this error, but it would be nice to know why it's happening.

srikanthccv: This happens because there is a chance of reading the closed connection in `ReceiverLoop`. I tried to capture the details of the issue here: https://github.com/open-telemetry/opamp-go/issues/163.

client/wsclient_test.go
```diff
@@ -59,6 +72,24 @@ func TestDisconnectWSByServer(t *testing.T) {
...
+
+	var settings types.StartSettings
+	settings.OpAMPServerURL = "ws://" + srv.Endpoint
+	client := NewWebSocket(testLogger{t})
+	done := make(chan struct{})
+	go func() {
+		startClient(t, settings, client)
+		done <- struct{}{}
+	}()
+	eventually(t, func() bool { <-done; return true })
```
evan-bradley, line 89: I don't think this will provide reliable confirmation that the client connected. `startClient` does not block on connecting to a server, so the `done` channel will likely emit a message almost immediately.

client/wsclient_test.go
```diff
@@ -59,6 +72,24 @@ func TestDisconnectWSByServer(t *testing.T) {
...
+	_ = client.Stop(context.Background())
```
evan-bradley, line 90: Should we check the error here?

evan-bradley:
@srikanthccv @tigrannajaryan I've drafted some changes that I believe may solve the same issue without ignoring the `net.ErrClosed` error: https://github.com/open-telemetry/opamp-go/pull/205. Please take a look at let me know what your thoughts are.

srikanthccv:
Closing since there are other pull requests to gracefully close the connection.

PR #205, opened 2023-09-20
Gracefully close websocket connections from the client

evan-bradley:
On calling `OpAMPClient.Stop` for websocket clients, the client closes the TCP connection without performing a closing handshake, which from my testing throws a handful of errors in the client and the server.

This change attempts to gracefully close the connection by initiating a closing handshake [as specified](https://www.rfc-editor.org/rfc/rfc6455.html#section-7.1.2) in RFC 6455. This gives the connection on both sides time to clean up before terminating the TCP connection, and is more compliant with the Websocket specification.

Testing with the Supervisor, this cleans up the following errors in the client:

```
2023-09-20T16:16:42.018-0400    ERROR   internal/wssender.go:77 Cannot write WS message: use of closed network connection
github.com/open-telemetry/opamp-go/client/internal.(*WSSender).sendMessage
        /home/ihsmwaqhcpfwcime/code/github.com/evan-bradley/opamp-go/client/internal/wssender.go:77
github.com/open-telemetry/opamp-go/client/internal.(*WSSender).sendNextMessage
        /home/ihsmwaqhcpfwcime/code/github.com/evan-bradley/opamp-go/client/internal/wssender.go:70
github.com/open-telemetry/opamp-go/client/internal.(*WSSender).run
        /home/ihsmwaqhcpfwcime/code/github.com/evan-bradley/opamp-go/client/internal/wssender.go:56
```

```
2023-09-20T16:16:42.018-0400    ERROR   internal/wsreceiver.go:53       Unexpected error while receiving: read tcp 127.0.0.1:45242->127.0.0.1:4320: use of closed network connection
github.com/open-telemetry/opamp-go/client/internal.(*wsReceiver).ReceiverLoop
        /home/ihsmwaqhcpfwcime/code/github.com/evan-bradley/opamp-go/client/internal/wsreceiver.go:53
github.com/open-telemetry/opamp-go/client.(*wsClient).runOneCycle
        /home/ihsmwaqhcpfwcime/code/github.com/evan-bradley/opamp-go/client/wsclient.go:243
github.com/open-telemetry/opamp-go/client.(*wsClient).runUntilStopped
        /home/ihsmwaqhcpfwcime/code/github.com/evan-bradley/opamp-go/client/wsclient.go:265
github.com/open-telemetry/opamp-go/client/internal.(*ClientCommon).StartConnectAndRun.func1
        /home/ihsmwaqhcpfwcime/code/github.com/evan-bradley/opamp-go/client/internal/clientcommon.go:199
```

This also cleans up the following error in the server that I have seen while running tests:
```
Cannot send message to WebSocket: write tcp 127.0.0.1:42589->127.0.0.1:48774: write: connection reset by peer
```

### DIFF
--- a/client/wsclient.go
+++ b/client/wsclient.go
@@ -7,7 +7,6 @@ import (
 	"net/http"
 	"net/url"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
@@ -41,15 +40,20 @@ type wsClient struct {
 	// The sender is responsible for sending portion of the OpAMP protocol.
 	sender *internal.WSSender
 
-	isStopped atomic.Bool
+	// Indicates whether the client is open for more messages to be sent.
+	// Should be protected by connectionOpenMutex.
+	connectionOpen bool
+	// Indicates the connection is being written to.
+	// A read lock on this mutex indicates that a message is being queued for writing.
+	// A write lock on this mutex indicates that the connection is being shut down.
+	connectionOpenMutex sync.RWMutex
 
-	// Sends a signal to stop background processors that asynchronously use the
-	// WebSocket connection, e.g. WSSender.
-	stopBGProcessors chan struct{}
-
-	// Responds to a signal from stopProcessors indicating that all processors
+	// Sends a signal to the background processors controller thread to stop
+	// all background processors.
+	stopBGProcessing chan struct{}
+	// Responds to a signal from stopBGProcessing indicating that all processors
 	// have been stopped.
-	bgProcessorsStopped chan struct{}
+	bgProcessingStopped chan struct{}
 
 	// Network connection timeout used for the WebSocket closing handshake.
 	// This field is currently only modified during testing.
@@ -68,8 +72,9 @@ func NewWebSocket(logger types.Logger) *wsClient {
 	w := &wsClient{
 		common:              internal.NewClientCommon(logger, sender),
 		sender:              sender,
-		stopBGProcessors:    make(chan struct{}, 1),
-		bgProcessorsStopped: make(chan struct{}, 1),
+		connectionOpen:      true,
+		stopBGProcessing:    make(chan struct{}, 1),
+		bgProcessingStopped: make(chan struct{}, 1),
 		connShutdownTimeout: 10 * time.Second,
 	}
 	return w
@@ -104,7 +109,11 @@ func (c *wsClient) Start(ctx context.Context, settings types.StartSettings) erro
 }
 
 func (c *wsClient) Stop(ctx context.Context) error {
-	c.isStopped.Store(true)
+	// Prevent any additional writers from writing to the connection
+	// and stop reconnecting if the connection closes.
+	c.connectionOpenMutex.Lock()
+	c.connectionOpen = false
+	c.connectionOpenMutex.Unlock()
 
 	// Close connection if any.
 	c.connMutex.RLock()
@@ -113,13 +122,16 @@ func (c *wsClient) Stop(ctx context.Context) error {
 
 	if conn != nil {
 		// Shut down the sender and any other background processors.
-		c.stopBGProcessors <- struct{}{}
+		c.stopBGProcessing <- struct{}{}
 		select {
-		case <-c.bgProcessorsStopped:
+		case <-c.bgProcessingStopped:
 		case <-ctx.Done():
-			_ = c.conn.Close()
+			c.closeConnection()
 			return errEarlyStop
 		}
+
+		// At this point all other writers to the connection should be stopped.
+		// We can write to the connection without any risk of contention.
 
 		defaultCloseHandler := conn.CloseHandler()
 		closed := make(chan struct{})
@@ -133,11 +145,14 @@ func (c *wsClient) Stop(ctx context.Context) error {
 			return err
 		})
 
+		// Start the closing handshake by writing a close message to the server.
+		// If the server responds with its own close message, the connection reader will
+		// shut down and there will be no more reads from or writes to the connection.
 		message := websocket.FormatCloseMessage(websocket.CloseNormalClosure, "")
 		err := conn.WriteControl(websocket.CloseMessage, message, time.Now().Add(c.connShutdownTimeout))
 
 		if err != nil {
-			_ = c.conn.Close()
+			c.closeConnection()
 			return fmt.Errorf("could not write close message to WebSocket, connection closed without performing closing handshake: %w", err)
 		}
 
@@ -146,9 +161,9 @@ func (c *wsClient) Stop(ctx context.Context) error {
 			// runOneCycle will close the connection if the closing handshake completed,
 			// so there's no need to close it here.
 		case <-time.After(c.connShutdownTimeout):
-			_ = c.conn.Close()
+			c.closeConnection()
 		case <-ctx.Done():
-			_ = c.conn.Close()
+			c.closeConnection()
 			return errEarlyStop
 		}
 	}
@@ -161,35 +176,45 @@ func (c *wsClient) AgentDescription() *protobufs.AgentDescription {
 }
 
 func (c *wsClient) SetAgentDescription(descr *protobufs.AgentDescription) error {
-	if c.isStopped.Load() {
+	c.connectionOpenMutex.RLock()
+	defer c.connectionOpenMutex.RUnlock()
+	if !c.connectionOpen {
 		return errStopping
 	}
 	return c.common.SetAgentDescription(descr)
 }
 
 func (c *wsClient) SetHealth(health *protobufs.AgentHealth) error {
-	if c.isStopped.Load() {
+	c.connectionOpenMutex.RLock()
+	defer c.connectionOpenMutex.RUnlock()
+	if !c.connectionOpen {
 		return errStopping
 	}
 	return c.common.SetHealth(health)
 }
 
 func (c *wsClient) UpdateEffectiveConfig(ctx context.Context) error {
-	if c.isStopped.Load() {
+	c.connectionOpenMutex.RLock()
+	defer c.connectionOpenMutex.RUnlock()
+	if !c.connectionOpen {
 		return errStopping
 	}
 	return c.common.UpdateEffectiveConfig(ctx)
 }
 
 func (c *wsClient) SetRemoteConfigStatus(status *protobufs.RemoteConfigStatus) error {
-	if c.isStopped.Load() {
+	c.connectionOpenMutex.RLock()
+	defer c.connectionOpenMutex.RUnlock()
+	if !c.connectionOpen {
 		return errStopping
 	}
 	return c.common.SetRemoteConfigStatus(status)
 }
 
 func (c *wsClient) SetPackageStatuses(statuses *protobufs.PackageStatuses) error {
-	if c.isStopped.Load() {
+	c.connectionOpenMutex.RLock()
+	defer c.connectionOpenMutex.RUnlock()
+	if !c.connectionOpen {
 		return errStopping
 	}
 	return c.common.SetPackageStatuses(statuses)
@@ -271,6 +296,21 @@ func (c *wsClient) ensureConnected(ctx context.Context) error {
 	}
 }
 
+func (c *wsClient) closeConnection() {
+	c.connMutex.Lock()
+	defer c.connMutex.Unlock()
+
+	if c.conn == nil {
+		return
+	}
+
+	// Close the connection.
+	_ = c.conn.Close()
+
+	// Unset the field to indicate that the connection is closed.
+	c.conn = nil
+}
+
 // runOneCycle performs the following actions:
 //  1. connect (try until succeeds).
 //  2. set up a background processor to send messages.
@@ -286,15 +326,7 @@ func (c *wsClient) runOneCycle(ctx context.Context) {
 		return
 	}
 
-	defer func() {
-		// Close the connection.
-		_ = c.conn.Close()
-
-		// Unset the closed connection to indicate that it is closed.
-		c.connMutex.Lock()
-		c.conn = nil
-		c.connMutex.Unlock()
-	}()
+	defer c.closeConnection()
 
 	if c.common.IsStopping() {
 		return
@@ -315,10 +347,10 @@ func (c *wsClient) runOneCycle(ctx context.Context) {
 	// will only stop when the connection closes or errors.
 	go func() {
 		select {
-		case <-c.stopBGProcessors:
+		case <-c.stopBGProcessing:
 			procCancel()
 			c.sender.WaitToStop()
-			c.bgProcessorsStopped <- struct{}{}
+			close(c.bgProcessingStopped)
 		case <-procCtx.Done():
 		}
 	}()
@@ -358,9 +390,13 @@ func (c *wsClient) runOneCycle(ctx context.Context) {
 func (c *wsClient) runUntilStopped(ctx context.Context) {
 	// Iterates until we detect that the client is stopping.
 	for {
-		if c.common.IsStopping() || c.isStopped.Load() {
+		c.connectionOpenMutex.RLock()
+		if c.common.IsStopping() || !c.connectionOpen {
+			c.connectionOpenMutex.RUnlock()
 			return
 		}
+		c.connectionOpenMutex.RUnlock()
+
 		c.runOneCycle(ctx)
 	}
 }
--- a/client/wsclient.go
+++ b/client/wsclient.go
@@ -85,7 +85,41 @@ func (c *wsClient) Stop(ctx context.Context) error {
 	conn := c.conn
 	c.connMutex.RUnlock()
 
+	ticker := time.NewTicker(50 * time.Millisecond)
+
+	// Wait for all remaining messages to be sent. Continuing to attempt
+	// to send messages after calling Stop will eventually result in lost
+	// messages.
+	select {
+	case <-ticker.C:
+		if !c.sender.NextMessage().IsPending() {
+			break
+		}
+	case <-time.After(3 * time.Second):
+		break
+	}
+
+	ticker.Stop()
+
 	if conn != nil {
+		defaultCloseHandler := conn.CloseHandler()
+		closed := make(chan bool)
+
+		// The server should respond with a close message of its own, which will
+		// trigger this callback. At this point the close sequence has been
+		// completed and the TCP connection can be gracefully closed.
+		conn.SetCloseHandler(func(code int, text string) error {
+			closed <- true
+			return defaultCloseHandler(code, text)
+		})
+
+		message := websocket.FormatCloseMessage(websocket.CloseNormalClosure, "")
+		conn.WriteControl(websocket.CloseMessage, message, time.Now().Add(time.Second))
+
+		select {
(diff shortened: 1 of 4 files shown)


srikanthccv:
Thanks, and this graceful close is good, but I think this also has the same issue. Isn't there a good chance that `ReceiverLoop` tries to read the message after the close message is sent? And the `ReadMessage` will return an error something like `websocket: close sent`?

evan-bradley:
>Isn't there a good chance that ReceiverLoop tries to read the message after the close message is sent?

In an ordinary circumstance it is expected that the ReceiverLoop will read one final message after the client sends its close message. The Websocket close handshake includes a close response from the server after the client sends its close message, so we will hit [this condition](https://github.com/open-telemetry/opamp-go/blob/main/client/internal/wsreceiver.go#L52) and the receiver loop will close.

>And the ReadMessage will return an error something like `websocket: close sent`?

The `websocket: close sent` error is only thrown for attempting to send messages. The code now waits for any pending messages to be sent to avoid this error. Any messages that the client implementation attempts to send after calling `Stop` will throw this error, but we could probably avoid that by throwing an error to the client implementation directly.

evan-bradley:
To the best of my understanding that error will only be thrown when writing to the WebSocket connection. I did a search for the error and the only places it came up were when writing to the WebSocket: https://github.com/search?q=repo%3Agorilla%2Fwebsocket%20ErrCloseSent&type=code.

From a specification point of view, I don't think it would make sense to throw that error on reading from the WebSocket connection. For the closing handshake the client needs a close message from the peer, so it will need to read messages from the WebSocket until it reads the peer's close message.

srikanthccv:
I couldn't recreate the error case I mentioned earlier. You are correct; this is clearly a better way to close.

evan-bradley:
>I couldn't recreate the error case I mentioned earlier. 

Thanks for confirming. I think this should be ready for review now.

client/wsclient.go
```diff
@@ -85,7 +87,41 @@ func (c *wsClient) Stop(ctx context.Context) error {
...
+
+	// Wait for all remaining messages to be sent. Continuing to attempt
+	// to send messages after calling Stop will eventually result in lost
+	// messages.
+	select {
+	case <-ticker.C:
+		if !c.sender.NextMessage().IsPending() {
+			break
+		}
+	case <-time.After(3 * time.Second):
```
tigrannajaryan, line 100: Is this ever going to fire? The first case is always going to happen first, right?

client/wsclient.go
```diff
@@ -85,7 +87,41 @@ func (c *wsClient) Stop(ctx context.Context) error {
...
 	c.connMutex.RUnlock()
 
+	ticker := time.NewTicker(50 * time.Millisecond)
+
+	// Wait for all remaining messages to be sent. Continuing to attempt
+	// to send messages after calling Stop will eventually result in lost
+	// messages.
+	select {
+	case <-ticker.C:
+		if !c.sender.NextMessage().IsPending() {
```
tigrannajaryan, line 97: Is this essentially a sleep of 50ms?

client/wsclient.go
```diff
@@ -85,7 +87,41 @@ func (c *wsClient) Stop(ctx context.Context) error {
 	conn := c.conn
 	c.connMutex.RUnlock()
 
+	ticker := time.NewTicker(50 * time.Millisecond)
+
+	// Wait for all remaining messages to be sent. Continuing to attempt
+	// to send messages after calling Stop will eventually result in lost
+	// messages.
+	select {
```
tigrannajaryan, line 95: I am confused by what these select and cases do.

client/wsclient.go
evan-bradley: Realistically the first case should always happen first, yes. The most likely situation where we would hit this case is if the client implementation was continuously scheduling messages to be sent to the server and the sender thread wasn't processing them quickly enough for us to hit this case.

client/wsclient.go
evan-bradley: The goal of this select is to allow the sender to deliver any queued messages by polling the sender for whether it has any messages remaining. Currently in the Supervisor we call `SetHealth` immediately before calling `Stop`, and the lack of delay between the two doesn't allow time for the health message to be sent and produces a `websocket: close sent` error.

This was the most obvious solution to the issue, but I'll take another look to see whether there are other viable approaches.

client/wsclient.go
evan-bradley: In most cases it will be, but I have seen this fire multiple times when running this code with a debugger.

client/wsclient.go
tigrannajaryan: I don't understand. How can this second case ever get selected? There are 2 timers, one fires in 50ms, the other in 3 seconds. What sort of timer discrepancy do you expect that selects the second case before the first case's channel is ready for reading?

I am probably missing something but it seems what effectively lines are 90-104 do is sleep for 50ms.

tigrannajaryan: >I inexplicably convinced myself there was a loop around this

I was suspecting this :-) I should have mentioned this in my comment, my fault.

client/wsclient.go
```diff
@@ -35,8 +38,15 @@ type wsClient struct {
 
 	// The sender is responsible for sending portion of the OpAMP protocol.
 	sender *internal.WSSender
+
+	isStopped atomic.Bool
+
+	stopProcessors    chan struct{}
+	processorsStopped chan struct{}
```
tigrannajaryan, line 45: Please document these fields.

client/wsclient.go
```diff
@@ -80,13 +92,40 @@ func (c *wsClient) Start(ctx context.Context, settings types.StartSettings) erro
...
+		conn.SetCloseHandler(func(code int, text string) error {
+			err := defaultCloseHandler(code, text)
+			closed <- struct{}{}
+			return err
+		})
+
+		message := websocket.FormatCloseMessage(websocket.CloseNormalClosure, "")
+		conn.WriteControl(websocket.CloseMessage, message, time.Now().Add(time.Second))
+
+		select {
```
tigrannajaryan, line 122: I think we also need to honour the `ctx`', so probably add a `case <-ctx.Done():`.

client/wsclient.go
```diff
@@ -80,13 +92,40 @@ func (c *wsClient) Start(ctx context.Context, settings types.StartSettings) erro
...
+		// trigger this callback. At this point the close sequence has been
+		// completed and the TCP connection can be gracefully closed.
+		conn.SetCloseHandler(func(code int, text string) error {
+			err := defaultCloseHandler(code, text)
+			closed <- struct{}{}
+			return err
+		})
+
+		message := websocket.FormatCloseMessage(websocket.CloseNormalClosure, "")
+		conn.WriteControl(websocket.CloseMessage, message, time.Now().Add(time.Second))
```
tigrannajaryan, line 120: What does a deadline of one second do? Is this the maximum time allowed for the message to reach our peer and to be acked? Seems too low in that case.

client/wsclient.go
```diff
@@ -242,26 +308,29 @@ func (c *wsClient) runOneCycle(ctx context.Context) {
...
 	procCancel()
 
-	// If we exited receiverLoop it means there is a connection error, we cannot
-	// read messages anymore. We need to start over.
+	// Wait for WSSender to stop.
+	c.sender.WaitToStop()
 
-	// Close the connection to unblock the WSSender as well.
+	// Close the connection.
 	_ = c.conn.Close()
```
tigrannajaryan, line 322: I am not sure about the swapping of the order of "close" vs "wait". The comment "Close the connection to unblock the WSSender as well" seems to indicate closing is necessary to ensure wait complete. Is that the case or the comment was misleading?

client/wsclient.go
```diff
@@ -80,13 +92,40 @@ func (c *wsClient) Start(ctx context.Context, settings types.StartSettings) erro
...
+			err := defaultCloseHandler(code, text)
+			closed <- struct{}{}
+			return err
+		})
+
+		message := websocket.FormatCloseMessage(websocket.CloseNormalClosure, "")
+		conn.WriteControl(websocket.CloseMessage, message, time.Now().Add(time.Second))
+
+		select {
+		case <-time.After(3 * time.Second):
```
tigrannajaryan, line 123: Why specifically 3 seconds? Is this expected to do a network roundtrip? It seems too low in that case.

client/wsclient.go
```diff
@@ -80,13 +92,40 @@ func (c *wsClient) Start(ctx context.Context, settings types.StartSettings) erro
...
 	// Close connection if any.
 	c.connMutex.RLock()
 	conn := c.conn
 	c.connMutex.RUnlock()
 
 	if conn != nil {
-		_ = conn.Close()
+		// Shut down the sender and any other background processors.
+		c.stopProcessors <- struct{}{}
+		<-c.processorsStopped
```
tigrannajaryan, line 105: Can this get stuck or take a very long time? Should we check for `<-ctx.Done()` too?

client/wsclient.go
evan-bradley: I meant to revisit this, but I mostly kept this at a low value since it's just for acking that the connection is shut down and I couldn't find a standard timeout to use. I agree it could be too low, do you think a higher value like 10 seconds would make sense?

client/wsclient.go
evan-bradley: I may have misunderstood it, but to me the comment was misleading. Closing the connection doesn't appear to have any effect on the WSSender since the only way to shut down WSSender is by [canceling the context](https://github.com/open-telemetry/opamp-go/blob/main/client/internal/wssender.go#L59), which is done by calling `procCancel` in this function.

I mostly reversed these for two reasons:
1. This allows us to clean up the application-level side of the connection before shutting down the transport-level component, which feels like the proper order of operations to me.
2. In the case where we had an error on receiving (e.g. we got bad proto data) and we're looking to reset the connection, this will allow us to send any remaining messages before closing it.

client/wsclient.go
evan-bradley: This is the time the WebSocket client will wait for other writers that have a lock on the underlying connection to release their lock before aborting sending the control message. From what I can tell, writers will release their lock on the socket once it is written to without regard for whether the other end receives it. I'll raise it for now in case a client sends an enormous payload before shutting down, but for the average case I think roughly a second should be ample. Do you have any suggestions for what a good value might be here?

client/wsclient.go
```diff
@@ -97,22 +161,37 @@ func (c *wsClient) AgentDescription() *protobufs.AgentDescription {
...
 func (c *wsClient) UpdateEffectiveConfig(ctx context.Context) error {
+	if c.isStopped.Load() {
+		return errStopping
+	}
 	return c.common.UpdateEffectiveConfig(ctx)
 }
 
 func (c *wsClient) SetRemoteConfigStatus(status *protobufs.RemoteConfigStatus) error {
+	if c.isStopped.Load() {
+		return errStopping
```
srikanthccv, line 186: This can be addressed separately, but should the HTTP Client be updated to return an error on `SetX` calls after `Stop` is called?

client/wsclient.go
```diff
@@ -80,13 +104,53 @@ func (c *wsClient) Start(ctx context.Context, settings types.StartSettings) erro
 }
 
 func (c *wsClient) Stop(ctx context.Context) error {
+	c.isStopped.Store(true)
```
srikanthccv, line 107: Can we rely on the `isStoppingFlag` from ClientCommon? I think the intent behind that flag and this is the same.

client/wsclient.go
```diff
@@ -220,12 +309,22 @@ func (c *wsClient) runOneCycle(ctx context.Context) {
 	// Create a cancellable context for background processors.
 	procCtx, procCancel := context.WithCancel(ctx)
 
+	// Stop processors if we receive a signal to do so.
+	go func() {
+		select {
+		case <-c.stopProcessors:
+			procCancel()
+			c.sender.WaitToStop()
+			c.processorsStopped <- struct{}{}
```
srikanthccv, line 318: Maybe add a comment that `stopProcessors` doesn't immediately stop the receiver processor but it is expected to read the last message with a close error and exit.

client/wsclient.go
evan-bradley: Agreed, we should be consistent. I've created an issue to track this here: https://github.com/open-telemetry/opamp-go/issues/206.

client/wsclient.go
evan-bradley: I tried to call this out in other places, but it's also fair to call it out here. I also renamed the struct fields to make their intent more clear. Let me know if this is still unclear.

client/wsclient.go
srikanthccv: Ah, I understand the complication it brings. This is fine.

client/wsclient.go
```diff
@@ -80,13 +104,53 @@ func (c *wsClient) Start(ctx context.Context, settings types.StartSettings) erro
...
+			return errEarlyStop
+		}
+
+		defaultCloseHandler := conn.CloseHandler()
+		closed := make(chan struct{})
+
+		// The server should respond with a close message of its own, which will
+		// trigger this callback. At this point the close sequence has been
+		// completed and the TCP connection can be gracefully closed.
+		conn.SetCloseHandler(func(code int, text string) error {
```
tigrannajaryan, line 130: Same concurrency question for `SetCloseHandler`.

client/wsclient.go
```diff
@@ -80,13 +104,53 @@ func (c *wsClient) Start(ctx context.Context, settings types.StartSettings) erro
...
+			err := defaultCloseHandler(code, text)
+			closed <- struct{}{}
+			return err
+		})
+
+		message := websocket.FormatCloseMessage(websocket.CloseNormalClosure, "")
+		err := conn.WriteControl(websocket.CloseMessage, message, time.Now().Add(c.connShutdownTimeout))
```
tigrannajaryan, line 137: Do we have a guarantee that `conn` is not written concurrently by the background processors?

client/wsclient.go
```diff
@@ -97,22 +161,37 @@ func (c *wsClient) AgentDescription() *protobufs.AgentDescription {
 }
 
 func (c *wsClient) SetAgentDescription(descr *protobufs.AgentDescription) error {
+	if c.isStopped.Load() {
+		return errStopping
+	}
```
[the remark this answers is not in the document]
evan-bradley: > What is the purpose of this check?
This check disallows client implementations from attempting to send messages after they have called `Stop`.

>What happens if Stop() is called immediately after this check, is there a danger of race?
There is, that's a great observation. We need to inhibit stopping for the duration of these methods to prevent the messages from being lost without an error being returned. I've updated the concurrency control mechanisms to account for this.

client/wsclient.go
```diff
@@ -80,13 +104,53 @@ func (c *wsClient) Start(ctx context.Context, settings types.StartSettings) erro
...
+
 	// Close connection if any.
 	c.connMutex.RLock()
 	conn := c.conn
 	c.connMutex.RUnlock()
 
 	if conn != nil {
-		_ = conn.Close()
+		// Shut down the sender and any other background processors.
+		c.stopBGProcessors <- struct{}{}
```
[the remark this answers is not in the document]
evan-bradley: The goal of this channel is to communicate between `Stop` and `runOneCycle`, since `runOneCycle` "owns" the background processors. I've renamed the channels and updated the comments to hopefully clarify this. Sending a message to the channel should be sufficient, but I've updated this line to close the channel to indicate that this is the last message on the channel.

client/wsclient.go
```diff
@@ -80,13 +104,53 @@ func (c *wsClient) Start(ctx context.Context, settings types.StartSettings) erro
...
+		select {
+		case <-c.bgProcessorsStopped:
+		case <-ctx.Done():
+			_ = c.conn.Close()
```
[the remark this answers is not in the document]
evan-bradley: Good catch, not sure where I missed that. I've updated all occurrences of this to lock.

client/wsclient.go
```diff
@@ -80,13 +104,53 @@ func (c *wsClient) Start(ctx context.Context, settings types.StartSettings) erro
...
+			err := defaultCloseHandler(code, text)
+			closed <- struct{}{}
+			return err
+		})
+
+		message := websocket.FormatCloseMessage(websocket.CloseNormalClosure, "")
+		err := conn.WriteControl(websocket.CloseMessage, message, time.Now().Add(c.connShutdownTimeout))
+
+		if err != nil {
+			_ = c.conn.Close()
```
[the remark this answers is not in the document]
evan-bradley: Again good catch, it should be fixed now.

client/wsclient.go
```diff
@@ -80,13 +104,53 @@ func (c *wsClient) Start(ctx context.Context, settings types.StartSettings) erro
...
+		// Shut down the sender and any other background processors.
+		c.stopBGProcessors <- struct{}{}
+		select {
+		case <-c.bgProcessorsStopped:
+		case <-ctx.Done():
+			_ = c.conn.Close()
+			return errEarlyStop
+		}
+
+		defaultCloseHandler := conn.CloseHandler()
```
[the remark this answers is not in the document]
evan-bradley: `CloseHandler` only returns the current close handler, it doesn't perform any reads from or writes to the connection. Our goal here is to wrap the close handler so we can signal when the WebSocket library is done closing.

client/wsclient.go
evan-bradley: `SetCloseHandler` only sets the close handler function, which will be called later when the client receives a close message from the server. We know that all writes have stopped once `bgProcessingStopped` has fired, which we wait on above in this method.

client/wsclient.go
evan-bradley: We should. `bgProcessingStopped` will emit a message when all background processors have stopped, and we wait on that above before starting the closing handshake.

client/wsclient.go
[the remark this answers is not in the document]
evan-bradley: As a follow-up question, do we expect there will ever be more background processors? Right now it's just the sender, which seems like it should be sufficient for the foreseeable future. If we don't think there will be more, I will make the naming a little more specific.

evan-bradley:
@srikanthccv @tigrannajaryan I'm growing increasingly concerned that the current implementation may be unnecessarily complex and hard to follow. I've done my best to leave comments explaining it, but it's possible I should revisit this as a larger refactor of how we manage the transport layer connection, readers, and writers. I still think what's here is an improvement over the way the connection is currently closed, but if there are concerns over verifying it's free of races, deadlocks, etc. we can look at other approaches.

Let me know your thoughts, maybe this is something we discuss at the next working group meeting.

tigrannajaryan:
> I'm growing increasingly concerned that the current implementation may be unnecessarily complex and hard to follow.

@evan-bradley this is my feeling as well. We could try to finalize this PR to handle this particular issue but I am also concerned that the codebase is not easy to understand and prove that it is correct, especially from concurrency perspective.

I would not mind a refactoring to make the code more understandable.

evan-bradley:
I'm closing this for now in favor of addressing this more holistically by refactoring the client. I believe an approach of controlling the sender, receiver, and transport from separate goroutines that communicate with channels will likely be easiest to reason about and map well to how Go handles concurrency in the standard library. I'm not sure whether this approach is viable or whether I will be able to attempt it soon.

PR #209, opened 2023-10-25
Add support for CustomCapabilities and CustomMessage

andykellr:
Implements the recent additions to the spec described in https://github.com/open-telemetry/opamp-spec/pull/132

### Client 

CustomCapabilities are added to types.StartSettings and these are sent to the server with ReportFullState or when SetCustomCapabilities is called which will update them. There is currently no separate flag for the server to request them. 

SetCustomMessage is added to the OpAMPClient interface and will add the CustomMessage to the next message and schedule a send. It reports ErrCustomCapabilityNotSupported if the Capability in the CustomMessage is not supported. This is to ensure that the client sets CustomCapabilities appropriately.

### Server

The server implementation is more basic. CustomCapabilities are added to server.Settings and are sent on the first ServerToAgent message (websocket) or on every response (http). I considered adding a flag to allow the client to request them and to avoid returning them on every http response, but I decided to wait for feedback.

CustomMessage is available on the ServerToAgent message but it is up to the server implementation to set it on the message before returning from OnMessage or when calling Send.

### Usage

I am using this implementation on a branch of BindPlane OP and bindplane-agent and it is working well.


client/client.go
```diff
@@ -80,4 +80,12 @@ type OpAMPClient interface {
...
 	// nil values are not allowed and will return an error.
 	SetPackageStatuses(statuses *protobufs.PackageStatuses) error
+
+	// SetCustomMessage sets the custom message that will be sent to the Server. May be
+	// called anytime after Start(), including from OnMessage handler.
+	//
+	// If the CustomMessage is nil or it specifies a capability that is not listed in the
+	// CustomCapabilities provided in the StartSettings for the client, it will return an
+	// error.
+	SetCustomMessage(message *protobufs.CustomMessage) error
```
tigrannajaryan, line 90: Does this have the semantics as the other "Set" functions? Does it overwrite a previously set value if it was not already sent? I am not sure this will result in an ergonomic API. If you want to send multiple messages how would you know that the previous one is already sent?

andykellr: Good feedback. No, it schedules a send like the other APIs, but it does not guarantee that one has not already been added to the next message. I suppose we could add a queue that we drain or use a channel. Did you have something in mind?

andykellr: Thanks, I'll revisit this function.

andykellr:
I updated the Client SetCustomMessage to return an `ErrCustomMessagePending` error and a channel that can be used to wait for the message to be sent if there is already a custom message on the AgentToServer message.

```go
	// SetCustomMessage sets the custom message that will be sent to the Server. May be
	// called anytime after Start(), including from OnMessage handler.
	//
	// If the CustomMessage is nil, ErrCustomMessageMissing will be returned. If the message
	// specifies a capability that is not listed in the CustomCapabilities provided in the
	// StartSettings for the client, ErrCustomCapabilityNotSupported will be returned.
	//
	// Only one message can be set at a time. If a message has already been set, it will
	// return ErrCustomMessagePending. To ensure that it is safe to set another
	// CustomMessage, the caller should wait for the channel to be closed before attempting
	// to set another custom message.
	SetCustomMessage(message *protobufs.CustomMessage) (messageSendingChannel chan struct{}, err error)
```

client/client.go
```diff
@@ -89,4 +89,17 @@ type OpAMPClient interface {
...
+	// SetCustomMessage sets the custom message that will be sent to the Server. May be
+	// called anytime after Start(), including from OnMessage handler.
+	//
+	// If the CustomMessage is nil, ErrCustomMessageMissing will be returned. If the message
+	// specifies a capability that is not listed in the CustomCapabilities provided in the
+	// StartSettings for the client, ErrCustomCapabilityNotSupported will be returned.
+	//
+	// Only one message can be set at a time. If a message has already been set, it will
+	// return ErrCustomMessagePending. To ensure that it is safe to set another
+	// CustomMessage, the caller should wait for the channel to be closed before attempting
```
tigrannajaryan, line 102:
```suggestion
	// CustomMessage, the caller should wait for the returned channel to be closed before attempting
```

client/internal/nextmessage.go
```diff
@@ -12,26 +12,31 @@ import (
...
-		nextMessage: &protobufs.AgentToServer{},
+		nextMessage:        &protobufs.AgentToServer{},
+		nextMessageSending: make(chan struct{}),
 	}
 }
 
 // Update applies the specified modifier function to the next message that
 // will be sent and marks the message as pending to be sent.
-func (s *NextMessage) Update(modifier func(msg *protobufs.AgentToServer)) {
+func (s *NextMessage) Update(modifier func(msg *protobufs.AgentToServer)) (messageSendingChannel chan struct{}) {
```
tigrannajaryan, line 33: Add an explanation of how messageSendingChannel works to the comment above.

client/client.go
```diff
@@ -89,4 +89,17 @@ type OpAMPClient interface {
...
+	// to set another custom message.
+	SetCustomMessage(message *protobufs.CustomMessage) (messageSendingChannel chan struct{}, err error)
```
tigrannajaryan, line 104: Do you think it is worth calling this `SendCustomMessage` instead? I think the semantics are now very much what you would expect a non-buffering, non-blocking "send" operation to do.

client/types/startsettings.go
```diff
@@ -44,6 +44,13 @@ type StartSettings struct {
 	// be set in this field, it will be set automatically since it is required by OpAMP protocol.
 	Capabilities protobufs.AgentCapabilities
 
+	// Defines the custom capabilities of the Agent. Each capability is a reverse FQDN with
+	// optional version information that uniquely identifies the custom capability and
+	// should match a capability specified in a supported CustomMessage. The client will
+	// automatically ignore any CustomMessage that contains a custom capability that is not
+	// specified in this field.
+	CustomCapabilities []string
```
tigrannajaryan, line 52: Do we want to specify more precisely how FQDN and optional version is encoded in the string? Perhaps link to spec for that?

tigrannajaryan:
SetCustomMessage and the use of channel looks good to me.

client/types/startsettings.go
andykellr: Added a link.

client/client.go
andykellr: Changed to `SendCustomMessage`.

client/client.go
```diff
@@ -89,4 +89,17 @@ type OpAMPClient interface {
...
+
+	// SendCustomMessage sends the custom message to the Server. May be called anytime after
+	// Start(), including from OnMessage handler.
+	//
+	// If the CustomMessage is nil, ErrCustomMessageMissing will be returned. If the message
+	// specifies a capability that is not listed in the CustomCapabilities provided in the
+	// StartSettings for the client, ErrCustomCapabilityNotSupported will be returned.
+	//
+	// Only one message can be sent at a time. If a message has already been sent, it will
+	// return ErrCustomMessagePending. To ensure that it is safe to send another
```
tigrannajaryan, line 101:
```suggestion
	// Only one message can be sent at a time. If SendCustomMessage has been already called by the message sending is still pending (in progress) then a subsequent call to SendCustomMessage will
	// return ErrCustomMessagePending, until the message is sending is complete. To ensure that previous sending is complete and it is safe to send another
```

client/internal/clientcommon.go
```diff
@@ -369,3 +375,47 @@ func (c *ClientCommon) SetPackageStatuses(statuses *protobufs.PackageStatuses) e
 
 	return nil
 }
+
+// SetCustomCapabilities sends a message to the Server with the new custom capabilities.
+func (c *ClientCommon) SetCustomCapabilities(customCapabilities *protobufs.CustomCapabilities) error {
```
tigrannajaryan, line 380: Is this func used anywhere? I can't find any usage.

server/server.go
```diff
@@ -18,6 +18,11 @@ type Settings struct {
 	// the compression is only effectively enabled if the client also supports compression.
 	// The data will be compressed in both directions.
 	EnableCompression bool
+
+	// Defines the custom capabilities of the Server. Each capability is a reverse FQDN with
+	// optional version information that uniquely identifies the custom capability and
+	// should match a capability specified in a supported CustomMessage.
```
tigrannajaryan, line 24: Add a link to the spec here as well.

client/clientimpl_test.go
```diff
@@ -1549,3 +1550,373 @@ func TestMissingPackagesStateProvider(t *testing.T) {
...
+				if !ok || msg == nil {
+					return false
+				}
+				return proto.Equal(customMessage2, msg)
+			},
+		)
+	})
+}
+
+// TestStreamCustomMessages tests the ability to send may custom messages in succession.
```
evan-bradley, line 1849:
```suggestion
// TestStreamCustomMessages tests the ability to send many custom messages in succession.
```

evan-bradley (commented):
Overall looks good to me, using the channel returned from `SendCustomMessage` looks like a straightfoward way to manage concurrency based on the usage in the tests.

client/internal/clientcommon.go
andykellr: I'm not sure how I missed this. Added to the Client interface, implemented and tested.

client/client.go
```diff
@@ -89,4 +89,29 @@ type OpAMPClient interface {
...
+
+	// SetCustomCapabilities modifies the set of customCapabilities supported by the client.
+	// The new customCapabilities will be sent with the next message to the server.
+	// May be called anytime after Start(), including from OnMessage handler.
+	// nil values are not allowed and will return an error.
+	//
+	// See
+	// https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#customcapabilities
+	// for more details.
+	SetCustomCapabilities(customCapabilities *protobufs.CustomCapabilities) error
```
tigrannajaryan, line 101: Do we need to be able to change custom capabilities after Start? For standard capabilities we allow specifying them in StartSettings only and can't change after Start. Any reason not to do that for custom one's as well?

andykellr: It's easy to allow them to be changed and I think it's useful to support it. Agents may support changing configuration or adding packages that modify (enable or disable) custom capabilities.

tigrannajaryan: I am worried about introducing races here. Let's say client removes a capability but before it gets delivered to the server the server sends a custom message of that capability. Do we expect client to ignore that? What about the opposite situation? We add a capability but before it is delivered a message from the server arrives but the client ignores it.

This may end up resulting in surprises and I don't think we have a good way to prevent them. There is just no way for the client to know the changed capabilities set is delivered and processed by the server so the client has to always assume the server may send custom messages that don't match what the client expects.

andykellr: I appreciate your concerns.

> Let's say client removes a capability but before it gets delivered to the server the server sends a custom message of that capability.

The client will ignore any messages that are not supported by the client. If the server is not yet aware that a message is no longer supported, it may send still send messages, but they will be ignored. If a server needs to know if a message was successfully received and processed, the custom capability should be implemented to send a response and the server will receive confirmation that the message was received.

> What about the opposite situation? We add a capability but before it is delivered a message from the server arrives but the client ignores it.

As soon as the client sets the message, the client will receive messages with that capability. This is because SetCustomCapabilities modifies ClientSyncedState and HasCustomCapability uses that field to determine if the message should be delivered. That said, I think that messages with the capability would only be sent once the server has received the updated list of supported capabilities.

I see two primary patterns for custom messages: streaming messages and request/response messages. It is up to the implementer to choose and implement the appropriate pattern.

For streaming messages (server to agent or agent to server), the only risk is that the streaming is delayed until the capability is known to be supported or that some messages are ignored until the capability is known to not be supported.

For request/response style messages, the sender should expect a response to its message and can handle that as appropriate.

andykellr: @tigrannajaryan I removed CustomCapabilities from the client StartSettings.

client/client.go
```diff
@@ -89,4 +89,36 @@ type OpAMPClient interface {
 	// May be called before or after Start().
 	// May be also called from OnMessage handler.
 	RequestConnectionSettings(request *protobufs.ConnectionSettingsRequest) error
+
+	// SetCustomCapabilities modifies the set of customCapabilities supported by the client.
+	// The new customCapabilities will be sent with the next message to the server. SHOULD
```
tigrannajaryan, line 94: Do we want to a "SHOULD" here? If someone is not using custom capabilities then they don't need to call this. Perhaps we can phrase a bit differently "If custom capabilities are used SHOULD be called before Start()". What do you think?

client/client.go
```diff
@@ -89,4 +89,36 @@ type OpAMPClient interface {
...
+	// specifies a capability that is not listed in the CustomCapabilities provided in the
+	// StartSettings for the client, ErrCustomCapabilityNotSupported will be returned.
+	//
+	// Only one message can be sent at a time. If SendCustomMessage has been already called
+	// and the message is still pending (in progress) then subsequent calls to
+	// SendCustomMessage will return ErrCustomMessagePending. To ensure that the previous
+	// send is complete and it is safe to send another CustomMessage, the caller should wait
+	// for the returned channel to be closed before attempting to send another custom
+	// message.
+	SendCustomMessage(message *protobufs.CustomMessage) (messageSendingChannel chan struct{}, err error)
```
tigrannajaryan, line 123: If an error is returned do we expect that messageSendingChannel is nil (or do we care?). Do you think we need to say anything about it or what we have is enough?

client/clientimpl_test.go
```diff
@@ -1549,3 +1550,470 @@ func TestMissingPackagesStateProvider(t *testing.T) {
...
+		// Shutdown the Server.
+		srv.Close()
+
+		// Shutdown the client.
+		err = client.Stop(context.Background())
+		assert.NoError(t, err)
+	})
+}
+
+func TestSendCustomMessageConflict(t *testing.T) {
```
tigrannajaryan, line 1784: Nit: please clarify why this is called "Conflict". I am not sure I see any conflicts, seems like a good test that demonstrates pending channel use and misuse.

client/clientimpl_test.go
```diff
@@ -1549,3 +1550,470 @@ func TestMissingPackagesStateProvider(t *testing.T) {
...
+					return false
+				}
+				return proto.Equal(customMessage2, msg)
+			},
+		)
+	})
+}
+
+// TestStreamCustomMessages tests the ability to send many custom messages in succession.
+func TestCustomMessagesSendAndWait(t *testing.T) {
```
tigrannajaryan, line 1861: Do we also need a similar test but for opposite direction, with many messages from the Server to the Client? Or perhaps a combined one with "send request/wait response" style of communication?

tigrannajaryan (commented):
Mostly LGTM, only some minor questions/comments remaining.

client/client.go
andykellr: If the error is `ErrCustomMessagePending`, then the messageSendingChannel is non-nil and the caller can wait on that channel before sending again. It is only currently nil if `ErrCustomMessageMissing` (nil message) or `ErrCustomCapabilityNotSupported` (unsupported message) are returned. I will tweak the comment to make that clear.

client/clientimpl_test.go
andykellr: I'll tweak the name. `ErrCustomMessagePending` used to be called `ErrCustomMessageConflict` and the test name made more sense then.

client/clientimpl_test.go
andykellr: Since the server implementation only uses the Callbacks and Connection interfaces, any waiting or queueing is left to the implementer of the OpAMP server. Therefore there isn't as much to test other than that when Send is called with a CustomMessage that the message is received by the client.

`TestReportCustomCapabilities` includes some request/response style communication, but I'll improve it or add a new test.

client/client.go
andykellr: Updated. Let me know if that is more clear.

client/clientimpl_test.go
andykellr: Let me know if you think the existing `TestReportCustomCapabilities` message is sufficient or if I should add a test that sends a bunch of messages from server => client.

tigrannajaryan (approved):
LGTM

PR #211, opened 2023-11-11
fix sample not works

JaredTan95:
fix `ui` dir cloud not be loaded and `certs` file path  in sample server.


srikanthccv:
What was the directory you used to run the example server?

JaredTan95:
> What was the directory you used to run the example server?

Thank you for reminding me that I was in the wrong directory. So, I will close this PR.

srikanthccv:
You can rely on the `run-examples` target without worrying about directory stuff.

PR #213, opened 2023-11-21
Gracefully shutdown of the websocket client

### LINKED ISSUE #163: wsclient: Unexpected error while receiving:read ... use of closed network connection

haoqixu:
Resolves #163

### DIFF
--- a/client/internal/wsreceiver.go
+++ b/client/internal/wsreceiver.go
@@ -63,6 +63,8 @@ func (r *wsReceiver) ReceiverLoop(ctx context.Context) {
 		err     error
 	}
 
+	defer func() { close(r.stopped) }()
+
 	for {
 		select {
 		case <-ctx.Done():
@@ -70,6 +72,7 @@ func (r *wsReceiver) ReceiverLoop(ctx context.Context) {
 		default:
 			result := make(chan receivedMessage, 1)
 
+			// To stop this goroutine, close the websocket connection
 			go func() {
 				var message protobufs.ServerToAgent
 				err := r.receiveMessage(&message)
--- a/client/internal/wsreceiver.go
+++ b/client/internal/wsreceiver.go
@@ -58,23 +58,38 @@ func (r *wsReceiver) IsStopped() <-chan struct{} {
 // ReceiverLoop runs the receiver loop.
 // To stop the receiver cancel the context and close the websocket connection
 func (r *wsReceiver) ReceiverLoop(ctx context.Context) {
-	processorCtx, stopProcessor := context.WithCancel(ctx)
-
-out:
-	for {
-		var message protobufs.ServerToAgent
-		if err := r.receiveMessage(&message); err != nil {
-			if ctx.Err() == nil && !websocket.IsCloseError(err, websocket.CloseNormalClosure) {
-				r.logger.Errorf(ctx, "Unexpected error while receiving: %v", err)
-			}
-			break out
-		} else {
-			r.processor.ProcessReceivedMessage(processorCtx, &message)
-		}
+	type receivedMessage struct {
+		message *protobufs.ServerToAgent
+		err     error
 	}
 
-	stopProcessor()
-	close(r.stopped)
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		default:
+			result := make(chan receivedMessage, 1)
+
+			go func() {
+				var message protobufs.ServerToAgent
+				err := r.receiveMessage(&message)
+				result <- receivedMessage{&message, err}
+			}()
+
+			select {
+			case <-ctx.Done():
+				return
+			case res := <-result:
+				if res.err != nil {
+					if !websocket.IsCloseError(res.err, websocket.CloseNormalClosure) {
+						r.logger.Errorf(ctx, "Unexpected error while receiving: %v", res.err)
+					}
+					return
+				}
+				r.processor.ProcessReceivedMessage(ctx, res.message)
+			}
+		}
+	}
 }
 
 func (r *wsReceiver) receiveMessage(msg *protobufs.ServerToAgent) error {
--- a/client/internal/wsreceiver.go
+++ b/client/internal/wsreceiver.go
@@ -17,6 +17,9 @@ type wsReceiver struct {
 	sender    *WSSender
 	callbacks types.Callbacks
 	processor receivedProcessor
+
+	// Indicates that the receiver has fully stopped.
+	stopped chan struct{}
 }
 
 // NewWSReceiver creates a new Receiver that uses WebSocket to receive
@@ -36,45 +39,42 @@ func NewWSReceiver(
 		sender:    sender,
 		callbacks: callbacks,
 		processor: newReceivedProcessor(logger, callbacks, sender, clientSyncedState, packagesStateProvider, capabilities),
+		stopped:   make(chan struct{}),
 	}
 
 	return w
 }
 
-// ReceiverLoop runs the receiver loop. To stop the receiver cancel the context.
+// Start starts the receiver loop.
+func (r *wsReceiver) Start(ctx context.Context) {
+	go r.ReceiverLoop(ctx)
+}
+
+// IsStopped returns a channel that's closed when the receiver is stopped.
+func (r *wsReceiver) IsStopped() <-chan struct{} {
+	return r.stopped
+}
+
+// ReceiverLoop runs the receiver loop.
+// To stop the receiver cancel the context and close the websocket connection
 func (r *wsReceiver) ReceiverLoop(ctx context.Context) {
-	type receivedMessage struct {
-		message *protobufs.ServerToAgent
-		err     error
-	}
+	processorCtx, stopProcessor := context.WithCancel(ctx)
 
+out:
 	for {
-		select {
-		case <-ctx.Done():
-			return
-		default:
-			result := make(chan receivedMessage, 1)
-
-			go func() {
-				var message protobufs.ServerToAgent
-				err := r.receiveMessage(&message)
-				result <- receivedMessage{&message, err}
-			}()
-
-			select {
-			case <-ctx.Done():
-				return
-			case res := <-result:
-				if res.err != nil {
-					if !websocket.IsCloseError(res.err, websocket.CloseNormalClosure) {
-						r.logger.Errorf(ctx, "Unexpected error while receiving: %v", res.err)
-					}
-					return
-				}
-				r.processor.ProcessReceivedMessage(ctx, res.message)
+		var message protobufs.ServerToAgent
+		if err := r.receiveMessage(&message); err != nil {
+			if ctx.Err() == nil && !websocket.IsCloseError(err, websocket.CloseNormalClosure) {
+				r.logger.Errorf(ctx, "Unexpected error while receiving: %v", err)
 			}
+			break out
+		} else {
+			r.processor.ProcessReceivedMessage(processorCtx, &message)
 		}
 	}
+
+	stopProcessor()
+	close(r.stopped)
 }
 
 func (r *wsReceiver) receiveMessage(msg *protobufs.ServerToAgent) error {
--- a/client/internal/wssender.go
+++ b/client/internal/wssender.go
@@ -49,13 +49,6 @@ func (s *WSSender) Start(ctx context.Context, conn *websocket.Conn) error {
 	return err
 }
 
-// WaitToStop blocks until the sender is stopped. To stop the sender cancel the context
-// that was passed to Start().
-func (s *WSSender) WaitToStop() error {
-	<-s.stopped
-	return s.err
-}
-
 func (s *WSSender) IsStopped() <-chan struct{} {
 	return s.stopped
 }
--- a/client/internal/wssender.go
+++ b/client/internal/wssender.go
@@ -55,8 +55,10 @@ func (s *WSSender) IsStopped() <-chan struct{} {
 	return s.stopped
 }
 
-// Err returns the error of sending close message.
-func (s *WSSender) Err() error {
+// StoppingErr returns an error if there was a problem with stopping the sender.
+// If stopping was successful will return nil.
+// StoppingErr() can be called only after IsStopped() is signalled.
+func (s *WSSender) StoppingErr() error {
 	return s.err
 }
 
--- a/client/internal/wssender.go
+++ b/client/internal/wssender.go
@@ -50,10 +50,12 @@ func (s *WSSender) Start(ctx context.Context, conn *websocket.Conn) error {
 	return err
 }
 
+// IsStopped returns a channel that's closed when the sender is stopped.
 func (s *WSSender) IsStopped() <-chan struct{} {
 	return s.stopped
 }
 
+// Err returns the error of sending close message.
 func (s *WSSender) Err() error {
 	return s.err
 }
--- a/client/internal/wssender.go
+++ b/client/internal/wssender.go
@@ -68,7 +68,9 @@ out:
 			s.sendNextMessage(ctx)
 
 		case <-ctx.Done():
-			s.err = s.sendCloseMessage()
+			if err := s.sendCloseMessage(); err != nil && err != websocket.ErrCloseSent {
+				s.err = err
+			}
 			break out
 		}
 	}
--- a/client/internal/wssender.go
+++ b/client/internal/wssender.go
@@ -2,6 +2,7 @@ package internal
 
 import (
 	"context"
+	"time"
 
 	"github.com/gorilla/websocket"
 	"google.golang.org/protobuf/proto"
@@ -11,13 +12,19 @@ import (
 	"github.com/open-telemetry/opamp-go/protobufs"
 )
 
+const (
+	defaultSendCloseMessageTimeout = 5 * time.Second
+)
+
 // WSSender implements the WebSocket client's sending portion of OpAMP protocol.
 type WSSender struct {
 	SenderCommon
 	conn   *websocket.Conn
 	logger types.Logger
+
 	// Indicates that the sender has fully stopped.
 	stopped chan struct{}
+	err     error
 }
 
 // NewSender creates a new Sender that uses WebSocket to send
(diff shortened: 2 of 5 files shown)


haoqixu:
Could you please take a look @evan-bradley @srikanthccv @tigrannajaryan

haoqixu:
I have watched the meeting record and I think I can add some explanation about this PR.

The solution here is making `wsclient.runOneCycle` responsible for the entire lifecyle of the connection and controlling the running of sender and receiver. It also move the receiver to run in a separate goroutine. `wsclient.runOneCycle` will ensure that both sender and receiver are stopped and the connection is closed before it returns. So we don't need to care about the connection outside `wsclient.runOneCycle`.

When the client is shutting down, the `sender` goroutine will be cancelled by the ctx, send the close message to server and return the error to `wsclient.runOneCycle` via sender.Err(). `wsclient.runOneCycle` will wait for the close message from server until timeout and ensure the connection is closed.

If the server wants to close the connection, it should send the close message to initiate the close handshake. When a close message is received, the connection read methods will handle it by calling the close handler (the default handler will send a close message back to the peer. I think this is what we want.) and return a CloseError. The receiver loop will stop on any error. `wsclient.runOneCycle` will stop the sender and close the connection when receiver is stopped.

client/wsclient.go
```diff
@@ -245,19 +250,33 @@ func (c *wsClient) runOneCycle(ctx context.Context) {
...
-	procCancel()
-
-	// If we exited receiverLoop it means there is a connection error, we cannot
-	// read messages anymore. We need to start over.
+	r.Start(ctx)
+
+	select {
+	case <-c.sender.IsStopped():
+		// sender will send close message to initiate the close handshake
+		if err := c.sender.Err(); err != nil && err != websocket.ErrCloseSent {
```
srikanthccv, line 258: Is this `err != websocket.ErrCloseSent` to handle the case when the server initiates the close?

client/internal/wssender.go
```diff
@@ -56,13 +66,22 @@ out:
 			s.sendNextMessage()
 
 		case <-ctx.Done():
+			s.err = s.sendCloseMessage()
```
srikanthccv, line 69: This means stopping the sender also closes the connection which might be unexpected because run should simply run until the context is cancelled but it also closes the connection. I think sending Close message should be part of client `Stop`.

haoqixu: This is by design. Stopping the sender would always try to send the close message for the close handshake and `runOneCycle` is responsible for the subsequent steps of connection closing. I updated the comment of the `runOneCycle()`.

client/wsclient.go
haoqixu: Yes. The sender and receiver run concurrently. There is a possibility that the server and client initiate the closing handshake at the same time. In this case, the close message may be sent by CloseHandler, which is called in the receiver.

client/wsclient.go
```diff
@@ -245,19 +250,33 @@ func (c *wsClient) runOneCycle(ctx context.Context) {
...
+			c.common.Logger.Debugf("failed to send close message, close without the handshake.")
+			break
+		}
 
-	// Close the connection to unblock the WSSender as well.
-	_ = c.conn.Close()
+		c.common.Logger.Debugf("waiting for close message from server.")
+		select {
+		case <-r.IsStopped():
+			c.common.Logger.Debugf("shutdown handshake complete.")
```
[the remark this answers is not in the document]
haoqixu: I think the receiver loop has handled it and printed a log for it. However, the log message here is not very accurate. The "complete" here actually means that the receiver loop has also exited. 🤔

srikanthccv (commented):
Overall LGTM, One main difference from this PR https://github.com/open-telemetry/opamp-go/pull/205 is that the sender initiates the close handshake which makes the implementation a little simpler but also the connection management is now done b/w wsclient and wssender. @evan-bradley what do you think?

client/wsclient.go
```diff
@@ -199,19 +199,25 @@ func (c *wsClient) ensureConnected(ctx context.Context) error {
 // runOneCycle performs the following actions:
 //  1. connect (try until succeeds).
 //  2. send first status report.
-//  3. receive and process messages until error happens.
+//  3. start the sender to wait for scheduled message and send it to the server.
+//  4. start the receiver to receive and process messages until error happens.
+//  5. wait until both sender and receiver are stopped.
```
evan-bradley, line 204:
```suggestion
//  3. start the sender to wait for scheduled messages and send them to the server.
//  4. start the receiver to receive and process messages until an error happens.
//  5. wait until both the sender and receiver are stopped.
```
Grammar nits.

client/wsclient.go
```diff
@@ -245,19 +250,33 @@ func (c *wsClient) runOneCycle(ctx context.Context) {
...
+			<-r.IsStopped()
+		}
+	case <-r.IsStopped():
+		// If we exited receiverLoop it means there is a connection error, we cannot
+		// read messages anymore. We need to start over.
 
-	// Wait for WSSender to stop.
-	c.sender.WaitToStop()
+		stopSender()
+		<-c.sender.IsStopped()
```
evan-bradley, line 278: Running `stopSender` will attempt to send a message to the server. Do you know whether this will time out in a reasonable timeframe, or should we race this with `c.connShutdownTimeout` to ensure it doesn't hang?

evan-bradley (commented):
Thanks for doing this, overall it looks good to me. I think this approach is an improvement over https://github.com/open-telemetry/opamp-go/pull/205. There may additional improvements we can make in the future, but for now this will resolve the errors we are seeing and will cleanly close the connection.

client/wsclient.go
haoqixu: `c.connShutdownTimeout` is controlling how long the wsclient will wait for the ACK from server, and the timeout for sendding the close message to the server is [`defaultSendCloseMessageTimeout`](https://github.com/open-telemetry/opamp-go/pull/213/files#diff-1a7764ac131862565b9f9783634aa0c05d383cc44c34268308865908e1c3d02dR16)(5s). So it will not hang here.

`c.connShutdownTimeout` is currently only modified during testing and both of them are not configurable for now. We can improve the Client API to make them configurable later.

srikanthccv (approved):

evan-bradley (approved):

client/wsclient.go
tigrannajaryan: Can we change `sender.Err()` to return nil if the error was `websocket.ErrCloseSent`? Let's consider it an internal implementation detail of the sender and avoid leaking it here. That we only need to check `err` for nil here. Sender seeing a `websocket.ErrCloseSent` response is a success, we don't need to consider it an error.

client/wsclient.go
```diff
@@ -245,19 +250,33 @@ func (c *wsClient) runOneCycle(ctx context.Context) {
...
-
-	// If we exited receiverLoop it means there is a connection error, we cannot
-	// read messages anymore. We need to start over.
+	r.Start(ctx)
+
+	select {
+	case <-c.sender.IsStopped():
+		// sender will send close message to initiate the close handshake
+		if err := c.sender.Err(); err != nil && err != websocket.ErrCloseSent {
+			c.common.Logger.Debugf("failed to send close message, close without the handshake.")
```
tigrannajaryan, line 259: Should we include `err` in the log?

client/wsclient.go
```diff
@@ -245,19 +250,33 @@ func (c *wsClient) runOneCycle(ctx context.Context) {
...
+			break
+		}
 
-	// Close the connection to unblock the WSSender as well.
-	_ = c.conn.Close()
+		c.common.Logger.Debugf("waiting for close message from server.")
```
tigrannajaryan, line 263:
```suggestion
		c.common.Logger.Debugf("Waiting for receiver to stop.")
```
I think receiver waiting for close message is a implementation detail from our perspective. Let's not leak implementation detail. If we want this exact log message then better to move it inside receiver.

client/wsclient.go
tigrannajaryan, line 259:
```suggestion
			c.common.Logger.Debugf("Error stopping the sender: %v", err)
```

client/wsclient.go
tigrannajaryan, line 266:
```suggestion
			c.common.Logger.Debugf(Receiver stopped.")
```

client/wsclient.go
```diff
@@ -245,19 +250,33 @@ func (c *wsClient) runOneCycle(ctx context.Context) {
...
+		select {
+		case <-r.IsStopped():
+			c.common.Logger.Debugf("shutdown handshake complete.")
+		case <-time.After(c.connShutdownTimeout):
+			c.common.Logger.Debugf("timeout waiting for close message.")
```
tigrannajaryan, line 268:
```suggestion
			c.common.Logger.Debugf("Timeout waiting for receiver to stop.")
```

client/wsclient.go
```diff
@@ -245,19 +250,33 @@ func (c *wsClient) runOneCycle(ctx context.Context) {
...
+			// not receive close message from the server, close the connection to force the receive loop to stop
```
tigrannajaryan, line 269:
```suggestion
			// Close the connection to force the receive loop to stop.
```

client/wsclient.go
```diff
@@ -245,19 +250,33 @@ func (c *wsClient) runOneCycle(ctx context.Context) {
...
+			_ = c.conn.Close()
+			<-r.IsStopped()
```
tigrannajaryan, line 271: Can this potentially block forever? To be defensive we probably need to have a timeout here.

client/wsclient.go
tigrannajaryan, line 278: Does this waiting require a timeout?

haoqixu: I don't think it's necessary, as mentioned above(https://github.com/open-telemetry/opamp-go/pull/213/files#r1446914020)

client/wsclient.go
haoqixu: Receiver is waiting to read data from the connection (via `ReadMessage`), I think closing the connection is enough to unblock the `ReadMessage`. 🤔

client/internal/wssender.go
```diff
@@ -56,13 +66,22 @@ out:
...
 	}
 
 	close(s.stopped)
 }
 
+func (s *WSSender) sendCloseMessage() error {
+	return s.conn.WriteControl(
+		websocket.CloseMessage,
+		websocket.FormatCloseMessage(websocket.CloseNormalClosure, "Normal closure"),
+		time.Now().Add(defaultSendCloseMessageTimeout),
```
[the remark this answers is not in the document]
tigrannajaryan: My understanding is that flushing data to a socket can take indefinitely long if the peer is not responsive. It is unfortunately there is no cancellation possible for this call.

client/wsclient.go
tigrannajaryan: I am not sure about it. 

Calling `stopSender()` here triggers context cancellation. That in turn causes the `sender` to call `sendCloseMessage()`. `sendCloseMessage()` can be blocked indefinitely if the peer server is not reading from the socket and there is no room in the local socket buffer. This means the `sender` will never be stopped and `IsStopped()` channel will never be closed. We can block indefinitely long here.

I think we need a timeout here and need to close the connection forcedly like we do it to force the receiver to stop above in line 270.

I may be missing something here. Please help me with the analysis.

client/internal/wsreceiver.go
```diff
@@ -36,14 +40,31 @@ func NewWSReceiver(
...
+	go r.ReceiverLoop(ctx)
+}
+
+// IsStopped returns a channel that's closed when the receiver is stopped.
+func (r *wsReceiver) IsStopped() <-chan struct{} {
+	return r.stopped
+}
+
+// Err returns the error that caused the receiver loop to exit.
+func (r *wsReceiver) Err() error {
```
tigrannajaryan, line 60: I cannot see this function used anywhere anymore. If it is not used please delete and also delete `r.err` field from the struct.

client/internal/wssender.go
```diff
@@ -37,15 +44,20 @@ func (s *WSSender) Start(ctx context.Context, conn *websocket.Conn) error {
...
-// that was passed to Start().
-func (s *WSSender) WaitToStop() {
-	<-s.stopped
+// IsStopped returns a channel that's closed when the sender is stopped.
+func (s *WSSender) IsStopped() <-chan struct{} {
+	return s.stopped
+}
+
+// Err returns the error of sending close message.
+func (s *WSSender) Err() error {
```
tigrannajaryan, line 59:
```suggestion
func (s *WSSender) StoppingErr() error {
```

client/internal/wssender.go
```diff
@@ -37,15 +44,20 @@ func (s *WSSender) Start(ctx context.Context, conn *websocket.Conn) error {
...
-// WaitToStop blocks until the sender is stopped. To stop the sender cancel the context
-// that was passed to Start().
-func (s *WSSender) WaitToStop() {
-	<-s.stopped
+// IsStopped returns a channel that's closed when the sender is stopped.
+func (s *WSSender) IsStopped() <-chan struct{} {
+	return s.stopped
+}
+
+// Err returns the error of sending close message.
```
tigrannajaryan, line 58:
```suggestion
// StoppingErr returns an error if there was a problem with stopping the sender.
// If stopping was successful will return nil.
// StoppingErr() can be called only after IsStopped() is signalled.
```

client/wsclient.go
```diff
@@ -245,19 +250,33 @@ func (c *wsClient) runOneCycle(ctx context.Context) {
...
-	procCancel()
-
-	// If we exited receiverLoop it means there is a connection error, we cannot
-	// read messages anymore. We need to start over.
+	r.Start(ctx)
+
+	select {
+	case <-c.sender.IsStopped():
+		// sender will send close message to initiate the close handshake
+		if err := c.sender.Err(); err != nil {
```
tigrannajaryan, line 258:
```suggestion
		if err := c.sender.StoppingErr(); err != nil {
```

client/wsclient.go
```diff
@@ -245,19 +250,33 @@ func (c *wsClient) runOneCycle(ctx context.Context) {
...
+			c.common.Logger.Debugf("Error stopping the sender: %v", err)
+			break
```
tigrannajaryan, line 260: Is there a reason we `break` here? If the sender stopping failed shouldn't we still try to stop the receiver and close the connection?

tigrannajaryan (commented):
Thanks for your patience @haoqixu 

Please bear with me while I am trying to understand all the concurrent flows.

client/wsclient.go
tigrannajaryan: OK, from the comment thread `sendCloseMessage()` will not blocked indefinitely, it will return at most after `defaultSendCloseMessageTimeout` so this does not need a timeout.

client/wsclient.go
haoqixu: I think this `break` was introduced during the updates of the PR. The initial code does not wait for the receiver to stop, leaving the closing of the connection to the defer. I update the branch to close the conneciton explicitly and wait for the receiver to stop before break.

tigrannajaryan:
Thanks for updates @haoqixu 
The PR mostly looks good to me, the only remaining question is https://github.com/open-telemetry/opamp-go/pull/213/files#r1462064174

haoqixu:
> The PR mostly looks good to me, the only remaining question is https://github.com/open-telemetry/opamp-go/pull/213/files#r1462064174

I have updated the PR with https://github.com/open-telemetry/opamp-go/pull/213/commits/5fa47a3bfe0212121d1031031dcda2362ba094b1

tigrannajaryan:
@haoqixu this looks good to me. Can you please resolve the conflicts?

haoqixu:
#240 puts the `receiveMessage` into a separate goroutine.

haoqixu:
By design, we need to canceling the context and closing the connection to stop the `ReceiverLoop`.  And the gracefully shutdown flow of `wsclient` will do this for us. Considering `ReceiverLoop` as an internal implementation detail of the `wsclient`, we could just gracefully shut down the `wsclient`.

`gorilla/websocket` doesn't provide a method to unblock the reading of connection and we always need to close the connection to unblock the reading.

#240 puts the reading of connections into a separate goroutine, which is unnecessary and may cause goroutine leaks. And the update here actually reverts #240

haoqixu:
please take a look. @tigrannajaryan

tigrannajaryan:
> By design, we need to canceling the context and closing the connection to stop the `ReceiverLoop`. And the gracefully shutdown flow of `wsclient` will do this for us. Considering `ReceiverLoop` as an internal implementation detail of the `wsclient`, we could just gracefully shut down the `wsclient`.
> 
> `gorilla/websocket` doesn't provide a method to unblock the reading of connection and we always need to close the connection to unblock the reading.
> 
> #240 puts the reading of connections into a separate goroutine, which is unnecessary and may cause goroutine leaks. And the update here actually reverts #240

@srikanthccv please take another look at this PR since it reverts your earlier PR #240.

srikanthccv:
We don't make it clear anywhere that `ReceiverLoop` is an internal implementation detail. By making both connection closing and context cancellation (I don't see any use of context in this case) mandatory for exiting the receiver loop we may prematurely close the connection. There would be leaks if `wsReceiver` were to work in isolation but given we will close the connection followed by the receiver loop exit, there shouldn't be goroutine leaks? 

We will need to send a final message(s) such as agent disconnect (not implemented yet) before stopping the client. I'd prefer to see `wsclient` have final say in when to close for such kind of reasons. I want to hear your thoughts on this.

srikanthccv:
I recall we had discussed this already once. The implementation can always evolve to satisfy the needs but We should explicitly call out what we consider as internal implementation details.

tigrannajaryan:
I agree it is more idiomatic to honor the cancellation in ReceiverLoop. I am not sure wha the problem is with the leak. If there is one let's fix it, otherwise we I think we don't need to worry about hypotheticals in this particular case.

srikanthccv:
I am not sure I follow the simplicity part. I don't see much benefit of one over the other. Regarding the memory leak concern - as we are going to gracefully close the connection with this change there should not be any go routine leak or did I miss something?

haoqixu:
I have updated the code. Please take another look.

jaronoff97:
@tigrannajaryan are we good to merge this? I just ran into an instance of this in the opampextension

tigrannajaryan:
Thank you @haoqixu

PR #216, opened 2023-12-01
Fix flaky TestAgentIdentification/ws

### LINKED ISSUE #191: TestAgentIdentification/ws is flaky
See https://github.com/open-telemetry/opamp-go/actions/runs/5487663480/jobs/9999430522

I can sometimes reproduce this locally by doing `go test -run TestAgentIdentification/ws -race -count 1000`

haoqixu:
Resolve #191

TestAgentIdentification executes the following steps to test the client:
1. `prepareClient()`: setup settings and call `SetAgentDescription` of client which schedules an message sending.
2. `client.Start()`:
    send the first message
    (a) launch an goroutine to handle scheduled message sending.
    (b) start the receiver loop to handle messages from server.
3. run the first `eventually` assert.
4. `client.SetAgentDescription()` to schedule an message sending.
5. run the second `eventually` assert.

scheduled message sending in (2.a) and messages handling (2.b) are running concurrently.

Case 1: 
the timeline is as follows:
- (1) calls `c.SetAgentDescription` to schedule an message sending.
- (2) send the first message to the server.
- (2.b) response from server is handled by the receiver loop, which update the instance ID of scheduled message.
- (2.a) handle scheduled message and send it to the server, server store the new instance ID to rcvAgentInstanceUid.
- (3) the first `eventually` assert fails to get the original instance ID

Case 2: 
the timeline is as follows:
- (1) calls `c.SetAgentDescription` to schedule an message sending.
- (2) send the first message to the server.
- (2.a) handle scheduled message. This is noop because there is no update since last sending.
- (3) the first `eventually` assert trys to get the original instance ID and will return successfully after server has stored the instance ID in the first message to rcvAgentInstanceUid.
- (4) `client.SetAgentDescription()` to schedule an message sending
- (2.a) sender goroutine send the scheduled message to the server with original ID (response from server hasn't arrived yet or the message handling hasn't completed)
- (2.b) response from server is handled by the receiver loop, which update the instance ID of scheduled message. But the client has no more chance to send a message.
- (5) the second `eventually` assert failed to get the new instance uid.

The solution here is replacing atomic.Value with a buffered channel to avoid case 1 and waiting for the first response from server to avoid case 2.

### DIFF
--- a/client/clientimpl_test.go
+++ b/client/clientimpl_test.go
@@ -638,7 +638,17 @@ func TestAgentIdentification(t *testing.T) {
 		}
 
 		// Start a client.
-		settings := types.StartSettings{}
+		hasRecvServerResponse := make(chan struct{}, 1)
+		settings := types.StartSettings{
+			Callbacks: types.CallbacksStruct{
+				OnMessageFunc: func(_ context.Context, _ *types.MessageData) {
+					select {
+					case hasRecvServerResponse <- struct{}{}:
+					default:
+					}
+				},
+			},
+		}
 		settings.OpAMPServerURL = "ws://" + srv.Endpoint
 		prepareClient(t, &settings, client)
 
@@ -657,6 +667,9 @@ func TestAgentIdentification(t *testing.T) {
 				return false
 			},
 		)
+
+		// Wait for the server response.
+		<-hasRecvServerResponse
 
 		// Send a dummy message
 		_ = client.SetAgentDescription(createAgentDescr())
--- a/client/clientimpl_test.go
+++ b/client/clientimpl_test.go
@@ -626,9 +626,9 @@ func TestAgentIdentification(t *testing.T) {
 		newInstanceUid := ulid.MustNew(
 			ulid.Timestamp(time.Now()), ulid.Monotonic(rand.New(rand.NewSource(0)), 0),
 		)
-		var rcvAgentInstanceUid atomic.Value
+		rcvAgentInstanceUids := make(chan string, 10)
 		srv.OnMessage = func(msg *protobufs.AgentToServer) *protobufs.ServerToAgent {
-			rcvAgentInstanceUid.Store(msg.InstanceUid)
+			rcvAgentInstanceUids <- msg.InstanceUid
 			return &protobufs.ServerToAgent{
 				InstanceUid: msg.InstanceUid,
 				AgentIdentification: &protobufs.AgentIdentification{
@@ -649,11 +649,12 @@ func TestAgentIdentification(t *testing.T) {
 		eventually(
 			t,
 			func() bool {
-				instanceUid, ok := rcvAgentInstanceUid.Load().(string)
-				if !ok {
-					return false
+				select {
+				case instanceUid := <-rcvAgentInstanceUids:
+					return instanceUid == oldInstanceUid
+				default:
 				}
-				return instanceUid == oldInstanceUid
+				return false
 			},
 		)
 
@@ -665,11 +666,12 @@ func TestAgentIdentification(t *testing.T) {
 		eventually(
 			t,
 			func() bool {
-				instanceUid, ok := rcvAgentInstanceUid.Load().(string)
-				if !ok {
-					return false
+				select {
+				case instanceUid := <-rcvAgentInstanceUids:
+					return instanceUid == newInstanceUid.String()
+				default:
 				}
-				return instanceUid == newInstanceUid.String()
+				return false
 			},
 		)
 


srikanthccv:
I am trying to follow the case 1 timeline you mentioned. 

>Case 1: 
>the timeline is as follows:
>- (1) calls `c.SetAgentDescription` to schedule an message sending.
>- (2) send the first message to the server.
>- (2.b) response from server is handled by the receiver loop, which update the instance ID of scheduled message.
>- (2.a) handle scheduled message and send it to the server, server store the new instance ID to rcvAgentInstanceUid.
>- (3) the first `eventually` assert fails to get the original instance ID



2.b is triggered by the client sending the scheduled message (from 2.a step) so why does the first assertion fail?

haoqixu:
I try to draw a sequence diagram of case 1:
```mermaid
sequenceDiagram
participant Client
participant Client.Receiver
participant Client.Sender
participant Server
    Client->>Client.Sender: (1) c.SetAgentDescription in `prepareClient` call `sender.ScheduleSend()`
    Client->>Server: (2) AgentToServer with original UID
alt
    Server->>Server: store the original instance ID to rcvAgentInstanceUid
end
    Client-->>Client.Sender: start the loop
    Client-->>Client.Receiver: start the loop
    Server->>Client.Receiver: (2.b)ServerToAgent with new UID
    Client.Receiver->>Client.Sender: Update the pendding Message with new UID
    Client.Sender->>Server: (2.a) send the pending scheduled message with new UID
alt Oops!
    Server->>Server: store the new instance ID to rcvAgentInstanceUid
end
   Server ->> Client.Receiver: reply ...
   Client->>Client: run the first eventually assertion and fail
```
2.b will just update the message without schedule the sending. If this happends before sender sending the pending message, rcvAgentInstanceUid will be updated with the new UID.

srikanthccv:
There are two times we can see when the ScheduleSend is invoked which sends the message to the Server. The first one is from the `SetAgentDescription` in prepareClient and then from the `SetAgentDescription` of _Send a dummy message_. The second dummy message is only sent after the first assertion is completed. So my question is why does the first assertion fail because, for rcvAgentInstanceUid to have the new UID, the dummy message should have been sent which can't be true because the dummy message is only sent after the the first asseetion passes. Or am I missing something?

srikanthccv:
I am not sure I follow. What does this PR attempt to fix then? The first assertion is expected to always pass. If you look at the last failed runs in the issue, they point out _Condition never satisfied_ at the first assertion. I think any fix should involve figuring out why it failed.

srikanthccv:
> In the timeline case 1, the server has received the original uid, the first assertion falied becase rcvAgentInstanceUid(atomic.Value) was overwritten by the second AgentToServer message.

How did you arrive at this conclusion? It could be because no value was set to `rcvAgentInstanceUid` and the assertion failed. I don't see how an override is the reason for the first assertion failure unless I am missing something.

haoqixu:
> It could be because no value was set to `rcvAgentInstanceUid` and the assertion failed.

If no value was set to `rcvAgentInstanceUid`, both the `eventually` assertions will failed. In the failed runs in the issue, only the first assertion failed.

haoqixu:
https://github.com/haoqixu/opamp-go/tree/debug-pr-216

haoqixu:
```
2023/12/05 23:33:17 test begin
2023/12/05 23:33:17 client: SetAgentDescription and ScheduleSend
2023/12/05 23:33:17 old instance uid is 01HGX9SCX1EQPHGGJ9X6Y1JSTY, starting client
2023/12/05 23:33:17 client: send message at Start
2023/12/05 23:33:17 server: store uid 01HGX9SCX1EQPHGGJ9X6Y1JSTY to rcvAgentInstanceUid
2023/12/05 23:33:17 client: set instance uid of pending message 01HGX9SCX106AFVGQT5ZYC0GEK
2023/12/05 23:33:17 client: sending pending message
2023/12/05 23:33:17 server: store uid 01HGX9SCX106AFVGQT5ZYC0GEK to rcvAgentInstanceUid
2023/12/05 23:33:17 client: set instance uid of pending message 01HGX9SCX106AFVGQT5ZYC0GEK
2023/12/05 23:33:17 first assert: want 01HGX9SCX1EQPHGGJ9X6Y1JSTY, got: 01HGX9SCX106AFVGQT5ZYC0GEK
2023/12/05 23:33:17 first assert: want 01HGX9SCX1EQPHGGJ9X6Y1JSTY, got: 01HGX9SCX106AFVGQT5ZYC0GEK
2023/12/05 23:33:17 first assert: want 01HGX9SCX1EQPHGGJ9X6Y1JSTY, got: 01HGX9SCX106AFVGQT5ZYC0GEK
2023/12/05 23:33:17 first assert: want 01HGX9SCX1EQPHGGJ9X6Y1JSTY, got: 01HGX9SCX106AFVGQT5ZYC0GEK
2023/12/05 23:33:17 first assert: want 01HGX9SCX1EQPHGGJ9X6Y1JSTY, got: 01HGX9SCX106AFVGQT5ZYC0GEK
(... first assert message ...)
2023/12/05 23:33:22 first assert: want 01HGX9SCX1EQPHGGJ9X6Y1JSTY, got: 01HGX9SCX106AFVGQT5ZYC0GEK
2023/12/05 23:33:22 first assert: want 01HGX9SCX1EQPHGGJ9X6Y1JSTY, got: 01HGX9SCX106AFVGQT5ZYC0GEK
2023/12/05 23:33:22 first assert: want 01HGX9SCX1EQPHGGJ9X6Y1JSTY, got: 01HGX9SCX106AFVGQT5ZYC0GEK
2023/12/05 23:33:22 first assert: want 01HGX9SCX1EQPHGGJ9X6Y1JSTY, got: 01HGX9SCX106AFVGQT5ZYC0GEK
2023/12/05 23:33:22 first assert: want 01HGX9SCX1EQPHGGJ9X6Y1JSTY, got: 01HGX9SCX106AFVGQT5ZYC0GEK
2023/12/05 23:33:22 first assert: want 01HGX9SCX1EQPHGGJ9X6Y1JSTY, got: 01HGX9SCX106AFVGQT5ZYC0GEK
2023/12/05 23:33:22 first assert: want 01HGX9SCX1EQPHGGJ9X6Y1JSTY, got: 01HGX9SCX106AFVGQT5ZYC0GEK
2023/12/05 23:33:22 send dummy message
2023/12/05 23:33:22 client: SetAgentDescription and ScheduleSend
2023/12/05 23:33:22 client: sending pending message
2023/12/05 23:33:22 server: store uid 01HGX9SCX106AFVGQT5ZYC0GEK to rcvAgentInstanceUid
2023/12/05 23:33:22 client: set instance uid of pending message 01HGX9SCX106AFVGQT5ZYC0GEK
2023/12/05 23:33:22 second assert: want 01HGX9SCX106AFVGQT5ZYC0GEK, got: 01HGX9SCX106AFVGQT5ZYC0GEK
2023/12/05 23:33:22 test end
--- FAIL: TestAgentIdentification (5.06s)
    --- FAIL: TestAgentIdentification/ws (5.04s)
        clientimpl_test.go:94: 
            	Error Trace:	clientimpl_test.go:94
            	            				clientimpl_test.go:653
            	            				clientimpl_test.go:77
            	Error:      	Condition never satisfied
            	Test:       	TestAgentIdentification/ws
```
This is the Case 1.

srikanthccv:
Were you able to easily reproduce the original issue? Can you share how did you reproduce?

haoqixu:
I can reproduce it by running `go test -run TestAgentIdentification/ws -race -count 10000 ./...`

srikanthccv:
Thanks, I will take a look.

srikanthccv:
Sorry for the lack of response here for some time. I have been sick and swamped with a lot of work at $job. I will find time to get back to review PRs soon.

srikanthccv:
>The first assertion is used to check whether the server has received the original uid. In the timeline case 1, the server has received the original uid, the first assertion falied becase rcvAgentInstanceUid(atomic.Value) was overwritten by the second AgentToServer message.

Coming back to this point, this says the first assertion failed because id was overwritten by second `AgentToServer`. However, the second `AgentToServer` is the _dummy message_ which is sent after the `eventually` which is blocking so my reasoning and question is how does the second message cause the first assertion failure because they are sequential instructions. If we look at the code we expect the second assertion to override but not simultaneously with the first assertion.

haoqixu:
> the second `AgentToServer` is the _dummy message_

The second `AgentToServer` is not the _dummy message_. It is pending message scheduled by `SetAgentDescription`.

图片

srikanthccv:
Thanks, I went through the ws sender code again. I think the additional `sendNextMessage` invocation shouldn't be there in ws sender https://github.com/open-telemetry/opamp-go/blob/ca0e1e68754fdf9b131419e0629ed2d99bdae713/client/internal/wssender.go#L34-L42 All the client messages that should be sent to server would already have called the `ScheduleSend` and we expect the `SetAgentDescription` must be called before `Start` which schedules the message. I believe this is why we don't see the same flaky behaviour in the http client.

haoqixu:
> I believe this is why we don't see the same flaky behaviour in the http client.

Yes. But I don't think test code should treat the above cases as a failure. Even with this extra message, the client uses the original ID first and then sends subsequent messages with the new ID after receiving the reply as expected. I think we should update the test code so that the above two cases pass. I would like to submit anohter PR to avoid the additional message.

srikanthccv:
The test is written with certain assumptions and it failed because of the additional message, if we fix that test would pass and we don't need to fix any tests.

haoqixu:
> Even with this extra message, the client uses the original ID first and then sends subsequent messages with the new ID after receiving the reply as expected

I don't think the extra message is incorrect (although it could definitely be improved). The observable behavior of the client is as expected, but our detection method cannot handle it and fail to produce the same outcome with each individual test run.

haoqixu:
> if we fix that test would pass and we don't need to fix any tests.

Case 2 is not caused by the additional message and we still need to need the test.

haoqixu:
This test is flaky, it usually succeeds but occasionally fails. If we want to avoid extra messages, it is not enough to prove it, we still need a separate test. So I prefer that this flaky test code should be fixed first to make its results consistent.

tigrannajaryan:
It has been a while and the test has been changed since it failed and does not appear to be failing any more, [see here](https://github.com/open-telemetry/opamp-go/issues/191#issuecomment-2192697229).

Closing this PR, but we can reopen if we see failures again.

PR #217, opened 2023-12-14
feat: allow to get the network address server is listening on

haoqixu:
Add an `Addr()` method to fetch the network address Server is listening on. I think this may be useful for the supervisor to start a server with 0 port to allocate an ephemeral port and get the allocated port for rendering an agent config.

### DIFF
--- a/server/serverimpl_test.go
+++ b/server/serverimpl_test.go
@@ -28,8 +28,8 @@ func startServer(t *testing.T, settings *StartSettings) *server {
 	srv := New(&sharedinternal.NopLogger{})
 	require.NotNil(t, srv)
 	if settings.ListenEndpoint == "" {
-		// allocate an ephemeral port to listne on.
-		settings.ListenEndpoint = "127.0.0.1:0"
+		// Find an avaiable port to listne on.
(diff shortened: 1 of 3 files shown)


server/serverimpl_test.go
```diff
@@ -28,8 +28,8 @@ func startServer(t *testing.T, settings *StartSettings) *server {
 	srv := New(&sharedinternal.NopLogger{})
 	require.NotNil(t, srv)
 	if settings.ListenEndpoint == "" {
-		// Find an avaiable port to listne on.
-		settings.ListenEndpoint = testhelpers.GetAvailableLocalAddress()
+		// allocate an ephemeral port to listne on.
+		settings.ListenEndpoint = "127.0.0.1:0"
```
tigrannajaryan, line 32: Instead of changing existing tests can we keep existing tests as is and add another simple test that demonstrates the use of 0 port and Addr() method?

tigrannajaryan (commented):
I think this is a good idea.

tigrannajaryan (approved):

tigrannajaryan:
Thank you @haoqixu !

PR #218, opened 2023-12-20
Fix set invalid instance UID error throw

sairam866:
Fixing error throw logger when setting invalid instance UID provided by the server.

### DIFF
--- a/client/internal/receivedprocessor.go
+++ b/client/internal/receivedprocessor.go
@@ -217,7 +217,7 @@ func (r *receivedProcessor) rcvAgentIdentification(agentId *protobufs.AgentIdent
 
 	err := r.sender.SetInstanceUid(agentId.NewInstanceUid)
 	if err != nil {
-		r.logger.Errorf("error while setting instance uid: %v", err)
+		r.logger.Errorf("Error while setting instance uid: %v", err)
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


tigrannajaryan:
Thank you for the fix @sairam866 
Please sign the CLA so that we can accept the PR.

sairam866:
@tigrannajaryan Updated my commit with my company email. CLA should be fine now.

client/internal/receivedprocessor.go
```diff
@@ -217,7 +217,7 @@ func (r *receivedProcessor) rcvAgentIdentification(agentId *protobufs.AgentIdent
 
 	err := r.sender.SetInstanceUid(agentId.NewInstanceUid)
 	if err != nil {
-		r.logger.Errorf("Error while setting instance uid: %v, err")
+		r.logger.Errorf("error while setting instance uid: %v", err)
```
tigrannajaryan, line 220: We capitalize log messages:
```suggestion
		r.logger.Errorf("Error while setting instance uid: %v", err)
```

sairam866: Noted. Updated the commit.

tigrannajaryan (approved):

tigrannajaryan:
Thanks @sairam866

PR #224, opened 2024-01-09
Make renovatebot to run go mod tidy

tigrannajaryan:
See https://docs.renovatebot.com/golang/#module-tidying


tigrannajaryan:
This is probably not necessary, we can do it manually, should be rare.

tigrannajaryan:
Looks like this is really needed, 2 last updates failed because of it. Let's try enabling.

srikanthccv (approved):

PR #236, opened 2024-01-10
chore: catch error when demo server start

Frapschen:
It's better to catch any error during the demo OpAMP Server starting. e.g. the 4320 port already in use.

before this PR: if we start an OpAMP Server in the background, and then start another OpAMP Server, we will get:
```
fraps@CMQ server % ./bin/server 
2024/01/10 10:46:33.961345 [MAIN] OpAMP Server starting...
2024/01/10 10:46:33.963579 [MAIN] OpAMP Server running...
```
however, it fails to start the OpAMP Server.

with this PR's enhance:
```
fraps@CMQ server % ./bin/server
2024/01/10 10:49:12.296051 [MAIN] OpAMP Server starting...
2024/01/10 10:49:12.298014 [OPAMP] OpAMP server start fail: listen tcp 127.0.0.1:4320: bind: address already in use
```

It response the real status of the OpAMP server.

### DIFF
--- a/client/types/logger.go
+++ b/client/types/logger.go
@@ -4,5 +4,4 @@ package types
 type Logger interface {
 	Debugf(format string, v ...interface{})
 	Errorf(format string, v ...interface{})
-	Fatalf(format string, v ...interface{})
 }
--- a/client/types/logger.go
+++ b/client/types/logger.go
@@ -4,4 +4,5 @@ package types
 type Logger interface {
 	Debugf(format string, v ...interface{})
 	Errorf(format string, v ...interface{})
+	Fatalf(format string, v ...interface{})
 }
--- a/internal/examples/server/opampsrv/opampsrv.go
+++ b/internal/examples/server/opampsrv/opampsrv.go
@@ -64,7 +64,9 @@ func (srv *Server) Start() {
 	}
 	settings.TLSConfig = tlsConfig
 
-	srv.opampSrv.Start(settings)
+	if err := srv.opampSrv.Start(settings); err != nil {
+		srv.logger.Fatalf("OpAMP server start fail: %v", err.Error())
+	}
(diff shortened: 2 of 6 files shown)


srikanthccv (approved):
Thank you

client/types/logger.go
```diff
@@ -4,4 +4,5 @@ package types
 type Logger interface {
 	Debugf(format string, v ...interface{})
 	Errorf(format string, v ...interface{})
+	Fatalf(format string, v ...interface{})
```
tigrannajaryan, line 7: I would not want to change the Logger interface which is part of public API. This is a breaking change and also increases the public API size.

Instead I suggest where necessary the example code calls the regular Logger.Errorf() followed by a whatever its preferred termination function is (e.g. os.Exit).

internal/examples/server/opampsrv/opampsrv.go
```diff
@@ -64,7 +64,9 @@ func (srv *Server) Start() {
 	}
 	settings.TLSConfig = tlsConfig
 
-	srv.opampSrv.Start(settings)
+	if err := srv.opampSrv.Start(settings); err != nil {
+		srv.logger.Fatalf("OpAMP server start fail: %v", err.Error())
```
tigrannajaryan, line 68:
```suggestion
		srv.logger.Errorf("OpAMP server start fail: %v", err.Error())
		os.Exit(1)
```

tigrannajaryan (changes requested):

client/types/logger.go
Frapschen: ok

tigrannajaryan (approved):

tigrannajaryan:
Thank you @Frapschen

Issue #214, opened 2023-11-21
Allow callbacks to contain a context


jaronoff97:
Right now context is handled in a way such that each method needs to create its own [background context](https://github.com/open-telemetry/opamp-go/blob/f0e220d59fa17849a128403d6383f8c0607a0a25/server/serverimpl.go#L234). Allowing each handler to receive a context would help with things like cancelling handling a message if a connection is severed as well as have a full context propagated for logging purposes.

PR #237, opened 2024-01-10
Add context and info to logging interface

jaronoff97:
Allow the logging interface to propagate context where possible. This will let logging sinks contain richer information about potential opamp IO errors.


jaronoff97:
Before i continue with this... @andykellr @evan-bradley @tigrannajaryan what do you all think of this approach? Propagating the context would make it easier for consumers to understand the potential problems that the impls run in to, but i'm not sure how i feel about the interface. Let me know your thoughts. Thank you.

tigrannajaryan:
I have a hard time reviewing the code due to all the junk codecov added. I am going to disable it for now.

tigrannajaryan:
The idea to pass the context to logging functions sounds good to me. I am not sure we need a separate set of functions for that. Perhaps just add context parameter to existing functions?

jaronoff97:
@tigrannajaryan sounds good! I'll update to do that.

mwear:
One benefit to the current approach is that the interface matches the new Go structured logger, [slog](https://pkg.go.dev/log/slog#Logger). Rather than adding context to the existing functions, it might make more sense to keep the context versions, and remove (or deprecate) the non-context versions. That way you can use any logger that is `slog` compatible.

jaronoff97:
yeah that was the intended purpose... I see benefits for both approaches... Maybe we should discuss at the next SIG call?

tigrannajaryan:
The `Logger` interface here is not a general-purpose logging API. It is on purpose minimalistic and designed to support only what is needed by opamp implementation and nothing else. If we want to pass the context I am happy to make it a required parameter, I don't see the need to give a choice (side note: I argued that slog also makes it a required parameter, but wasn't able to convince the author).

I prefer to keep Logger interface as small as possible and bridge to any other logging API as needed. Coding a bridge from `Logger` to `slog` is so easy that I don't see the need to make the interface identical.

We also do not have stability promises in this repo yet, so I am not too worried about making a breaking change that is very easy to adopt. I don't think we need a deprecations process for this.

jaronoff97:
done! I will also work on the follow up described [here](#214) which will make these even more valuable.

client/types/logger.go
```diff
@@ -1,7 +1,10 @@
 package types
 
+import "context"
+
 // Logger is the logging interface used by the OpAMP Client.
 type Logger interface {
-	Debugf(format string, v ...interface{})
-	Errorf(format string, v ...interface{})
+	Infof(ctx context.Context, format string, v ...interface{})
```
tigrannajaryan, line 7: This does not seem to be used anywhere. Can we remove?

tigrannajaryan (commented):
Mostly LGTM, except I think we don't need the Infof() for now. Can be added later when needed.

jaronoff97:
the codecov is failing for probably obvious reasons – testing error logs isn't really done right now

tigrannajaryan (approved):

tigrannajaryan:
This is a breaking change in the API, allowed because the API is not declared stable yet.

Issue #239, opened 2024-01-13
`ReceiverLoop` doesn't stop on cancelling context


srikanthccv:
https://github.com/open-telemetry/opamp-go/blob/62d8b6039be3617927ad94df836f4359393e311a/client/internal/wsreceiver.go#L44-L62

The receiver loop here doesn't stop when the context is cancelled. The break out is reached only when there is an error from `receiveMessage`.

PR #240, opened 2024-01-13
ReceiverLoop should stop on cancelling context

srikanthccv:
Here is test run that shows it doesn't stop without fix https://github.com/open-telemetry/opamp-go/actions/runs/7512286887/job/20452817400

Fixes https://github.com/open-telemetry/opamp-go/issues/239


client/internal/wsreceiver_test.go
```diff
@@ -175,3 +178,36 @@ func TestDecodeMessage(t *testing.T) {
...
+	sender := WSSender{}
+	capabilities := protobufs.AgentCapabilities_AgentCapabilities_AcceptsRestartCommand
+	receiver := NewWSReceiver(TestLogger{t}, callbacks, conn, &sender, &clientSyncedState, nil, capabilities)
+	ctx, cancel := context.WithCancel(context.Background())
+
+	go func() {
+		receiver.ReceiverLoop(ctx)
+		receiverLoopStopped.Store(true)
+	}()
+	cancel()
```
tigrannajaryan, line 208: By design WSReceiver currently stops when the connection is closed and reading from the socket returns an error. There is no expectation that it will stop when the context is cancelled. Why do you need it to stop on context cancellation?

srikanthccv: The go doc for function says _// ReceiverLoop runs the receiver loop. To stop the receiver cancel the context_ and the check `ctx.Err()` I thought this was designed to respond to context cancellations. Until today (on the main and latest release) the `Stop` call is blocking even after the context is cancelled as I observed in supervisor tests. This behaviour would change when https://github.com/open-telemetry/opamp-go/pull/213 gets merged. If it's by design we should update the go doc and I am not sure if it still makes sense to propagate context to `ReceiverLoop`.

tigrannajaryan: My statement was not correct. Cancellation of the context is supposed to interrupt any ongoing callbacks. ProcessReceivedMessage() passes a cancellable context to the callbacks.

So you are right and the comment in the code I believe is also correct.

client/internal/wsreceiver.go
```diff
@@ -43,22 +43,39 @@ func NewWSReceiver(
...
 	for {
-		var message protobufs.ServerToAgent
-		if err := r.receiveMessage(&message); err != nil {
-			if ctx.Err() == nil && !websocket.IsCloseError(err, websocket.CloseNormalClosure) {
-				r.logger.Errorf("Unexpected error while receiving: %v", err)
+		select {
+		case <-ctx.Done():
+			return
+		default:
+			result := make(chan struct {
```
tigrannajaryan, line 51: Nit: declare the struct as a type to avoid duplicating the declaration again on line 59.

client/internal/wsreceiver.go
```diff
@@ -43,22 +43,39 @@ func NewWSReceiver(
...
+				message *protobufs.ServerToAgent
+				err     error
+			}, 1)
+
+			go func() {
```
tigrannajaryan, line 56: Can we avoid creating a goroutine for every message received?

Probably not a big deal for the client, we don't receive that many messages for this to have a major performance impact.

If there is an easy way to avoid it it would be nice, otherwise probably don't bother, this is ok.

srikanthccv: I couldn't think of a clean way to do it; Happy to update if you have suggestions.

tigrannajaryan (approved):

Issue #245, opened 2024-01-19
ServerOfferedVersion and ServerOfferedHash are never changed


phanidevavarapu:
As per the documentation 

`	// The version of the package that the Server offered to the Agent.
	// MUST be set if the installation of the package is initiated by an earlier offer
	// from the Server to install this package.
	//
	// MUST be empty if the Agent has this package but it was installed locally and
	// was not offered by the Server.
	//
	// Note that it is possible for both agent_has_version and server_offered_version
	// fields to be set and to have different values. This is for example possible if
	// the Agent already has a version of the package successfully installed, the Server
	// offers a different version, but the Agent fails to install that version.`

but if we look at the [code](https://github.com/open-telemetry/opamp-go/blob/main/client/internal/packagessyncer.go#L150), the ServerOfferedVersion and ServerOfferedHash is only initialized when the status is null and never changes after. 

Let's take case when the Client is started for the first time and the PackageStatus is nil, now server has offered a package a1 version v1 and is successful. In This case, agenthasversion and serverofferedversion will the same as v1. IN the next check-in server has provided a v2 version for the package a1, upon successful only the agenthasversion fied is upgraded to v2 but the serverofferedversion  still remains at v1(at least this is what I am observing). 

Please do look into and clarify be if my understanding is wrong.

tigrannajaryan:
Fixed in https://github.com/open-telemetry/opamp-go/pull/246

PR #246, opened 2024-01-19
Changes to set the ServerOfferedVersion and ServerOfferedHash always

phanidevavarapu:
This PR will fix the bug https://github.com/open-telemetry/opamp-go/issues/245, the ServerOfferedVersion and ServerOfferedHash are only set when the package is sent from the server the very first time. 

The changes in this PR will make sure that the info is set all the time.


srikanthccv:
Please add a test case. Let try to include cases one where the installation of new version fails and another that succeeds.

phanidevavarapu:
> Please add a test case. Let's try to include cases one where the installation of the new version fails and another that succeeds.

@srikanthccv the unit test you suggested will cut across all many functions with in this file itself, it will be huge undertaking compared to the code change that I made.

srikanthccv:
Regardless of the size of the change, we should add tests. That's how we prove the change actually fixes it and having a test ensures the issue will not appear again in future.

tigrannajaryan:
> Regardless of the size of the change, we should add tests. That's how we prove the change actually fixes it and having a test ensures the issue will not appear again in future.

+1.  For all bug fixes we want tests that demonstrate that there is a bug (tests fail before the fix) and that the PR fixes the bug (tests pass after the fix).

I understand this adds more work, but this is important for maintaining the quality of the code and being confident about bug fixes.

srikanthccv:
@phanidevavarapu I had opened a PR to add tests against your fork here https://github.com/phanidevavarapu/opamp-go/pull/1. Did you get a chance to look at it?

srikanthccv:
Can you fix the conflicts and update the packagessyncer.go because I commented out the lines to show tests fail.

phanidevavarapu:
@srikanthccv updated the PR, There is a test failing, but I don't see any failures in local.

tigrannajaryan:
TestOfferUpdatedVersion/ws test got stuck, likely an indication of a real problem.

phanidevavarapu:
@tigrannajaryan I am not sure how to approach this now, tests seems to failing and passing at random. Any help to have everything green is really appreciated.

srikanthccv:
While the individual test case `go test -timeout 30s -run ^TestOfferUpdatedVersion$ github.com/open-telemetry/opamp-go/client` passes successfully, running the entire test suite fails with a timeout failure consistently. We need to check why the collective test execution fails and fix it.

phanidevavarapu:
Can we approve and merge this ticket?

srikanthccv:
We need to fix the failing test first. Were you able to find out why the test times out?

srikanthccv:
The changes look good to me. I tried running the workflow multiple times to see if shows flaky behaviour but it passed all runs. I still suspect the test is flaky since it passes now without any changes to the test case.

tigrannajaryan (approved):

tigrannajaryan:
@phanidevavarapu this looks good to me, thank you. Can you please resolve the conflicts so that we can merge?

phanidevavarapu:
@tigrannajaryan the merge conflicts are resolved, please go ahead and merge.

PR #247, opened 2024-01-22
Change callback signatures to pass down context where applicable

### LINKED ISSUE #214: Allow callbacks to contain a context

jaronoff97:
Changes a few server callback interface methods to pass down a context to propagate request information effectively. This closes #214

### DIFF
--- a/client/internal/receivedprocessor.go
+++ b/client/internal/receivedprocessor.go
@@ -129,7 +129,7 @@ func (r *receivedProcessor) ProcessReceivedMessage(ctx context.Context, msg *pro
 		}
 
 		if msg.AgentIdentification != nil {
-			err := r.rcvAgentIdentification(msg.AgentIdentification)
+			err := r.rcvAgentIdentification(ctx, msg.AgentIdentification)
 			if err == nil {
 				msgData.AgentIdentification = msg.AgentIdentification
 			}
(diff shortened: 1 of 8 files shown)


client/clientimpl_test.go
```diff
@@ -627,12 +627,24 @@ func TestAgentIdentification(t *testing.T) {
 			ulid.Timestamp(time.Now()), ulid.Monotonic(rand.New(rand.NewSource(0)), 0),
 		)
 		var rcvAgentInstanceUid atomic.Value
+		var sentInvalidId atomic.Bool
 		srv.OnMessage = func(msg *protobufs.AgentToServer) *protobufs.ServerToAgent {
 			rcvAgentInstanceUid.Store(msg.InstanceUid)
+			if sentInvalidId.Load() {
```
jaronoff97, line 633: codecov was complaining that this was untested, so i added something that forced an error for an empty string and then resolves it. This will trigger the error checking logic.

gdfast (approved):

andykellr (approved):

PR #248, opened 2024-01-26
Add client context propagation

jaronoff97:
This is a follow up to #237 and #247, adding context propagation for client methods. 

**This involves a breaking change for the client interfaces**

### DIFF
--- a/client/internal/receivedprocessor.go
+++ b/client/internal/receivedprocessor.go
@@ -56,7 +56,7 @@ func (r *receivedProcessor) ProcessReceivedMessage(ctx context.Context, msg *pro
 		// to process.
 		if msg.Command != nil {
 			if r.hasCapability(protobufs.AgentCapabilities_AgentCapabilities_AcceptsRestartCommand) {
-				r.rcvCommand(msg.Command)
+				r.rcvCommand(ctx, msg.Command)
 				// If a command message exists, other messages will be ignored
 				return
 			} else {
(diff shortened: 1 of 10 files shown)


client/internal/receivedprocessor.go
```diff
@@ -198,16 +198,17 @@ func (r *receivedProcessor) rcvOpampConnectionSettings(ctx context.Context, sett
...
 		}
 	} else {
 		r.logger.Debugf(ctx, "Ignoring Opamp, agent does not have AcceptsOpAMPConnectionSettings capability")
 	}
 }
 
 func (r *receivedProcessor) processErrorResponse(ctx context.Context, body *protobufs.ServerErrorResponse) {
-	// TODO: implement this.
-	r.logger.Errorf(ctx, "received an error from server: %s", body.ErrorMessage)
+	if body != nil {
```
jaronoff97, line 209: This never called the OnError method when it should have.

client/types/callbacks.go
```diff
@@ -45,18 +45,18 @@ type Callbacks interface {
 	// May be called after Start() is called and every time a connection is established to the Server.
 	// For WebSocket clients this is called after the handshake is completed without any error.
 	// For HTTP clients this is called for any request if the response status is OK.
-	OnConnect()
+	OnConnect(ctx context.Context)
```
tigrannajaryan, line 48: Can we add a comment (one for the entire interface) that explains how the callback methods are supposed to treat the context? I think we expect them to honour the cancellation.

jaronoff97: yep. Can do!

tigrannajaryan (approved):

gdfast (commented):
Great change!

PR #251, opened 2024-02-12
Follow HTTP redirects after failed WS dials

### LINKED ISSUE #250: Follow HTTP redirects on websocket handshake failure
It would be useful to check for a valid redirect on handshake failure in the `opamp-go` client, and follow it to attempt to establish a websocket connection to the location specified by the redirect. This would allow more flexibility for operators who are supporting a large network of opamp servers. It is also a backwards-compatible change.

The websocket standard permits websocket servers to offer redirection to websocket clients. (See https://www.rfc-editor.org/rfc/rfc6455#section-4.2.2)

In particular,

> 3.  The server MAY redirect the client using a 3xx status code
       [[RFC2616](https://www.rfc-editor.org/rfc/rfc2616)].  Note that this step can happen together with, before,
       or after the optional authentication step described above.

The `gorilla/websocket` library does not implement this optional redirect following, but it does return a valid HTTP response when a handshake error occurs. (See https://github.com/gorilla/websocket/blob/main/client.go#L412)

What I am proposing specifically:

In https://github.com/open-telemetry/opamp-go, check for an HTTP response status `>= 300 && <400`. Parse the location header, and if valid, set `c.url` to be the URL from the location header. Then, call `tryConnectOnce` again.

I am happy to do the implementation work this would require, and have a changelog I can contribute in the near future.

echlebek:
This commit allows the opamp client to follow redirects after websocket handshake failures. Redirect following is not implemented by gorilla/websocket, but can be handled by inspecting the returned response object for 3xx status and Location header.

Closes #250

### DIFF
--- a/client/wsclient.go
+++ b/client/wsclient.go
@@ -141,7 +141,8 @@ func (c *wsClient) tryConnectOnce(ctx context.Context) (err error, retryAfter sh
 					c.common.Logger.Errorf(ctx, "%d redirect, but no valid location: %s", resp.StatusCode, err)
 					return err, duration
 				}
-				if redirect.Scheme == "http" || redirect.Scheme == "" {
+				// rewrite the scheme for the sake of tolerance
+				if redirect.Scheme == "http" {
 					redirect.Scheme = "ws"
 				} else if redirect.Scheme == "https" {
 					redirect.Scheme = "wss"
--- a/client/wsclient.go
+++ b/client/wsclient.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/url"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
@@ -35,6 +36,10 @@ type wsClient struct {
 
 	// The sender is responsible for sending portion of the OpAMP protocol.
 	sender *internal.WSSender
+
+	// last non-nil internal error that was encountered in the conn retry loop,
+	// currently used only for testing.
+	lastInternalErr atomic.Pointer[error]
 }
 
 // NewWebSocket creates a new OpAMP Client that uses WebSocket transport.
@@ -136,7 +141,6 @@ func (c *wsClient) tryConnectOnce(ctx context.Context) (err error, retryAfter sh
 				// very liberal handling of 3xx that largely ignores HTTP semantics
 				redirect, err := resp.Location()
 				if err != nil {
-					c.common.Callbacks.OnConnectFailed(ctx, err)
 					c.common.Logger.Errorf(ctx, "%d redirect, but no valid location: %s", resp.StatusCode, err)
 					return err, duration
 				}
@@ -187,6 +191,7 @@ func (c *wsClient) ensureConnected(ctx context.Context) error {
 		case <-timer.C:
 			{
 				if err, retryAfter := c.tryConnectOnce(ctx); err != nil {
+					c.lastInternalErr.Store(&err)
 					if errors.Is(err, context.Canceled) {
 						c.common.Logger.Debugf(ctx, "Client is stopped, will not try anymore.")
 						return err
--- a/client/wsclient.go
+++ b/client/wsclient.go
@@ -3,6 +3,7 @@ package client
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net/http"
 	"net/url"
 	"sync"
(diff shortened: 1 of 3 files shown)


tigrannajaryan:
Thank you for the PR @echlebek 
Please sign the CLA.

echlebek:
@tigrannajaryan Thanks for taking a look, I've signed the CLA.

client/wsclient.go
```diff
@@ -57,7 +57,7 @@ func (c *wsClient) Start(ctx context.Context, settings types.StartSettings) erro
 	}
 
 	// Prepare connection settings.
-	c.dialer = *websocket.DefaultDialer
+	c.dialer = websocket.DefaultDialer
```
tigrannajaryan, line 60: We modify `c.dialer` fields later in the code. That would modify `websocket.DefaultDialer` which is not what we want.

client/wsclient_test.go
```diff
@@ -177,3 +180,50 @@ func TestVerifyWSCompress(t *testing.T) {
...
 	}
 }
+
+func redirectServer(to string) *httptest.Server {
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		http.Redirect(w, req, to, http.StatusSeeOther)
+	}))
+}
+
+func TestRedirectWS(t *testing.T) {
```
tigrannajaryan, line 190: Can you add another test case that return 3xx response but does not include "Location" to test for the error path?

tigrannajaryan: See coverage report: https://app.codecov.io/github/open-telemetry/opamp-go/commit/353be65d240a6e5129d9d2cb4fd10f1434489aca/blob/client/wsclient.go#L135

It would be nice to cover the paths that are not covered.

client/wsclient.go
tigrannajaryan: We don't modify slices. We modify a bool and a pointer field, which is safe. See the code later.

If there is an easy way to make deep copy that would be better and more future-proof, I agree, but for now the code is safe as long as `c.dialer` does not point to `websocket.DefaultDialer`.

echlebek: OK, it's fair to say this is outside the scope of the changeset, and I'll revert this :+1:

client/wsclient.go
```diff
@@ -131,8 +131,26 @@ func (c *wsClient) tryConnectOnce(ctx context.Context) (err error, retryAfter sh
...
+				}
+				if redirect.Scheme == "http" || redirect.Scheme == "" {
+					redirect.Scheme = "ws"
+				} else if redirect.Scheme == "https" {
+					redirect.Scheme = "wss"
+				}
+				c.common.Logger.Debugf(ctx, "%d redirect to %s", resp.StatusCode, redirect)
+				// Set the URL to the redirect, so that it connects to it on the
+				// next cycle.
+				c.url = redirect
```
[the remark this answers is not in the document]
tigrannajaryan: On the second thought I am probably wrong. If the server is misconfigured and returns an infinite loop of redirects we don't want to bombard it with requests, better keep the intervals.

echlebek:
Coverage is now improved, with only one line missing coverage from the method. It would require a TLS server in test to cover it. Let me know if you think it's worthwhile. Thanks!

client/wsclient.go
```diff
@@ -131,8 +131,28 @@ func (c *wsClient) tryConnectOnce(ctx context.Context) (err error, retryAfter sh
 			c.common.Callbacks.OnConnectFailed(ctx, err)
 		}
 		if resp != nil {
-			c.common.Logger.Errorf(ctx, "Server responded with status=%v", resp.Status)
 			duration := sharedinternal.ExtractRetryAfterHeader(resp)
+			if resp.StatusCode >= 300 && resp.StatusCode < 400 {
+				// very liberal handling of 3xx that largely ignores HTTP semantics
+				redirect, err := resp.Location()
+				if err != nil {
+					c.common.Callbacks.OnConnectFailed(ctx, err)
```
tigrannajaryan, line 139: I am not sure we need this second call to OnConnectFailed. The callback was already called on line 131. This would be the second call per connection attempt and I don't think that's a good thing to do.
Unclear what extra value does this call give to the user. They could log the error, but we already do that here.

tigrannajaryan (commented):
Mostly LGTM, coverage looks good. 
Just one comment remaining.

client/wsclient.go
echlebek: That's a fair callout. In this case it was just propping up the test, as I didn't have another way of easily inspecting the error condition here: https://github.com/open-telemetry/opamp-go/pull/251/files#diff-9adf0e52d7c19860aa5771807f97b10d8ab7d5acc71acc2fa1c7f322bc4e46c6R234

I don't think the tests should push the implementation around, but I also am not sure how better to deal with this in test. Any suggestions?

tigrannajaryan: One possible way is to have an unexported field in wsClient that we set here (e.g. `lastRespErr` or similar) and use it in the test.

tigrannajaryan:
Thank you @echlebek !

PR #254, opened 2024-02-23
Add staticcheck github action, resolve existing lint

echlebek:
This commit adds staticcheck as a github action and resolves existing issues flagged by staticcheck. I've grouped the issues into individual commits so that they can be considered independently. Most of the issues are deprecated uses of stdlib, unused errors, and ineffective branches or statements. But, there is one stylistic issue too that is more subjective.

staticcheck is a very conservative linter that has a focus on minimizing false positives. I've found that staticcheck has improved the quality of my own work, and has also made pull requests easier to evaluate.

Each issue is tagged with an identifier that can be consulted here: https://staticcheck.dev/docs/checks/

### DIFF
--- a/client/httpclient_test.go
+++ b/client/httpclient_test.go
@@ -114,10 +114,12 @@ func TestHTTPClientSetPollingInterval(t *testing.T) {
 	srv := internal.StartMockServer(t)
 	var rcvCounter int64
 	srv.OnMessage = func(msg *protobufs.AgentToServer) *protobufs.ServerToAgent {
-		assert.EqualValues(t, rcvCounter, msg.SequenceNum)
-		if msg != nil {
-			atomic.AddInt64(&rcvCounter, 1)
+		if msg == nil {
+			t.Error("unexpected nil msg")
+			return nil
 		}
+		assert.EqualValues(t, rcvCounter, msg.SequenceNum)
+		atomic.AddInt64(&rcvCounter, 1)
 		return nil
 	}
 
--- a/client/httpclient_test.go
+++ b/client/httpclient_test.go
@@ -22,10 +22,12 @@ func TestHTTPPolling(t *testing.T) {
 	srv := internal.StartMockServer(t)
 	var rcvCounter int64
 	srv.OnMessage = func(msg *protobufs.AgentToServer) *protobufs.ServerToAgent {
-		assert.EqualValues(t, rcvCounter, msg.SequenceNum)
-		if msg != nil {
-			atomic.AddInt64(&rcvCounter, 1)
+		if msg == nil {
+			t.Error("unexpected nil msg")
+			return nil
 		}
+		assert.EqualValues(t, rcvCounter, msg.SequenceNum)
+		atomic.AddInt64(&rcvCounter, 1)
 		return nil
 	}
 
--- a/client/httpclient_test.go
+++ b/client/httpclient_test.go
@@ -3,7 +3,7 @@ package client
 import (
 	"compress/gzip"
 	"context"
-	"io/ioutil"
+	"io"
 	"net/http"
 	"sync/atomic"
 	"testing"
@@ -63,7 +63,7 @@ func TestHTTPClientCompression(t *testing.T) {
 		assert.Equal(t, "gzip", r.Header.Get("Content-Encoding"))
 		reader, err := gzip.NewReader(r.Body)
 		assert.NoError(t, err)
-		body, err := ioutil.ReadAll(reader)
+		body, err := io.ReadAll(reader)
 		assert.NoError(t, err)
 		_ = r.Body.Close()
 		var response protobufs.AgentToServer
--- a/client/internal/mockserver.go
+++ b/client/internal/mockserver.go
@@ -40,8 +40,8 @@ const contentTypeProtobuf = "application/x-protobuf"
 func newMockServer(t *testing.T) (*MockServer, *http.ServeMux) {
 	srv := &MockServer{
 		t:                t,
-		expectedHandlers: make(chan receivedMessageHandler, 0),
-		expectedComplete: make(chan struct{}, 0),
+		expectedHandlers: make(chan receivedMessageHandler),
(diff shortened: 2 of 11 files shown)


client/httpclient_test.go
```diff
@@ -22,10 +22,12 @@ func TestHTTPPolling(t *testing.T) {
 	srv := internal.StartMockServer(t)
 	var rcvCounter int64
 	srv.OnMessage = func(msg *protobufs.AgentToServer) *protobufs.ServerToAgent {
-		assert.EqualValues(t, rcvCounter, msg.SequenceNum)
-		if msg != nil {
-			atomic.AddInt64(&rcvCounter, 1)
+		if msg == nil {
+			t.Error("unexpected nil msg")
+			return nil
```
tigrannajaryan, line 27: Good catch!

client/internal/mockserver.go
```diff
@@ -112,6 +112,11 @@ func (m *MockServer) EnableExpectMode() {
 
 func (m *MockServer) handlePlainHttp(w http.ResponseWriter, r *http.Request) {
 	msgBytes, err := io.ReadAll(r.Body)
+	if err != nil {
+		// request could not be read for some reason
+		http.Error(w, err.Error(), 500)
```
tigrannajaryan, line 117: We don't wan't to accidentally leak sensitive info that may be in the error message, so best not to use it directly in the response.
```suggestion
		http.Error(w, "Error reading request body", http.StatusInternalServerError)
```

internal/retryafter_test.go
```diff
@@ -32,7 +33,6 @@ func assertDuration(t *testing.T, duration OptionalDuration, expected time.Durat
 
 func TestExtractRetryAfterHeaderDelaySeconds(t *testing.T) {
 	// Generate random n > 0 int
-	rand.Seed(time.Now().UnixNano())
```
tigrannajaryan, line 35: Why the deletion?

internal/wsmessage.go
```diff
@@ -23,9 +23,9 @@ func DecodeWSMessage(bytes []byte, msg proto.Message) error {
 		}
 		// Skip the header. It really is just a single zero byte for now.
 		bytes = bytes[n:]
-	} else {
+	} /*else {
 		// Old message format. No header present.
-	}
+	}*/
```
tigrannajaryan, line 28: Ahh, this looks a bit ugly to my eyes. Perhaps just do this:
```go
}
// If no header was present (the "if" check above), then this is the old message format. No header is present, so we don't do anything about the header.
```
Maybe you can phrase it a bit better. 
I just don't like the commented "else", it looks as if someone left half-done code that needs cleaning.

tigrannajaryan (commented):
Thanks for adding this.

client/internal/mockserver.go
echlebek: I think in this case it is OK, since it's a mock server we only expect to use in test?

internal/retryafter_test.go
echlebek: Seed is deprecated: https://pkg.go.dev/math/rand#Seed

client/internal/mockserver.go
tigrannajaryan: Yes, you are right. I missed that it is a mock server.

tigrannajaryan: Can still use `http.StatusInternalServerError`, no harm in that.

internal/retryafter_test.go
tigrannajaryan: I see. The new behavior is:
>If Seed is not called, the generator is seeded randomly at program startup.

That's all we need, so good to delete it.

internal/certs.go
```diff
@@ -79,19 +78,18 @@ func CreateServerTLSConfig(caCertPath, serverCertPath, serverKeyPath string) (*t
 		InsecureSkipVerify: true,
 		ClientCAs:          caCertPool,
 	}
-	tlsConfig.BuildNameToCertificate()
```
tigrannajaryan, line 82: Reading the deprecation notice this _seems_ to be fine, but I am not entirely sure since the notice is a bit vague.
Are you able to verify that the example server that uses this functionality still works fine? The clients should be able to connect if everything is working correctly.

echlebek: Sure, I can add a test with a TLS server.

echlebek:
As an aside, I tried to fix the data race in the example, but architecturally it's actually a bit difficult to do. The Wait() on the command is causing a race when other bits of the system write, and it won't be a simple fix unfortunately. Just thought I'd mention my findings.

internal/certs.go
echlebek: Thoughts on using https://github.com/open-telemetry/opamp-go/pull/202?

tigrannajaryan: Yeah #202 will need to be updated, it has conflicts. I don't think you need to add a test as part of this particular PR. If you can confirm by running the examples manually that they still work we should be good and can merge this PR.

tigrannajaryan:
> As an aside, I tried to fix the data race in the example, but architecturally it's actually a bit difficult to do. The Wait() on the command is causing a race when other bits of the system write, and it won't be a simple fix unfortunately. Just thought I'd mention my findings.

Can you please create an issue and describe your findings so that we can fix it later?

internal/certs.go
tigrannajaryan: `make run-examples` should build and run the example server and agent and if you go to http://localhost:4321/ you should be able to see a connected agent. 

image


Click on it, scroll down and ensure you see it connected using a client certificate, e.g.:

image

echlebek: Verified that the client certificate is shown on this page. Thanks for the instructions!

image

echlebek:
Filed https://github.com/open-telemetry/opamp-go/issues/256

tigrannajaryan (approved):

tigrannajaryan:
Thank you for the PR @echlebek !

Issue #253, opened 2024-02-22
opamp-go server impl is breaking traces


gdfast:
### Scenario

We have an Envoy gateway in front of an OpAMP server built on top of opamp-go, specifically using [Start](https://github.com/open-telemetry/opamp-go/blob/4d07a6af062f025852e9060379710b4da2220a8e/server/serverimpl.go#L67). The gateway is instrumented to emit OTLP spans. However, for 

#### Problem

Broken traces making for more challenging debugging and monitoring of the OpAMP server

### Goal

_Minimum_: opamp-go's server implementation persists traceID in its context
_ideal_: opamp-go's server impl, in addition to persisting traceID, also instruments spans

### Some solution ideas

I think given that opamp-go is an OpenTelemetry project, it makes sense to Do The Right Thing and update serviceimpl.go – particular the `Handler` –  to extract baggage and context from the request span and instrument a new server span.

A more flexible approach would be to organize the code to allow apps built on top of opamp-go server to inject Middleware. However, I think opamp-go's server should maintain trace integrity, by default, so I would've vote for this as Plan A.

tigrannajaryan:
Thanks for opening this.

I support instrumenting opamp-go with one caveat: I don't want Otel Go SDK to be a required dependency. It should be possible to use opamp-go without Otel Go SDK. The reason for is that we believe it is important for OpAMP to be possible to use even if you are not an OpenTelemetry user otherwise.

There are probably a couple options:
- Add instrumentation in a separate Go module in this repository and allow the users to easily opt-in to it.
- Add instrumentation for opamp-go at https://github.com/open-telemetry/opentelemetry-go-contrib/

I am open to exploring other options.

@open-telemetry/opamp-go-maintainers @open-telemetry/opamp-go-approvers thoughts?

gdfast:
Thanks for the note Tigran, I'm glad I read it before trying to implement anything! And I see what you're going for with avoiding the OTel Go SDK dependency. I'm a bit new to golang but I'll check out those two options you shared.

PR #263, opened 2024-03-15
feat: Add optional HTTP Middleware function to StartSettings for serverimpl

gdfast:
Problem
---------------------

Traces for HTTP requests to the opamp-go server break, see https://github.com/open-telemetry/opamp-go/issues/253

Solution
---------------------

- Add an HTTP Handler middleware function to `StartSettings`
- If this function is configured, apply it in serverimpl's `Start` where the HTTP Handler is set
- (add unit tests)

Code review notes
---------------------

- This is a step in addressing https://github.com/open-telemetry/opamp-go/issues/253 but mostly just for HTTP clients and requests. There is likely more to do for maintaining trace linkage through requests that come over websocket connections
- I figured if users are using `Attach` instead of `Start`, they might have their own middleware configured for their HTTP server, so it makes more sense to hook this into `StartSettings` and the `Start` function


andykellr:
I was curious why we weren't running into this and realized that we're using Attach. I'm fine with adding this, but I would expect most users to use Attach. @tigrannajaryan @evan-bradley please have a look.

andykellr (approved):

server/server.go
```diff
@@ -42,6 +42,10 @@ type StartSettings struct {
 
 	// Server's TLS configuration.
 	TLSConfig *tls.Config
+
+	// HTTPMiddlewareFunc specifies middleware for HTTP messages received by the server.
+	// This function is optional to set.
+	HTTPMiddlewareFunc func(handlerFunc http.HandlerFunc) http.HandlerFunc
```
tigrannajaryan, line 48: I think it would be useful to add a comment explaining that the middleware func will be called for every message when HTTP transport is used and once per connection when WebSocket transport is used.

tigrannajaryan:
> * This is a step in addressing [opamp-go server impl is breaking traces #253](https://github.com/open-telemetry/opamp-go/issues/253) but mostly just for HTTP clients and requests. There is likely more to do for maintaining trace linkage through requests that come over websocket connections

@gdfast Can you please post the outline of the full plan with all the steps? I want to see the big picture before we merge this PR.

tigrannajaryan:
> [CLA Not Signed](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/9522512/424240089/263/#/?version=2)
> 
> * [❌](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/9522512/424240089/263/#/?version=2) - login: @gdfast / name: Greg F . The commit ([bdd080a](https://github.com/open-telemetry/opamp-go/commit/bdd080a61668bdec070c98adef73feb65d24cf86)) is not authorized under a signed CLA. [Please click here to be authorized](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/9522512/424240089/263/#/?version=2). For further assistance with EasyCLA, [please submit a support request ticket](https://jira.linuxfoundation.org/servicedesk/customer/portal/4).

Please sign the CLA.

server/server.go
gdfast: good call, will update

gdfast:
> > * This is a step in addressing [opamp-go server impl is breaking traces #253](https://github.com/open-telemetry/opamp-go/issues/253) but mostly just for HTTP clients and requests. There is likely more to do for maintaining trace linkage through requests that come over websocket connections
> 
> @gdfast Can you please post the outline of the full plan with all the steps? I want to see the big picture before we merge this PR.

My company/team has some HTTP Middleware code that adds OTel instrumentation that I'm planning to use in our OpAMP service that is built around opamp-go. I'm planning to follow up with my leads/management about contributing some version of this middleware in a separate, public Go module as you recommended, but I don't know what they'll say. I think giving users the ability to add their own HTTP Middleware is net good though, right?

Meanwhile, I'll need to start looking into how to propagate context for the messages that come in on a websocket connection. I imagine we may need some similar hook to add middleware in order to facilitate Otel Go SDK instrumentation, but I'm not sure.

I think I mentioned that I'm pretty new to Go and OTel instrumentation, so I'm learning as I go

tigrannajaryan:
> My company/team has some HTTP Middleware code that adds OTel instrumentation that I'm planning to use in our OpAMP service that is built around opamp-go. I'm planning to follow up with my leads/management about contributing some version of this middleware in a separate, public Go module as you recommended, but I don't know what they'll say. I think giving users the ability to add their own HTTP Middleware is net good though, right?

If you end up not contributing your implementation to Otel I would still want to see _some_ example showing how the new capability can be used. It can be a very minimal example, just to demonstrate the idea.

gdfast:
> > My company/team has some HTTP Middleware code that adds OTel instrumentation that I'm planning to use in our OpAMP service that is built around opamp-go. I'm planning to follow up with my leads/management about contributing some version of this middleware in a separate, public Go module as you recommended, but I don't know what they'll say. I think giving users the ability to add their own HTTP Middleware is net good though, right?
> 
> If you end up not contributing your implementation to Otel I would still want to see _some_ example showing how the new capability can be used. It can be a very minimal example, just to demonstrate the idea.

Sure thing. Where did you have in mind for this? I could add to https://github.com/open-telemetry/opamp-go/blob/main/internal/examples/server/opampsrv/opampsrv.go but I want to be careful not to add the OTel Go SDK to the repo unnecessarily. I guess adding to the examples directory is fine though?

tigrannajaryan:
> > If you end up not contributing your implementation to Otel I would still want to see _some_ example showing how the new capability can be used. It can be a very minimal example, just to demonstrate the idea.
> 
> Sure thing. Where did you have in mind for this? I could add to https://github.com/open-telemetry/opamp-go/blob/main/internal/examples/server/opampsrv/opampsrv.go but I want to be careful not to add the OTel Go SDK to the repo unnecessarily. I guess adding to the examples directory is fine though?

Yes, it is fine to add to opampsrv in the examples directory since it is a separate Go module. The example can use the Otel Go SDK.

server/server.go
```diff
@@ -42,6 +42,11 @@ type StartSettings struct {
 
 	// Server's TLS configuration.
 	TLSConfig *tls.Config
+
+	// HTTPMiddleware specifies middleware for HTTP messages received by the server.
+	// Note that the function will be called once for websockets upon connecting and will
+	// be called for every HTTP request. This function is optional to set.
+	HTTPMiddleware func(handler http.Handler) http.Handler
```
gdfast, line 49: I changed the middleware setting to be `func(handler http.Handler) http.Handler` from `func(handler http.HandlerFunc) http.HandlerFunc` because I'm finding that lots of middleware examples – including the middleware created by `otelhttp.NewMiddleware` – follow the `Handler -> Handler` pattern.

To make this change work, I ended up renaming the `server.httpHandler` function in _serverimpl` to `server.ServeHTTP` so that it conforms to the `http.Handler` interface.

gdfast:
> > > If you end up not contributing your implementation to Otel I would still want to see _some_ example showing how the new capability can be used. It can be a very minimal example, just to demonstrate the idea.
> > 
> > 
> > Sure thing. Where did you have in mind for this? I could add to https://github.com/open-telemetry/opamp-go/blob/main/internal/examples/server/opampsrv/opampsrv.go but I want to be careful not to add the OTel Go SDK to the repo unnecessarily. I guess adding to the examples directory is fine though?
> 
> Yes, it is fine to add to opampsrv in the examples directory since it is a separate Go module. The example can use the Otel Go SDK.

@tigrannajaryan I went with a very minimal example of using `otelhttp.NewMiddleware` to create the HTTP middleware, which I think should actually be a good idea for users who want to add OTel tracing!

server/serverimpl.go
```diff
@@ -151,7 +155,7 @@ func (s *server) Addr() net.Addr {
 	return s.addr
 }
 
-func (s *server) httpHandler(w http.ResponseWriter, req *http.Request) {
+func (s *server) ServeHTTP(w http.ResponseWriter, req *http.Request) {
```
tigrannajaryan, line 158: Instead of having an exported `ServeHTTP` function in the `server` struct can we make this unexported and wrap it with a struct that has an exported `ServeHTTP` function and use that on line 86 instead?

I don't want to unnecessarily export a function that we don't intend to be called by our users.

tigrannajaryan (commented):
Mostly LGTM, just one small comment to reduce the public API surface.

internal/examples/server/opampsrv/opampsrv.go
```diff
@@ -54,6 +56,7 @@ func (srv *Server) Start() {
 			},
 		},
 		ListenEndpoint: "127.0.0.1:4320",
+		HTTPMiddleware: otelhttp.NewMiddleware("HTTP opamp-go"),
```
tigrannajaryan, line 59: Nit: [http semconv](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md) says span name should be `{method} {http.route}`. Our default route is `/v1/opamp`. Method is tricky, I think it is "GET" for websocket and "POST" for plain http and we don't here which it will be, so perhaps we skip it and use just the route, e.g.
```suggestion
		HTTPMiddleware: otelhttp.NewMiddleware("/v1/opamp"),
```

gdfast: Good call! I wasn't sure what to put here and keep forgetting to check semcov's

server/serverimpl.go
gdfast: I wasn't too worried about making the change because `server` itself is already unexported and `ServerHTTP` isn't part of the `OpAMPServer interface` . I did go ahead and make that wrapper though. Let me know if this more-or-less is what you were looking for!

gdfast:
Hoping this is good to go now @tigrannajaryan . Thanks for your guidance and feedback!

tigrannajaryan (approved):

tigrannajaryan:
Thank you @gdfast

PR #264, opened 2024-03-19
Upgrade OTel SDK dependencies in the internal/example module

gdfast:
*Problem*
The example code has a number of dependencies on otel packages that are either deprecated or very old. This makes developing and improving the examples challenging, and also doesn't make for the best examples for new users.

*Solution*
- Upgrade the OTel packages to v1.24.0 (the latest version)
- Remove old and deprecated dependent modules where possible


tigrannajaryan:
Thank you @gdfast 

If you haven't already please also run `go mod tidy` in the example directory.

tigrannajaryan:
@arminru we have a codecov report failing with some message about keys. I remember you were doing something with tokens a few days ago. 

Can this be the reason it fails? https://github.com/open-telemetry/opamp-go/actions/runs/8349792083/job/22871138122?pr=264

arminru:
@tigrannajaryan the token should be available via the org secret already but the workflow probably needs to be configured to actually use it. The action might also need an update at some point.

https://docs.codecov.com/docs/adding-the-codecov-token

arminru:
@tigrannajaryan I just checked the workflow yaml and it's already set up correctly. The token itself was the problem, seems like the repo had its own invalid or expired token configured locally that was overriding the org secret. I deleted it, now it should work.

tigrannajaryan:
Great, thanks @arminru !

gdfast:
> Thank you @gdfast
> 
> If you haven't already please also run `go mod tidy` in the example directory.

✔️ I did this before pushing

tigrannajaryan (approved):

PR #269, opened 2024-04-11
Add CheckRedirect callback

echlebek:
This commit adds a CheckRedirect callback that opamp-go will call before
following a redirect from the server it's trying to connect to. Like in
net/http, CheckRedirect can be used to observe the request chain that
the client is taking while attempting to make a connection.

The user can optionally terminate redirect following by returning an
error from CheckRedirect.

Unlike in net/http, the via parameter for CheckRedirect is a slice of
responses. Since the user would have no other way to access these in the
context of opamp-go, CheckRedirect makes them available so that users
can know exactly what status codes and headers are set in the response.

Another small improvement is that the error callback is no longer called
when redirecting. This should help to prevent undue error logging by
opamp-go consumers. Since the CheckRedirect callback is now available,
it also doesn't represent any loss in functionality to opamp-go
consumers.

### DIFF
--- a/client/httpclient_test.go
+++ b/client/httpclient_test.go
@@ -3,13 +3,17 @@ package client
 import (
 	"compress/gzip"
 	"context"
+	"errors"
 	"io"
 	"net/http"
+	"net/http/httptest"
+	"net/url"
 	"sync/atomic"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
 	"google.golang.org/protobuf/proto"
 
 	"github.com/open-telemetry/opamp-go/client/internal"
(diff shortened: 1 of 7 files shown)


client/types/callbacks.go
```diff
@@ -110,6 +111,19 @@ type Callbacks interface {
...
+	// following redirects early.
+	//
+	// CheckRedirect is intended to be similar, although not exactly equivalent,
+	// to net/http.Client's CheckRedirect feature. Unlike in net/http, the via
+	// parameter is a slice of HTTP responses, instead of requests. This gives
+	// an opportunity to users to know what the exact response headers and
+	// status were. The request itself can be obtained from the response.
+	//
+	// The responses in the via parameter are passed with their bodies closed.
+	CheckRedirect(req *http.Request, via []*http.Response) error
```
tigrannajaryan, line 126: Is it valid to return `ErrUseLastResponse`? I assume it is not valid at least for WS transport since it typically means there is no WS connection established.

tigrannajaryan: For clarity we should probably also link to https://pkg.go.dev/net/http#Client.CheckRedirect somewhere in the comment above.

tigrannajaryan: This add `CheckRedirect` to `Callbacks` but implements it only for `wsClient`. I think we also need an implementation for `httpClient`.

echlebek: @tigrannajaryan Thank you for the review! 

Regarding `ErrUseLastResponse`, I would agree it is not applicable here. I think it should result in the same behaviour as returning any other error, however. I'll mention it specifically in the docs.

I agree that we should have an implementation for `httpClient`. Perhaps the name of this callback should be `WSCheckRedirect` or similar, denoting that it is only for websocket clients. I don't think the function signature, as implemented, is valid for HTTP clients.

I looked at `httpsender` and it uses the default HTTP client (with an override method for TLS config). Perhaps `httpClient` could accept a `net/http.Client` supplied by the library consumer, instead of having a specific API for things like CheckRedirect. Then the API surface would not grow, and library consumers would be able to use any custom configuration of the HTTP client they desire. Any thoughts on this?

echlebek: HTTP clients should rightly expect to have the signature `func(*http.Request, []*http.Request) error`, as that would be 1:1 with `net/http`.

tigrannajaryan: > HTTP clients should rightly expect to have the signature `func(*http.Request, []*http.Request) error`, as that would be 1:1 with `net/http`.

Do we have the need to use a different signature for WS? I see it is mentioned in the comment but it is not clear to me how exactly you would use the response for WS and why it is useful for WS and not for http.

Ideally we should use the same signature as the standard lib and make this setting applicable to both WS and plain HTTP implementations.

echlebek: I think there is a need for situations where the opamp-go users wants to know what status the server wrote. 301 vs 302 for instance, where the client may want to cache permanent redirects but not cache temporary redirects. Since the opamp-go user isn't privy to the HTTP response, there is no other opportunity to know that information.

I'm definitely open to another design, as I agree it would be better to maintain parity with net/http in general.

tigrannajaryan: @echlebek  Sorry for long pause, I finally got back to this PR.

What do you think if we change `CheckRedirect` to look like this:
```go
CheckRedirect(req *http.Request, viaReq []*http.Request, viaResp []*http.Response) error
```

The wsClient will pass both viaReq and viaResp parameters, while httpClient will only pass viaReq and will leave viaResp nil.

echlebek: Sorry for the even longer pause! That suggestion sounds good to me, I will rebase this PR and make the change.

echlebek:
@tigrannajaryan Apologies, I had actually forgotten to implement the HTTP client logic and tests, so it took a bit longer. However, I think now the feature is complete for both WS and HTTP clients. Please take another look when you get a chance.

client/internal/httpsender.go
```diff
@@ -98,6 +98,19 @@ func (h *HTTPSender) Run(
 	h.callbacks = callbacks
 	h.receiveProcessor = newReceivedProcessor(h.logger, callbacks, h, clientSyncedState, packagesStateProvider, capabilities, packageSyncMutex)
 
+	if callbacks != nil {
+		cs, ok := callbacks.(types.CallbacksStruct)
```
tigrannajaryan, line 102: The `CallbacksStruct` is not supposed to be the only way to implement the callbacks. What is the caller supplies a different implementation?

I think we need to find a way to make the default behavior work without relying on the fact that CallbacksStruct is a struct and may have a nil CheckRedirectFunc field.

echlebek: What can be done alternatively is to remove the conditional check here and unconditionally call the CheckRedirect callback. The `CallbacksStruct` can do its own nil check instead, and use the default `net/http` redirect policy. However, this has a side effect of placing the burden for setting redirect policy on any implementer of the type, which is going to break any existing implementations in a way that is probably worse than just requiring another method to return nil on.

I think this is kind of an inherent problem in the callbacks API. It's not possible to know if a callback has been applied, so all callbacks need to be called unconditionally.

tigrannajaryan: @echlebek I agree. I created a draft PR that gets rid of Callbacks interface in favour of the struct and allows default behaviors to be defined per callback func. Please take a look: https://github.com/open-telemetry/opamp-go/pull/324

echlebek:
Please forgive a bit of churn as I fix up some issues found after rebasing.

echlebek:
I think the PR is ready for review again.

tigrannajaryan (approved):
LGTM.
I will keep it open for a bit in case other approvers want to take a look.

client/httpclient_test.go
```diff
@@ -223,3 +227,87 @@ func TestHTTPClientStartWithZeroHeartbeatInterval(t *testing.T) {
...
+		},
+		{
+			Name:         "check redirect",
+			Redirector:   redirectServer("http://"+redirectee.Endpoint, 302),
+			MockRedirect: mockRedirectHTTP(t, 1, nil),
+		},
+		{
+			Name:         "check redirect returns error",
+			Redirector:   redirectServer("http://"+redirectee.Endpoint, 302),
+			MockRedirect: mockRedirect(t, 1, errors.New("hello")),
```
srikanthccv, line 261: should this be `mockRedirectHTTP`?

echlebek: Good catch, yes it should. I've changed it.

srikanthccv (approved):

PR #279, opened 2024-05-27
Fix test TestOfferUpdatedVersion

tigrannajaryan:
The test sometimes fails on my machine. The EventuallyExpect function needs a condition that returns false instead of failing when the expected message is not received, since a different message may be sent by the client.

This fixes the test. I run it 100 times and it passes: `go test -run TestOfferUpdatedVersion/ws -v -race -count 100`

(Apologies for whitespace changes, but they appear to be incorrectly formatted, my editor autoformatted it and `go fmt` passes on the text).

### DIFF
--- a/client/clientimpl_test.go
+++ b/client/clientimpl_test.go
@@ -533,7 +533,7 @@ func TestSetEffectiveConfig(t *testing.T) {
 			t,
 			func() bool {
 				return rcvConfig.Load() != nil &&
-						proto.Equal(sendConfig, rcvConfig.Load().(*protobufs.EffectiveConfig))
+					proto.Equal(sendConfig, rcvConfig.Load().(*protobufs.EffectiveConfig))
 			},
 		)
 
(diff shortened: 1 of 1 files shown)


client/clientimpl_test.go
```diff
@@ -1208,12 +1208,15 @@ type packageTestCase struct {
...
+	testCase packageTestCase,
+	msg *protobufs.AgentToServer) (*protobufs.ServerToAgent, bool) {
 	expectedStatusReceived := false
 
 	status := msg.PackageStatuses
-	require.NotNil(t, status)
+	if status == nil {
+		// PackageStatuses is not yet reported, keep waiting.
+		return nil, false
+	}
```
tigrannajaryan, line 1219: This is the fix.

tigrannajaryan:
@phanidevavarapu @srikanthccv please review.

srikanthccv (approved):
Thanks

Issue #198, opened 2023-09-11
Capability to set RequestInstanceUid flag in AgentToServer msg as per the spec


sairam866:
I am trying to implement our use case of the Agent requesting the Server for the identifier. The specification supports it, but its yet implemented in the opamp-go.
https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agenttoserverinstance_uid. 
I can contribute to the go implementation. Any pointers for the changes will be helpful.

srikanthccv:
We need to add `SetFlags` on the OpAMPClient interface. And `ClientCommon` implementation to schedule to send it in the next message.  https://github.com/open-telemetry/opamp-go/blob/9efb265ce3818016ed36ecfa0fc3ce86c5c1cf96/client/client.go#L11-L83

Just to note, If your server implementation sends a non-nil `AgentIdentification` field, the new instance ID is always honoured and will be set for subsequent messages.

PR #286, opened 2024-06-10
Enable clients to set flags

tpaschalis:
Updates #198 

This PR adds a new method to the OpAMPClient that allows clients to set their own flags (such as `RequestInstanceUid`).

This new method deviates a little bit from other ones as it just sets a uint64 instead of a pointer struct field, and doesn't use the same errors and nil checks.

Thank you in advance for taking and look, and please let me know if I'm missing anything! 🙌

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -108,7 +108,7 @@ type OpAMPClient interface {
 	SetCustomCapabilities(customCapabilities *protobufs.CustomCapabilities) error
 
 	// SetFlags modifies the set of flags supported by the client.
-	// May be called anytime after Start(), including from OnMessage handler.
+	// May be called before or after Start(), including from OnMessage handler.
 	// The zero value of protobufs.AgentToServerFlags corresponds to FlagsUnspecified
 	// and is safe to use.
 	//
--- a/client/client.go
+++ b/client/client.go
@@ -107,6 +107,16 @@ type OpAMPClient interface {
 	// for more details.
 	SetCustomCapabilities(customCapabilities *protobufs.CustomCapabilities) error
 
+	// SetFlags modifies the set of flags supported by the client.
+	// May be called anytime after Start(), including from OnMessage handler.
+	// The zero value of protobufs.AgentToServerFlags corresponds to FlagsUnspecified
+	// and is safe to use.
+	//
+	// See
+	// https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agenttoserverflags
+	// for more details.
+	SetFlags(flags protobufs.AgentToServerFlags)
+
 	// SendCustomMessage sends the custom message to the Server. May be called anytime after
 	// Start(), including from OnMessage handler.
 	//
--- a/client/clientimpl_test.go
+++ b/client/clientimpl_test.go
@@ -2122,3 +2122,53 @@ func TestSetCustomCapabilities(t *testing.T) {
 		assert.NoError(t, err)
 	})
 }
+
+// TestSetFlags tests the ability for the client to change the set of flags it sends.
+func TestSetFlags(t *testing.T) {
+	testClients(t, func(t *testing.T, client OpAMPClient) {
+
+		// Start a Server.
+		srv := internal.StartMockServer(t)
+		var rcvCustomFlags atomic.Value
+		var flags protobufs.AgentToServerFlags
+
+		srv.OnMessage = func(msg *protobufs.AgentToServer) *protobufs.ServerToAgent {
+			if msg.Flags != 0 {
+				rcvCustomFlags.Store(msg.Flags)
+			}
+			return nil
+		}
+
+		settings := types.StartSettings{}
+		settings.OpAMPServerURL = "ws://" + srv.Endpoint
+		prepareClient(t, &settings, client)
+
+		assert.NoError(t, client.Start(context.Background(), settings))
+
+		// The zero value of AgentToServerFlags is ready to use
+		client.SetFlags(flags)
+
+		// Update flags to send
+		flags |= protobufs.AgentToServerFlags_AgentToServerFlags_RequestInstanceUid
+		client.SetFlags(flags)
+
+		// Verify new flags were delivered to the server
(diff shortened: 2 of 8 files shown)


tpaschalis:
cc @srikanthccv I followed your hint over at #198, let me know if I'm missing anything 🙌

tigrannajaryan:
I think this is missing a couple additional things:

1. Use the Flags from ClientSyncedState in [PrepareFirstMessage](https://github.com/open-telemetry/opamp-go/blob/2fa19cf0e5eb7045c839f3024311d9bce700c165/client/internal/clientcommon.go#L210). This will allow calling SetFlags() before Start() too, which I imagine is useful so that we ask for a new instance id immediately, not after the first request is sent.

2. Also use the Flags from ClientSyncedState when the server asks `ServerToAgentFlags_ServerToAgentFlags_ReportFullState` [here](https://github.com/open-telemetry/opamp-go/blob/2fa19cf0e5eb7045c839f3024311d9bce700c165/client/internal/receivedprocessor.go#L187). This is important since the agent may reconnect to a different server instance after the first sending attempt is done but reply carrying the new instance id was not received.

3. Once 1 and 2 are added we need to reset the `RequestInstanceUid` flag after we receive `new_instance_uid` from the Server, otherwise we will keep asking for new ids every time. Perhaps do it in [rcvAgentIdentification](https://github.com/open-telemetry/opamp-go/blob/2fa19cf0e5eb7045c839f3024311d9bce700c165/client/internal/receivedprocessor.go#L234) after SetInstanceUid returns success.

I may be missing some more things. This needs a bit more thought and additional test cases that demonstrate the real use cases and behavior in presence of failures.

tpaschalis:
Thanks for the feedback @tigrannajaryan

I focused only on the client-side implementation but I realize it's a more interconnected change. I'll try to set up an end-to-end test to have at hand and figure out the entire lifecycle of the IDs.

tpaschalis:
Hey there @tigrannajaryan @srikanthccv 👋 Hope your week is going well!

I've pushed a couple of commits; bd001cf that handles the cases Tigran mentioned on your last message and ca66f1d that adds a test for how messages containing valid/invalid AgentIdentification fields interact with the state of the flag we're tracking.

Let me know how this looks, and if you have any concerns, or ideas. I'm still trying to understand all the moving pieces, so if I'm missing anything please let me know!

client/internal/receivedprocessor.go
```diff
@@ -190,8 +190,9 @@ func (r *receivedProcessor) rcvFlags(
 				msg.RemoteConfigStatus = r.clientSyncedState.RemoteConfigStatus()
 				msg.PackageStatuses = r.clientSyncedState.PackageStatuses()
 				msg.CustomCapabilities = r.clientSyncedState.CustomCapabilities()
+				msg.Flags = uint64(r.clientSyncedState.flags)
```
tigrannajaryan, line 193: Ditto.

client/internal/clientstate.go
```diff
@@ -168,3 +169,11 @@ func (s *ClientSyncedState) HasCustomCapability(capability string) bool {
 
 	return false
 }
+
+// SetFlags sets the flags in the state.
+func (s *ClientSyncedState) SetFlags(flags protobufs.AgentToServerFlags) {
```
tigrannajaryan, line 174: Let's also add a mutex-protected Flags() getter and use it elsewhere.

client/clientimpl_test.go
```diff
@@ -93,7 +93,7 @@ func eventually(t *testing.T, f func() bool) {
 	assert.Eventually(t, f, 5*time.Second, 10*time.Millisecond)
 }
 
-func newInstanceUid(t *testing.T) types.InstanceUid {
+func genNewInstanceUid(t *testing.T) types.InstanceUid {
```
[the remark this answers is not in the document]
tpaschalis: This change is related to making `TestAgentIdentification` more realistic; the test has a `newInstanceUid := newInstanceUid(t)` statement in which the function shadows the variable of the same name, rendering the test harder to follow, needing more variables etc.

I'll see if there's an easy way to fix it, otherwise I can revert changes and leave that test as-is.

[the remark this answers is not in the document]
tigrannajaryan: That's fine, keep the change.

client/client.go
```diff
@@ -107,6 +107,16 @@ type OpAMPClient interface {
 	// for more details.
 	SetCustomCapabilities(customCapabilities *protobufs.CustomCapabilities) error
 
+	// SetFlags modifies the set of flags supported by the client.
+	// May be called anytime after Start(), including from OnMessage handler.
```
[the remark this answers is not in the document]
tpaschalis: Good catch, amending the docstring.

client/clientimpl_test.go
```diff
@@ -2122,3 +2119,53 @@ func TestSetCustomCapabilities(t *testing.T) {
...
+		}
+
+		settings := types.StartSettings{}
+		settings.OpAMPServerURL = "ws://" + srv.Endpoint
+		prepareClient(t, &settings, client)
+
+		assert.NoError(t, client.Start(context.Background(), settings))
+
+		// The zero value of AgentToServerFlags is ready to use
+		client.SetFlags(flags)
```
tigrannajaryan, line 2146: Would it be possible to add one more test that verifies the SetFlags() call before Start()?

tigrannajaryan (approved):
LGTM. Just one comment left, would be great to add one more test and we can merge.

client/clientimpl_test.go
tpaschalis: I've added a testcase in https://github.com/open-telemetry/opamp-go/pull/286/commits/1f74dfd599541fa5266e7dfe3b659945616e09a2 that calls SetFlags before Start, and verifies that they were correctly set up in the very first message. Is that along the lines of what you had in mind?

tigrannajaryan:
Thank you @tpaschalis

PR #291, opened 2024-07-01
Fix race condition around supervisor's Commander

### LINKED ISSUE #290: TestNewSupervisor has a flaky race condition
I suddenly got a flake from `TestNewSupervisor` while running `make all` on main (ed38d5f4bf930b57e04581919fbfb676aaa5a5af).

```
$ make all
go test -race ./...
?       github.com/open-telemetry/opamp-go/client/types [no test files]
?       github.com/open-telemetry/opamp-go/internal/testhelpers [no test files]
?       github.com/open-telemetry/opamp-go/protobufs    [no test files]
?       github.com/open-telemetry/opamp-go/protobufshelpers     [no test files]
?       github.com/open-telemetry/opamp-go/server/types [no test files]
ok      github.com/open-telemetry/opamp-go/client       18.988s
ok      github.com/open-telemetry/opamp-go/client/internal      (cached)
ok      github.com/open-telemetry/opamp-go/internal     (cached)
ok      github.com/open-telemetry/opamp-go/server       (cached)
cd internal/examples && go test -race ./...
?       github.com/open-telemetry/opamp-go/internal/examples/agent      [no test files]
?       github.com/open-telemetry/opamp-go/internal/examples/agent/agent        [no test files]
?       github.com/open-telemetry/opamp-go/internal/examples/server     [no test files]
?       github.com/open-telemetry/opamp-go/internal/examples/server/certman     [no test files]
?       github.com/open-telemetry/opamp-go/internal/examples/server/data        [no test files]
?       github.com/open-telemetry/opamp-go/internal/examples/server/opampsrv    [no test files]
?       github.com/open-telemetry/opamp-go/internal/examples/server/uisrv       [no test files]
?       github.com/open-telemetry/opamp-go/internal/examples/supervisor [no test files]
?       github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor/commander    [no test files]
?       github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor/config       [no test files]
?       github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor/healthchecker        [no test files]
2024/06/30 20:05:03.280981 [OPAMP] Could not load TLS config, working without TLS: open ../../certs/certs/ca.cert.pem: no such file or directory
==================
WARNING: DATA RACE
Write at 0x00c0001e3160 by goroutine 10:
  github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor/commander.(*Commander).Start()
      /home/tpaschalis/GitRepos/opamp-go/internal/examples/supervisor/supervisor/commander/commander.go:52 +0x226
  github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor.(*Supervisor).startAgent()
      /home/tpaschalis/GitRepos/opamp-go/internal/examples/supervisor/supervisor/supervisor.go:415 +0x64
  github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor.(*Supervisor).runAgentProcess()
      /home/tpaschalis/GitRepos/opamp-go/internal/examples/supervisor/supervisor/supervisor.go:512 +0x79a
  github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor.NewSupervisor.func1()
      /home/tpaschalis/GitRepos/opamp-go/internal/examples/supervisor/supervisor/supervisor.go:111 +0x33

Previous read at 0x00c0001e3160 by goroutine 7:
  github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor/commander.(*Commander).Stop()
      /home/tpaschalis/GitRepos/opamp-go/internal/examples/supervisor/supervisor/commander/commander.go:119 +0x64
  github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor.(*Supervisor).Shutdown()
      /home/tpaschalis/GitRepos/opamp-go/internal/examples/supervisor/supervisor/supervisor.go:540 +0xae
  github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor.TestNewSupervisor()
      /home/tpaschalis/GitRepos/opamp-go/internal/examples/supervisor/supervisor/supervisor_test.go:63 +0x175
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1595 +0x261
  testing.(*T).Run.func1()
      /usr/local/go/src/testing/testing.go:1648 +0x44

Goroutine 10 (running) created at:
  github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor.NewSupervisor()
      /home/tpaschalis/GitRepos/opamp-go/internal/examples/supervisor/supervisor/supervisor.go:111 +0x8f9
  github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor.TestNewSupervisor()
      /home/tpaschalis/GitRepos/opamp-go/internal/examples/supervisor/supervisor/supervisor_test.go:60 +0x147
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1595 +0x261
  testing.(*T).Run.func1()
      /usr/local/go/src/testing/testing.go:1648 +0x44

Goroutine 7 (running) created at:
  testing.(*T).Run()
      /usr/local/go/src/testing/testing.go:1648 +0x845
  testing.runTests.func1()
      /usr/local/go/src/testing/testing.go:2054 +0x84
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1595 +0x261
  testing.runTests()
      /usr/local/go/src/testing/testing.go:2052 +0x8ad
  testing.(*M).Run()
      /usr/local/go/src/testing/testing.go:1925 +0xcd7
  main.main()
      _testmain.go:47 +0x2bd
==================
--- FAIL: TestNewSupervisor (0.00s)
    testing.go:1465: race detected during execution of test
FAIL
FAIL    github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor      0.009s
FAIL
make: *** [makefile:18: test] Error 1
```

The race condition is _not_ caught every time by the race detector.

tpaschalis:
From what I'd see under normal circumstances (and also the example), starting up the supervisor which also launches the commander would happen in a separate goroutine than the one that inspects its state or calls Stop.

The current implementation has a race condition on the `cmd *exec.Cmd` field, using it both for figuring out whether the commander is running and also actually executing a new Agent process, which could potentially lead to a panic.

This PR fixes a small race condition around the supervisor's Commander by utilizing an existing `running` field and turning it into a `atomic.Bool` to store the commander's state.

Fixes #290

### DIFF
--- a/internal/examples/supervisor/supervisor/commander/commander.go
+++ b/internal/examples/supervisor/supervisor/commander/commander.go
@@ -125,15 +125,14 @@ func (c *Commander) IsRunning() bool {
 // and if the process does not finish kills it forcedly by sending SIGKILL.
 // Returns after the process is terminated.
 func (c *Commander) Stop(ctx context.Context) error {
-	if !c.IsRunning() {
-		// Not started, nothing to do.
-		return nil
-	}
-
 	c.isStoppingMutex.Lock()
 	c.isStoppingFlag = true
 	c.isStoppingMutex.Unlock()
 
+	if !c.IsRunning() {
+		// Not started, nothing to do.
+		return nil
+	}
 	c.logger.Debugf(ctx, "Stopping agent process, PID=%v", c.cmd.Process.Pid)
 
 	// Gracefully signal process to stop.
--- a/internal/examples/supervisor/supervisor/commander/commander.go
+++ b/internal/examples/supervisor/supervisor/commander/commander.go
@@ -23,7 +23,7 @@ type Commander struct {
 	cmd     *exec.Cmd
 	doneCh  chan struct{}
 	waitCh  chan struct{}
-	running int64
+	running atomic.Bool
 }
 
 func NewCommander(logger types.Logger, cfg *config.Agent, args ...string) (*Commander, error) {
@@ -63,7 +63,7 @@ func (c *Commander) Start(ctx context.Context) error {
 	}
 
 	c.logger.Debugf(ctx, "Agent process started, PID=%d", c.cmd.Process.Pid)
-	atomic.StoreInt64(&c.running, 1)
+	c.running.Store(true)
 
 	go c.watch()
 
@@ -83,7 +83,7 @@ func (c *Commander) Restart(ctx context.Context) error {
 func (c *Commander) watch() {
 	c.cmd.Wait()
 	c.doneCh <- struct{}{}
-	atomic.StoreInt64(&c.running, 0)
+	c.running.Store(false)
 	close(c.waitCh)
 }
 
@@ -102,21 +102,21 @@ func (c *Commander) Pid() int {
 
 // ExitCode returns Agent process exit code if it exited or 0 if it is not.
 func (c *Commander) ExitCode() int {
-	if c.cmd == nil || c.cmd.ProcessState == nil {
+	if c.IsRunning() {
 		return 0
 	}
 	return c.cmd.ProcessState.ExitCode()
 }
 
 func (c *Commander) IsRunning() bool {
-	return atomic.LoadInt64(&c.running) != 0
+	return c.running.Load()
 }
 
 // Stop the Agent process. Sends SIGTERM to the process and wait for up 10 seconds
 // and if the process does not finish kills it forcedly by sending SIGKILL.
 // Returns after the process is terminated.
 func (c *Commander) Stop(ctx context.Context) error {
-	if c.cmd == nil || c.cmd.Process == nil {
+	if !c.IsRunning() {
 		// Not started, nothing to do.
 		return nil
 	}
@@ -159,7 +159,7 @@ func (c *Commander) Stop(ctx context.Context) error {
 	// Wait for process to terminate
 	<-c.waitCh
 
-	atomic.StoreInt64(&c.running, 0)
+	c.running.Store(false)
 
 	// Let goroutine know process is finished.
 	close(finished)
--- a/internal/examples/supervisor/supervisor/commander/commander.go
+++ b/internal/examples/supervisor/supervisor/commander/commander.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"sync"
 	"sync/atomic"
 	"syscall"
 	"time"
@@ -24,6 +25,10 @@ type Commander struct {
 	doneCh  chan struct{}
 	waitCh  chan struct{}
 	running atomic.Bool
+
+	// True when stopping is in progress.
+	isStoppingFlag  bool
+	isStoppingMutex sync.RWMutex
 }
 
 func NewCommander(logger types.Logger, cfg *config.Agent, args ...string) (*Commander, error) {
@@ -41,6 +46,10 @@ func NewCommander(logger types.Logger, cfg *config.Agent, args ...string) (*Comm
 // Start the Agent and begin watching the process.
 // Agent's stdout and stderr are written to a file.
 func (c *Commander) Start(ctx context.Context) error {
+	if c.IsStopping() {
+		return nil
+	}
+
 	c.logger.Debugf(ctx, "Starting agent %s", c.cfg.Executable)
 
 	logFilePath := "agent.log"
@@ -121,6 +130,10 @@ func (c *Commander) Stop(ctx context.Context) error {
 		return nil
 	}
 
+	c.isStoppingMutex.Lock()
+	c.isStoppingFlag = true
+	c.isStoppingMutex.Unlock()
+
 	c.logger.Debugf(ctx, "Stopping agent process, PID=%v", c.cmd.Process.Pid)
 
 	// Gracefully signal process to stop.
@@ -165,4 +178,11 @@ func (c *Commander) Stop(ctx context.Context) error {
 	close(finished)
 
 	return innerErr
+}
+
+// IsStopping returns true if Stop() was called.
+func (c *Commander) IsStopping() bool {
+	c.isStoppingMutex.RLock()
+	defer c.isStoppingMutex.RUnlock()
+	return c.isStoppingFlag
 }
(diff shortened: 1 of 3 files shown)


internal/examples/supervisor/supervisor/commander/commander.go
```diff
@@ -102,21 +102,21 @@ func (c *Commander) Pid() int {
 
 // ExitCode returns Agent process exit code if it exited or 0 if it is not.
 func (c *Commander) ExitCode() int {
-	if c.cmd == nil || c.cmd.ProcessState == nil {
+	if c.IsRunning() {
```
tpaschalis, line 105: Here I've actually flipped the condition from what it was before, as per the method's docstring.

srikanthccv: Is there a chance this leaves the agent process running in such cases? Since the `NewSupervisor` starts `runAgentProcess` go routine to start the agent process. The `Shutdown` could be called before `c.cmd.Start()` returns. If the `c.cmd.Start()` returns successfully it will continue to run which is not expected.

tpaschalis: Hmm, let me think about this (and if the previous code prevented that from happening). We could potentially store the running value a little earlier to avoid that IIUC.

tpaschalis: Hey there, sorry for taking so long to get back to you. As far as I can tell from the code, this was already the case and it isn't a regression from this PR.

To avoid this, we'd make it the caller's responsibility to call only call Stop() after IsRunning() returns true (for example having a backoff mechanism for properly cleaning up resources), or add more explicit synchronization like we do in clientcommon

https://github.com/open-telemetry/opamp-go/blob/ed38d5f4bf930b57e04581919fbfb676aaa5a5af/client/internal/clientcommon.go#L53-L55

IMO this is a separate issue than the race condition the one's fixed here, so tackling the cleanup of resoueces might be a new discussion. WDYT?

srikanthccv: I do not think it was possible to leave the process running in the background because we are checking if `c.cmd` is non-nil (which is the source of the race condition) instead of running status. I think the `Stop` should be updated to handle this.

tigrannajaryan:
@tpaschalis @srikanthccv is this still in progress?

internal/examples/supervisor/supervisor/commander/commander.go
tpaschalis: Hey there 👋 I've reused the syncrhonization mechanism around stopping from ClientCommon: 1337ec44d3d33aea44dcc8435ae9a5ef9eecc74f

This means that if Shutdown has been called, then the runAgentProcess goroutine will not kickstart a new process. Is that something along the lines of what you had in mind?

tpaschalis:
> @tpaschalis @srikanthccv is this still in progress?

Srikanth had suggested using more explicit synchronization to avoid potentially leaving a dangling process; if the Shutdown method of the supervisor is called, then the `isStoppingFlag` will be set and the call to the commander's Start will abort before starting the new process.

I'm looking forward to feedback!

internal/examples/supervisor/supervisor/commander/commander.go
srikanthccv: Sorry for the delay.

Here is the case I think this still doesn't address

1. `Start()` is called
2. The process begins starting up but hasn't yet reached the point where `c.running.Store(true)` is executed
3. `Shutdown()` is called during this window
4. Since `!c.IsRunning()` evaluates to true at this point, `Shutdown()` returns immediately
5. This could result in the process continuing to run

Does that make sense?

tpaschalis: FYI I've pushed another commit in the Stop method to set the isStoppingFlag immediately.

tpaschalis: So, let me try to get this right.

The commander's Start is called _only_ in two cases
a) In the Supervisor's startAgent.
b) During a restart (not pertinent to our discussion)

The startAgent method is only called via the runAgentProcess; in turn, this is only called as a new goroutine in the NewSupervisor function.

In the Shutdown method, the Commander's Stop sets the isStopping flag.

So let's say that NewSupervisor starts, and the new goroutine is launched.

Immediately, the commander's Shutdown method is called, to call the commander's Stop and set the isStoppingFlag. When the call stack gets to the commander's Start it will immediately exit with the IsStopping method without having a chance to launch any process.

tpaschalis: cc @srikanthccv I've put out a test in f135e3b0ef6c4282388497bb9df315973f7e6120 to try and verify the behavior we want.

I couldn't find any more elegant way than the ugly DelayLogger, but I'd like to see what you think. (This test fails in main, without my fix here).

srikanthccv: >So let's say that NewSupervisor starts, and the new goroutine is launched.

>Immediately, the commander's Shutdown method is called, to call the commander's Stop and set the isStoppingFlag. When the call stack gets to the commander's Start it will immediately exit with the IsStopping method without having a chance to launch any process.

Take this case

NewSupervisor starts, and the goroutine `runAgentProcess` is launched, -> `startAgent` -> `s.commander.Start` -> the execution reaches `c.cmd.Start` https://github.com/open-telemetry/opamp-go/blob/7cdd395ee5cccdc05f7eb332ddb28d4399086aba/internal/examples/supervisor/supervisor/commander/commander.go#L61

And say `cmd.Start` didn't return yet

Now, assume the commander's Shutdown method is called (`c.cmd.Start` still didn't return) so `c.running` is false. In this flow `isStoppingFlag` is not helping because we would still return as `IsRunning` evaluates to true.

I think if we add a waitgroup before the `c.cmd.Start` and wait in shutdown should solve the issue.

tigrannajaryan:
@tpaschalis @srikanthccv I assume you are working on this PR. Please ping me when it is ready.

tigrannajaryan:
@tpaschalis I am closing this. Please reopen if you plan to continue working on it.

PR #296, opened 2024-08-23
Support client heartbeat

haoqixu:
Implement spec change open-telemetry/opamp-spec#190


haoqixu:
@jaronoff97, considering you are the author of open-telemetry/opamp-spec#190, you might find this PR of interest and perhaps may want to review it.

client/internal/wssender.go
```diff
@@ -25,15 +27,24 @@ type WSSender struct {
...
 // messages to the server.
 func NewSender(logger types.Logger) *WSSender {
-	return &WSSender{
+	s := &WSSender{
 		logger:       logger,
 		SenderCommon: NewSenderCommon(),
 	}
+	s.heartbeatIntervalUpdated = make(chan struct{}, 1)
+	s.heartbeatIntervalSeconds.Store(defaultHeartbeatIntervalSeconds)
+	s.heartbeatTimer = time.NewTimer(0)
```
jaronoff97, line 45: Is there a reason we're going with a timer instead of a ticker? In my custom implementation of [this](https://github.com/open-telemetry/opentelemetry-operator/blob/main/cmd/operator-opamp-bridge/agent/agent.go#L64), i used a ticker and I think the code is a bit easier to reason about. Though maybe doing the interval reset is more challenging

jaronoff97 (commented):
just one thought, otherwise this looks great.

client/internal/wssender.go
haoqixu: The [specification](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#opampconnectionsettingsheartbeat_interval_seconds) states that:
> The Agent must count heartbeating interval from the last AgentToServer message sent.

I think a timer is preferable to a ticker as the interval should be reset after each AgentToServer message.

jaronoff97 (approved):

client/internal/httpsender.go
```diff
@@ -275,6 +275,13 @@ func (h *HTTPSender) receiveResponse(ctx context.Context, resp *http.Response) {
 	h.receiveProcessor.ProcessReceivedMessage(ctx, &response)
 }
 
+func (h *HTTPSender) SetHeartbeatInterval(duration time.Duration) {
+	// 0 is invalid for a http sender
```
tigrannajaryan, line 279: Can we return an `error` from this method to indicate invalid values? Negative values should also be invalid.

client/internal/receivedprocessor.go
```diff
@@ -209,6 +210,10 @@ func (r *receivedProcessor) rcvOpampConnectionSettings(ctx context.Context, sett
 		return
 	}
 
+	if r.hasCapability(protobufs.AgentCapabilities_AgentCapabilities_ReportsHeartbeat) {
+		r.sender.SetHeartbeatInterval(time.Duration(settings.Opamp.HeartbeatIntervalSeconds) * time.Second)
```
tigrannajaryan, line 214: Once the error is returned by `SetHeartbeatInterval` we can probably also log the error here to indicate the setting was ignored.

client/internal/wssender.go
tigrannajaryan, line 45: Nit: this uses a mix of initialization styles. Can we use one style?

client/internal/wssender.go
```diff
@@ -62,10 +73,43 @@ func (s *WSSender) StoppingErr() error {
...
+
+func (s *WSSender) shouldSendHeartbeat() <-chan time.Time {
+	t := s.heartbeatTimer
+
+	if !t.Stop() {
+		select {
+		case <-t.C:
+		default:
+		}
+	}
```
tigrannajaryan, line 93: Please add a comment to explain what this does.

client/internal/wssender.go
```diff
@@ -62,10 +73,43 @@ func (s *WSSender) StoppingErr() error {
...
+	// Heartbeat interval is set to Zero, disable heartbeat.
+	return nil
+}
+
 func (s *WSSender) run(ctx context.Context) {
 out:
 	for {
 		select {
+		case <-s.shouldSendHeartbeat():
+			s.NextMessage().Update(func(msg *protobufs.AgentToServer) {})
```
tigrannajaryan, line 109: Is this guaranteed to include `instance_uid` field in the next message like the spec requires for heartbeats?

client/internal/wssender.go
haoqixu: Added.

client/internal/wssender.go
haoqixu: `ClientCommon.PrepareStart()` will set `instance_uid` at the start and `SenderCommon` will include the `instance_uid` field in the message. This works in the same way as HTTP Polling.

tigrannajaryan (approved):

tigrannajaryan:
Thanks @haoqixu !

PR #298, opened 2024-09-06
Add HeaderFunc to allow modifying headers before every request

### LINKED ISSUE #297: Support dynamically generating headers for each client request
I would like to use the headers for token-based authentication, e.g. oAuth2. In this case, tokens may expire and need to be refreshed.

However, opamp-go currently only allows specifying a fixed set of headers in the start settings for the client. 

Instead, it would be useful to provide a function that can dynamically determine headers, so that tokens could be refreshed for authentication, e.g.:

```go
// StartSettings defines the parameters for starting the OpAMP Client.
type StartSettings struct {
  // Current settings remain unchanged ...
 
  HeaderFunc func(base http.Header) http.Header
}
```

Here, HeaderFunc is a function that takes in the "base" static headers, and outputs a set of new headers. This would be called on every request sent to the server, and would allow for dynamic authentication headers in case of token refresh, for instance.

BinaryFissionGames:
Adds a new HeaderFunc to the StartSettings that allows for dynamically editing the headers before each HTTP request made by the OpAMP library.

Closes #297

### DIFF
--- a/client/types/startsettings.go
+++ b/client/types/startsettings.go
@@ -17,6 +17,10 @@ type StartSettings struct {
 	// Optional additional HTTP headers to send with all HTTP requests.
 	Header http.Header
 
+	// Optional function that can be used to modify the HTTP headers
+	// before each HTTP request.
+	HeaderFunc func(http.Header) http.Header
+
 	// Optional TLS config for HTTP connection.
 	TLSConfig *tls.Config
 
--- a/client/types/startsettings.go
+++ b/client/types/startsettings.go
@@ -19,6 +19,7 @@ type StartSettings struct {
 
 	// Optional function that can be used to modify the HTTP headers
 	// before each HTTP request.
+	// Can modify and return the argument or return the argument without modifying.
 	HeaderFunc func(http.Header) http.Header
 
 	// Optional TLS config for HTTP connection.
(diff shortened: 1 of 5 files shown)


andykellr (approved):
LGTM

andykellr:
@tigrannajaryan Please have a look and merge if you approve.

client/types/startsettings.go
```diff
@@ -17,6 +17,10 @@ type StartSettings struct {
 	// Optional additional HTTP headers to send with all HTTP requests.
 	Header http.Header
 
+	// Optional function that can be used to modify the HTTP headers
+	// before each HTTP request.
+	HeaderFunc func(http.Header) http.Header
```
tigrannajaryan, line 22: Can we further clarify the expectations? Since we always clone before passing the header parameter, I assume it is safe to modify and return the argument, no need to clone it again, right? Perhaps say something like:

// Can modify and return the argument or return the argument without modifying.

tigrannajaryan (commented):
LGTM, just one small comment.

client/types/startsettings.go
BinaryFissionGames: Yeah, that's correct, it's safe to modify. Updated the comment.

tigrannajaryan (approved):

PR #301, opened 2024-09-25
Add signature to UpdateContent

### LINKED ISSUE #300: PackageSyncer should verify the signature somehow
See todo here:
https://github.com/open-telemetry/opamp-go/blob/b33ab76c401d0f5c3f179c5905409c1502176f97/client/internal/packagessyncer.go#L293

The packagesSyncer does not have a way to verify the signature. I think passing the signature into UpdateContent would make sense here.

BinaryFissionGames:
Passes the signature field to UpdateContent, so the implementation may verify the signature of the file.

Closes #300


andykellr (approved):
LGTM, thanks

srikanthccv (approved):

PR #304, opened 2024-10-18
Pass opts to otlpmetrichttp.New()

scrohde:
opts is created but never used.  It's necessary to pass opts the otlpmetrichttp.New() so that the configured destinationEnpoint is correctly used.


tigrannajaryan (approved):
Thanks @scrohde

PR #305, opened 2024-10-23
Allow setting initial heartbeat interval

### LINKED ISSUE #303: Allow setting initial heartbeat interval
For agent/server heartbeats, the agent initially starts with a 30 second heartbeat interval, which can then be updated by the server if it chooses to.

However, there doesn't seem to be a way to update the initial interval through the client or in the start settings, which means that unless the server updates the interval, the agent is stuck at this 30 second interval.

For us, this is a problem because we want to disable heartbeats (set the interval to 0) unless the server explicitly requests them by setting the interval, for backwards compatibility reasons.

haoqixu:
Add a `HeartbeatIntervalSecond` option to `StartSettings` to allow setting initial heartbeat interval.

resolve #303

### DIFF
--- a/client/types/startsettings.go
+++ b/client/types/startsettings.go
@@ -53,4 +53,14 @@ type StartSettings struct {
 	// the compression is only effectively enabled if the Server also supports compression.
 	// The data will be compressed in both directions.
 	EnableCompression bool
+
+	// Optional HeartbeatIntervalSecond to configure the heartbeat interval for client.
+	// If nil, the default heartbeat interval (30s) will be used.
+	// If zero, heartbeat will be disabled for a Websocket-based client.
+	//
+	// Note that an HTTP-based client will use the heartbeat interval as its polling interval
+	// and zero is invalid for an HTTP-based client.
+	//
+	// If the ReportsHeartbeat capability is disabled, this option has no effect.
+	HeartbeatIntervalSecond *int
 }
--- a/client/types/startsettings.go
+++ b/client/types/startsettings.go
@@ -3,6 +3,7 @@ package types
 import (
 	"crypto/tls"
 	"net/http"
+	"time"
 
 	"github.com/open-telemetry/opamp-go/protobufs"
 )
@@ -54,7 +55,7 @@ type StartSettings struct {
 	// The data will be compressed in both directions.
 	EnableCompression bool
 
-	// Optional HeartbeatIntervalSecond to configure the heartbeat interval for client.
+	// Optional HeartbeatInterval to configure the heartbeat interval for client.
 	// If nil, the default heartbeat interval (30s) will be used.
 	// If zero, heartbeat will be disabled for a Websocket-based client.
 	//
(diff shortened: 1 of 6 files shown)


client/types/startsettings.go
```diff
@@ -53,4 +53,14 @@ type StartSettings struct {
...
+
+	// Optional HeartbeatIntervalSecond to configure the heartbeat interval for client.
+	// If nil, the default heartbeat interval (30s) will be used.
+	// If zero, heartbeat will be disabled for a Websocket-based client.
+	//
+	// Note that an HTTP-based client will use the heartbeat interval as its polling interval
+	// and zero is invalid for an HTTP-based client.
+	//
+	// If the ReportsHeartbeat capability is disabled, this option has no effect.
+	HeartbeatIntervalSecond *int
```
tigrannajaryan, line 65: Would it be nicer to change the type to `time.Duration` and use shorter `HeartbeatInterval` name? This would also allow shorter than 1 second intervals which is useful for tests.

client/wsclient_test.go
```diff
@@ -93,6 +93,62 @@ func TestWSSenderReportsHeartbeat(t *testing.T) {
...
+
+			// Wait for connection to be established.
+			eventually(t, func() bool { return conn.Load() != nil })
+
+			if tt.expectHeartbeats {
+				assert.Eventually(t, func() bool {
+					return msgCount.Load() >= 2
+				}, 3*time.Second, 100*time.Millisecond)
+			} else {
+				assert.Never(t, func() bool {
```
tigrannajaryan, line 140: I think this case is going to block the test for 3 seconds, waiting for a heartbeat that never arrives. We don't want to extend the duration of tests for marginal gain.

If there is no way to shorten this test I would prefer not to have it. This is a minor test case and I think the increase in test suite duration is not worth it. One possible approach is to perhaps keep the code for this case, but skip it normally. If there is a need to re-test this feature in the future we can remove the Skip() temporarily.

Same for the http client tests.

client/wsclient_test.go
```diff
@@ -93,6 +93,62 @@ func TestWSSenderReportsHeartbeat(t *testing.T) {
...
+				conn.Store(c)
+			}
+			var msgCount atomic.Int64
+			srv.OnMessage = func(msg *protobufs.AgentToServer) *protobufs.ServerToAgent {
+				msgCount.Add(1)
+				return nil
+			}
+
+			// Start an OpAMP/WebSocket client.
+			heartbeatSec := 1
```
tigrannajaryan, line 121: After changing HeartbeatIntervalSecond to duration we can make this shorter (e.g. 1 ms) so that the `clientEnableHeartbeat=true` case finishes quickly and doesn't take 1 second.

client/types/startsettings.go
haoqixu: Initially, I used `time.Duration` and `HeartbeatInterval` but noticed `wssender` only supports second granularity. I'm reconsidering and now think it's better to update `wssender` to support milliseconds.

client/wsclient_test.go
haoqixu: I've reduced the test duration to 50 ms for this test case. If this isn't sufficient, we could consider removing the `expectHeartbeats == false` test case.

tigrannajaryan (approved):

PR #324, opened 2024-12-18
Breaking: Replace Callbacks interface by Callbacks struct

tigrannajaryan:
The interface has the following downsides:
- Impossible to define non-trivial default behavior. Here is an example where it was needed: https://github.com/open-telemetry/opamp-go/pull/269#discussion_r1874634261
- Adding new callbacks requires expanding the interface, which is a breaking change for existing client users.

Getting rid of the interface and keeping just a struct for callbacks solves both problems:
- Arbitrarily complex default behavior can be now defined on the struct if the user does not provide the particular callback func.
- Adding new callback funcs is not a breaking change, existing users won't be affected.

### DIFF
--- a/client/clientimpl_test.go
+++ b/client/clientimpl_test.go
@@ -167,8 +167,8 @@ func TestOnConnectFail(t *testing.T) {
 	testClients(t, func(t *testing.T, client OpAMPClient) {
 		var connectErr atomic.Value
 		settings := createNoServerSettings()
-		settings.Callbacks = types.CallbacksStruct{
-			OnConnectFailedFunc: func(ctx context.Context, err error) {
+		settings.Callbacks = types.Callbacks{
+			OnConnectFailed: func(ctx context.Context, err error) {
 				connectErr.Store(err)
 			},
 		}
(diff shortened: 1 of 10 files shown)


tigrannajaryan:
If we like this direction I will also make similar changes to Server callbacks.

echlebek:
Thanks @tigrannajaryan, I think this is a nice simplification of the callbacks API!

I might suggest preserving the CallbacksStruct type name (although the name is a bit awkward now) for backwards compatibility.

Perhaps add a type alias, to avoid breaking the API for people using CallbacksStruct. It could also be given a "Deprecated" comment, so that users can be made aware of its deprecated status by static analysis tools. (https://go.dev/wiki/Deprecated)
```
// Callbacks is blah blah
type Callbacks struct { ... }

// CallbacksStruct is an alias for Callbacks
// Deprecated: CallbacksStruct will be removed in the future
type CallbacksStruct = Callbacks
```

andykellr:
Thanks, LGTM

tigrannajaryan:
> I might suggest preserving the CallbacksStruct type name (although the name is a bit awkward now) for backwards compatibility.

I am not sure about this. It is not just the struct name, we also have to preserve the field names, which all have the "Func" suffix now, otherwise it is a breaking change anyway.

echlebek:
That's a good point, I didn't notice the field names had changed as well.

andykellr (approved):

PR #326, opened 2024-12-19
Breaking: Replace Callbacks interface by Callbacks struct (server)

tigrannajaryan:
This continues work start in https://github.com/open-telemetry/opamp-go/pull/324 for Client.

The interface has the following downsides:
- Impossible to define non-trivial default behavior. Here is an example where it was needed: https://github.com/open-telemetry/opamp-go/pull/269#discussion_r1874634261
- Adding new callbacks requires expanding the interface, which is a breaking change for existing client users.

Getting rid of the interface and keeping just a struct for callbacks solves both problems:
- Arbitrarily complex default behavior can be now defined on the struct if the user does not provide the particular callback func.
- Adding new callback funcs is not a braking change, existing users won't be affected.


echlebek (commented):
Looks good to me. The change is removing a level of nesting from a lot of places in the codebase too. :+1: 

I do see a minor possible improvement, defaults could be defined as free standing functions and set by name in `SetDefaults`, which would avoid the possibility of closing over function vars by accident. But, that's really a tiny nit.

server/types/callbacks.go
```diff
@@ -28,28 +28,57 @@ type Callbacks interface {
...
+		c.OnConnecting = func(r *http.Request) ConnectionResponse {
+			return ConnectionResponse{Accept: true}
+		}
+	}
 }
 
 // ConnectionCallbacks receives callbacks for a specific connection. An implementation of
 // this interface MUST be set on the ConnectionResponse returned by the OnConnecting
 // callback if Accept=true. The implementation can be shared by all connections or can be
 // unique for each connection.
```
srikanthccv, line 45: Can we change the wording here because it still refers to "interface" and "implementation" here?

server/types/callbacks.go
```diff
@@ -15,7 +15,7 @@ type ConnectionResponse struct {
 	ConnectionCallbacks ConnectionCallbacks
 }
 
-type Callbacks interface {
+type Callbacks struct {
```
srikanthccv, line 18: Same here

> ConnectionCallbacks MUST be set to an implementation of the ConnectionCallbacks interface to handle the connection.

srikanthccv (commented):
minor doc comments 👍

server/types/callbacks.go
tigrannajaryan: Fixed.

server/types/callbacks.go
tigrannajaryan: Fixed.

tigrannajaryan:
> I do see a minor possible improvement, defaults could be defined as free standing functions and set by name in `SetDefaults`, which would avoid the possibility of closing over function vars by accident. But, that's really a tiny nit.

Agreed. Done.

srikanthccv (approved):

PR #331, opened 2024-12-27
feat: Supports callbacks when reading a message fails

### LINKED ISSUE #330: [Feature] Callback on message read error
I want to store the error in the database when there is an error reading the message.

tttoad:
Resolves #330

This adds a new Server callback OnReadMessageError that is called when a read or deserialization error occurs.

### DIFF
--- a/server/serverimpl.go
+++ b/server/serverimpl.go
@@ -5,6 +5,7 @@ import (
 	"compress/gzip"
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"net"
 	"net/http"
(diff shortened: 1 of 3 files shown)


tttoad:
> Where is the new callback used?

Sorry, missed submitting a file change.
PTAL.

tigrannajaryan:
This appears stale - no response on comments. This also conflicts with other changes to callbacks. Closing. Please recreate if still relevant.

server/serverimpl_test.go
```diff
@@ -315,6 +315,52 @@ func TestServerReceiveSendMessage(t *testing.T) {
...
+	srv := startServer(t, settings)
+	defer srv.Stop(context.Background())
+
+	// Connect using a WebSocket client.
+	conn, _, _ := dialClient(settings)
+	require.NotNil(t, conn)
+	defer conn.Close()
+
+	// Send a message to the Server.
+	err := conn.WriteMessage(websocket.TextMessage, []byte(""))
```
tigrannajaryan, line 353: Include some payload to verify it gets delivered to the callback ,e.g.: 
```suggestion
	err := conn.WriteMessage(websocket.TextMessage, []byte("abc"))
```

server/serverimpl_test.go
```diff
@@ -315,6 +315,52 @@ func TestServerReceiveSendMessage(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// Wait until Server receives the message.
+	eventually(t, func() bool { return rcvMsg.Load() != nil })
+	errInfo := rcvMsg.Load().(ErrorInfo)
+	assert.EqualValues(t, websocket.TextMessage, errInfo.mt)
+	assert.EqualValues(t, []byte(""), errInfo.msgByte)
```
tigrannajaryan, line 360:
```suggestion
	assert.EqualValues(t, []byte("abc"), errInfo.msgByte)
```

server/serverimpl.go
```diff
@@ -226,27 +227,39 @@ func (s *server) handleWSConnection(reqCtx context.Context, wsConn *websocket.Co
...
+			err = internal.DecodeWSMessage(msgBytes, &request)
+			if err != nil {
+				s.logger.Errorf(msgContext, "Cannot decode message from WebSocket: %v", err)
+				return false, err
+			}
+			return false, nil
+		}()
 		if err != nil {
-			s.logger.Errorf(msgContext, "Cannot decode message from WebSocket: %v", err)
+			connectionCallbacks.OnReadMessageError(agentConn, mt, msgBytes, err)
```
tigrannajaryan, line 259: In addition, do we want `OnReadMessageError` to return a bool/enum, indicating that the connection must be kept or terminated? This gives additional control to the callback owner, for example to inspect the error and decide that it doesn't want to keep receiving messages from clients that send invalid messages.

I would expect strict serves to prefer this behavior, since it is likely that a misbehaving client will keep misbehaving and we would rather cut the communication early.

Since we are introducing a new callback in the public API, I'd rather make it more universally useful for a more than one case if we can, not just for logging (at a cost of tiny extra complexity of returning and checking a value from callback).

@andykellr thoughts?

server/serverimpl_test.go
tttoad: done.

server/serverimpl_test.go
tttoad: done.

server/types/callbacks.go
```diff
@@ -62,6 +62,9 @@ type ConnectionCallbacks struct {
 
 	// OnConnectionClose is called when the OpAMP connection is closed.
 	OnConnectionClose func(conn Connection)
+
+	// OnConnectionError is called when an error occurs while reading or serializing a message.
```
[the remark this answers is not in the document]
tttoad: > 

done.

server/serverimpl.go
tigrannajaryan: I am resolving this now. We can add the return value later if desired.

tttoad:
@tigrannajaryan PTAL.

server/serverimpl_test.go
```diff
@@ -315,6 +315,52 @@ func TestServerReceiveSendMessage(t *testing.T) {
...
+	}}
+	srv := startServer(t, settings)
+	defer srv.Stop(context.Background())
+
+	// Connect using a WebSocket client.
+	conn, _, _ := dialClient(settings)
+	require.NotNil(t, conn)
+	defer conn.Close()
+
+	// Send a message to the Server.
```
tigrannajaryan, line 352:
```suggestion
	// Send an invalid message to the Server. This should result in calling OnReadMessageError().
```

tigrannajaryan (approved):
LGTM. Just one minor suggestion to improve the comment and we can merge.

server/serverimpl_test.go
tttoad: @tigrannajaryan Done.

Issue #335, opened 2025-01-07
[BUG]:The context of the web socket connection is canceled.


tttoad:
When callback function `OnConnected` is called, `context` is already canceled.

tigrannajaryan:
Closing, see the last comment in https://github.com/open-telemetry/opamp-go/pull/336

PR #336, opened 2025-01-07
fix: `OnConnected` is called with a copied context.

tttoad:
see #335.

### DIFF
--- a/server/context.go
+++ b/server/context.go
@@ -1,0 +1,17 @@
+package server
+
+import (
+	"context"
+	"time"
+)
+
+// Detach returns a context that keeps all the values of its parent context
+// but detaches from the cancellation and error handling.
+func Detach(ctx context.Context) context.Context { return detachedContext{ctx} }
+
+type detachedContext struct{ parent context.Context }
+
+func (v detachedContext) Deadline() (time.Time, bool)       { return time.Time{}, false }
+func (v detachedContext) Done() <-chan struct{}             { return nil }
+func (v detachedContext) Err() error                        { return nil }
+func (v detachedContext) Value(key interface{}) interface{} { return v.parent.Value(key) }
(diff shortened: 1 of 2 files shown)


tttoad:
> Unclear what problem this is solving. Also, we need tests verifying and demonstrating the bug.

When the connection is established, I will write the information about the connection to storage. 
I will pass the context into e.g. (go-redis,gorm...) and they will check if the context has been canceled. When the http request is disconnected, the context will be canceled and the data writing is not finished yet (`handleWSConnection` is asynchronous) and it will report `context canceled`.

[That context is canceled if the HTTP client disconnects or cancels the HTTP request (possible in HTTP/2).](https://go.dev/doc/database/cancel-operations)

tigrannajaryan:
I am not sure I understand the problem correctly, but it seems you are saying you want the OnConnected callback that the user provides to ignore the cancellation of the http request that the server is handling. In that case you can do this in the OnConnected implementation, I don't see why the `httpHandler` needs to be modified.

tttoad:
> I am not sure I understand the problem correctly, but it seems you are saying you want the OnConnected callback that the user provides to ignore the cancellation of the http request that the server is handling. In that case you can do this in the OnConnected implementation, I don't see why the `httpHandler` needs to be modified.

I think we shouldn't pass a context that has been cancelled. So I hope to modify it in `httpHandler`.

tigrannajaryan:
> I think we shouldn't pass a context that has been cancelled. So I hope to modify it in `httpHandler`.

I think it is idiomatic Go to pass the context as is. If the callee needs to check if the context is cancelled and detach from it then the callee can easily do it. I don't see a reason to add this check in the caller. The exact same code with `detachedContext` that you submitted in this PR can be in your callback.

I am going to close this PR as rejected. Please feel free to reopen if I misunderstood the goal.

PR #341, opened 2025-01-10
Client reports download status and download details

michel-laterman:
Add implementation of downloading status along with download details details as described in https://github.com/open-telemetry/opamp-spec/pull/206

When the client is downloading a package it will report the download status, and report download details (bytes per second and percent download) periodically.


client/internal/packagessyncer.go
```diff
@@ -290,7 +296,20 @@ func (s *packagesSyncer) downloadFile(ctx context.Context, pkgName string, file
...
-	err = s.localState.UpdateContent(ctx, pkgName, resp.Body, file.ContentHash, file.Signature)
+	// Package length is required to be able to report download percent.
+	packageLength := -1
+	if contentLength := resp.Header.Get("Content-Length"); contentLength != "" {
+		if length, err := strconv.Atoi(contentLength); err == nil {
+			packageLength = length
+		}
+	}
+	// start the download reporter
+	detailsReporter := newDownloadReporter(downloadReporterDefaultInterval, packageLength) // TODO set interval
```
michel-laterman, line 307: I need some guidance on where the user of a client library can be expected to set a value for the update interval, or if we even need to expose it

srikanthccv: I think this should be made customisable.

client/internal/package_download_details_reporter.go
```diff
@@ -0,0 +1,75 @@
...
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-p.done:
+				return
+			case <-timer.C:
+				downloadTime := time.Now().Sub(p.start)
+				downloaded := p.downloaded.Load()
+				bps := downloaded / uint64(downloadTime/time.Second)
```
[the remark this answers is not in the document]
michel-laterman: I can update the spec for changing the bps to a float

client/internal/packagessyncer.go
michel-laterman: Yes, my comment is more about my confusion as to where we should try to pass the config option from; Should we add `PackageReporterInterval/SetPackageReporterInterval) funcs to the `types.PackagesStateProvider` interface or does it belong somewhere else?

srikanthccv: I think it should be part of `StartSettings`

michel-laterman: `StartSettings` is not currently being passed to the `NewPackagesSyncer` function; do we want to add it as another parameter?

srikanthccv: NewPackagesSyncer is internal, so passing the interval param should be fine. This is similar to how we enabled heartbeat interval.

client/internal/package_download_details_reporter.go
```diff
@@ -0,0 +1,78 @@
...
+				return
+			case <-ticker.C:
+				downloadTime := time.Now().Sub(p.start)
+				downloaded := float64(p.downloaded.Load())
+				bps := downloaded / float64(downloadTime/time.Second)
+				var downloadPercent float64
+				if p.packageLength > 0 {
+					downloadPercent = downloaded / p.packageLength * 100
+				}
+				status.DownloadDetails = &protobufs.PackageDownloadDetails{
```
tigrannajaryan, line 63: I am a bit worried that `status` is a struct owned by someone else and we modify it from our goroutine here, while `status`'s other fields are also simultaneously being modified by the owner. If the code is accidentally modified in the future to modify the same fields we will have a problem.

Can we instead change `updateFn` to accept the `DownloadPercent` and `DownloadBytesPerSecond` as parameters and eliminate the `status` parameter from `downloadReporter`? This way the responsibility for getting concurrency correctly will be sole responsibility of the caller of `downloadReporter.report` instead of being a shared responsibility.

client/internal/package_download_details_reporter.go
```diff
@@ -0,0 +1,78 @@
...
+func (p *downloadReporter) Write(b []byte) (int, error) {
+	n := len(b)
+	p.downloaded.Add(uint64(n))
+	return n, nil
+}
+
+// report periodically updates the package status details and calls the passed upateFn to send the new status.
+func (p *downloadReporter) report(ctx context.Context, status *protobufs.PackageStatus, updateFn func(context.Context, bool) error) {
+	go func() {
+		// Make sure we wait at least 1s before reporting the download rate in bps to avoid a panic
```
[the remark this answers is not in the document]
michel-laterman: I've enforced a 1s min; what used to fail was the bps calc:
`bps := downloaded / float64(downloadTime/time.Second)`

tigrannajaryan:
@michel-laterman are you still working on this? I left some comments above and the build is failing.

michel-laterman:
@tigrannajaryan yes I'm working on it, but my time for this is very limited

client/internal/package_download_details_reporter.go
```diff
@@ -0,0 +1,68 @@
...
+				return
+			case <-ticker.C:
+				downloadTime := time.Now().Sub(p.start)
+				downloaded := float64(p.downloaded.Load())
+				bps := downloaded / float64(downloadTime/time.Second)
+				var downloadPercent float64
+				if p.packageLength > 0 {
+					downloadPercent = downloaded / p.packageLength * 100
+				}
+				_ = updateFn(ctx, downloadPercent, bps)
```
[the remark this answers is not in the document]
michel-laterman: I've changed to use the `protobufs.PackageDownloadDetails` directly

client/internal/inmempackagestore.go
```diff
@@ -92,18 +96,22 @@ func (l *InMemPackagesStore) SetAllPackagesHash(hash []byte) error {
...
 func (l *InMemPackagesStore) LastReportedStatuses() (*protobufs.PackageStatuses, error) {
+	l.statusLock.RLock()
+	defer l.statusLock.RUnlock()
 	return l.lastReportedStatuses, nil
 }
 
 func (l *InMemPackagesStore) SetLastReportedStatuses(statuses *protobufs.PackageStatuses) error {
+	l.statusLock.Lock()
 	l.lastReportedStatuses = statuses
+	l.statusLock.Unlock()
```
tigrannajaryan, line 115: Nit: the function above uses the `defer` form. Would be nice to do the same here for consistency.

tigrannajaryan (approved):
LGTM. Please resolve the conflicts.

client/internal/inmempackagestore.go
```diff
@@ -92,18 +96,22 @@ func (l *InMemPackagesStore) SetAllPackagesHash(hash []byte) error {
 }
 
 func (l *InMemPackagesStore) GetContent() map[string][]byte {
-	return l.fileContents
+	return maps.Clone(l.fileContents)
```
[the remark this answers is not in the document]
michel-laterman: There were test failures if we returned the map directly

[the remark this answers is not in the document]
tigrannajaryan: Unclear why this would be the case. I see it is used in place only, in the test, supposedly after the syncing is done, so there should not be any races, unless we have bugs in the code.
I would like to understand what the problem is. Will you be able to look into it? Cloning the map may be masking the problem. Blocking until we have clarity and until we are certain this PR did not introduce a problem.

tigrannajaryan (changes requested):

michel-laterman:
There's currently a race condition in a test I have yet to debug

client/internal/receivedprocessor.go
```diff
@@ -41,6 +45,7 @@ func newReceivedProcessor(
 	packagesStateProvider types.PackagesStateProvider,
 	capabilities protobufs.AgentCapabilities,
 	packageSyncMutex *sync.Mutex,
+	reporterInt time.Duration,
```
tigrannajaryan, line 48:
```suggestion
	downloadReporterInt time.Duration,
```

client/internal/receivedprocessor.go
```diff
@@ -50,6 +55,7 @@ func newReceivedProcessor(
 		packagesStateProvider: packagesStateProvider,
 		capabilities:          capabilities,
 		packageSyncMutex:      packageSyncMutex,
+		reporterInterval:      reporterInt,
```
tigrannajaryan, line 58:
```suggestion
		downloadReporterInterval:      reporterInt,
```

client/clientimpl_test.go
```diff
@@ -1384,11 +1385,15 @@ func verifyUpdatePackages(t *testing.T, testCase packageTestCase) {
 			// Wait until all syncing is done.
 			<-syncerDoneCh
 
-			for pkgName, receivedContent := range localPackageState.GetContent() {
+			// Clone localPackageState instead of using directly to avoid a race when testing for the DOWNLADING status
```
tigrannajaryan, line 1388: The previous line supposedly wait until all synching is done. Doesn't that mean status can no longer be modified after the previous line and that localPackageState will not change anymore? What code is doing a concurrent write that requires cloning the state? And if there is code that is doing concurrent write, isn't Clone() operation itself under the risk of racing with that?

michel-laterman: I think what's occurring is that `<-syncerDonCh` gets unblocked when the expected status is found; if this is set to `Downloading`, the package sync continues in the background. I'll try to reqork the test to avoid cloning

client/clientimpl_test.go
```diff
@@ -1501,6 +1525,8 @@ func createPackageTestCase(name string, downloadSrv *httptest.Server) packageTes
 		expectedSignature: map[string][]byte{
 			"package1": {6, 7},
 		},
+
+		temporaryStatuses: make([]*protobufs.PackageStatuses, 0),
```
tigrannajaryan, line 1529: Nit: I think a `nil` slice should also work here, no need to allocate zero-length slice.

client/clientimpl_test.go
```diff
@@ -1266,6 +1266,12 @@ type packageTestCase struct {
 	expectedFileContent map[string][]byte
 	expectedSignature   map[string][]byte
 	expectedError       string
+
+	// temporaryStatuses is a slice used by a test case to check if a non-final package status occurs.
+	// When a PackageStatuses message is added to the slice only the Packages name and Status will be checked
+	// If they are successfully obsereved then the corresponding entry in observedTemporartStatuses will be marked as true
```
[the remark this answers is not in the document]
tigrannajaryan: Looking at the test on line 1556, we only have one status in the slice, so does this need to be a slice if we never put more than one item in it?

tigrannajaryan (approved):
Thanks @michel-laterman

PR #344, opened 2025-01-21
Add client metrics for http and ws transports

echlebek:
**Not ready for merge**

I've opened this PR in order to spur some discussion on how metrics could be reported by opamp-go clients and servers without introducing a dependency on open-telemetry or another metrics library. The goal here is to let library users inspect opamp-go's internal client metrics.

This PR adds client-side metrics collection for WS and HTTP transports. Collected metrics are in a custom format that can be translated to other metrics definitions by the library user. A TODO at this point is to add some library routines that can translate these metrics into common formats. It's unclear to me exactly where that should happen at the moment.

Counters have been added for message transmission metadata, as well as a ringbuffer that collects message latencies as well as key message attributes.

Library users can pass a `types.ClientMetrics` object to `StartSettings` and then inspect the results periodically. The methods that the counters and ringbuffers expose are goroutine-safe. If the metrics object is not passed, a default one with a minimal buffer is created, so this change should non-breaking for library users.

Some implementation details have been modified in order to ease testing. Specifically, some internal types have been replaced with interfaces in senders and receivers so that they can be replaced with fakes in tests.

The use of `*websocket.Conn` has been replaced with `internal.WebsocketConn` in several places for this reason as well.

Finally, some definitions of the server have changed in order to reflect the need for the number of bytes transmitted to be known by the caller. While this is not necessary for client-side metrics collection, it will help facilitate a similar change to the server so that metrics can be collected there too.

If this change or a change like it is merged, then we could also implement this part of the spec, building on top of the internal collection: https://github.com/observIQ/opamp-spec/blob/808917457e973c251fd6c968bbb2b9dd38cb9f3d/specification.md#own-telemetry-reporting (thanks @rnishtala-sumo for the suggestion)

### DIFF
--- a/client/clientimpl_test.go
+++ b/client/clientimpl_test.go
@@ -141,6 +141,7 @@ func startClient(t *testing.T, settings types.StartSettings, client OpAMPClient)
 func createNoServerSettings() types.StartSettings {
 	return types.StartSettings{
 		OpAMPServerURL: "ws://" + testhelpers.GetAvailableLocalAddress(),
+		Metrics:        types.NewClientMetrics(64),
 	}
 }
 
--- a/client/httpclient.go
+++ b/client/httpclient.go
@@ -37,6 +37,8 @@ func NewHTTP(logger types.Logger) *httpClient {
 
 // Start implements OpAMPClient.Start.
 func (c *httpClient) Start(ctx context.Context, settings types.StartSettings) error {
+	c.sender.SetMetrics(settings.Metrics)
+
 	if err := c.common.PrepareStart(ctx, settings); err != nil {
 		return err
 	}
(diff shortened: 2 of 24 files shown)


tigrannajaryan:
My initial reaction is that this adds significant amount of code to maintain and it is not clear to me whether the added value is worth it. I would want to see a very strong demand for this capability before seriously considering a solution. Absent that I prefer to not have this capability and keep logs the only troubleshooting mechanism in opamp-go for now.

I would like to know the opinion of other @open-telemetry/opamp-go-approvers 

I won't comment on whether the proposed solution is the right one, as I have not spent enough time thinking about it.

tigrannajaryan:
I am going to close this for now since in my opinion it is too much code and complexity to maintain for relatively small gain.

Happy to discuss and reconsider if other approvers chime in and support.

Issue #347, opened 2025-02-11
[HTTP] Server shutdown completes before serve() terminates


rogercoll:
Starting and stopping the HTTP server are not idempotent operations; if the server endpoint has not yet been released, an error is returned. This error can be reproduced with the following test:

```go
func TestServerStartStopIdempotency(t *testing.T) {
	endpoint := testhelpers.GetAvailableLocalAddress()
	for i := 0; i < 10; i++ {
		t.Run(fmt.Sprintf("Attempt #%d: ", i), func(t *testing.T) {
			srv := startServer(t, &StartSettings{
				ListenEndpoint: endpoint,
			})

			err := srv.Start(StartSettings{})
			assert.ErrorIs(t, err, errAlreadyStarted)

			err = srv.Stop(context.Background())
			assert.NoError(t, err)
		})
	}
}
```

Result:

```
$ go test -run TestServerStartStopIdempotency
--- FAIL: TestServerStartStopIdempotency (0.00s)
    --- FAIL: TestServerStartStopIdempotency/Attempt_#1:_ (0.00s)
        serverimpl_test.go:38:
                Error Trace:    serverimpl_test.go:38
                                                        serverimpl_test.go:64
                Error:          Received unexpected error:
                                listen tcp 127.0.0.1:41099: bind: address already in use
                Test:           TestServerStartStopIdempotency/Attempt_#1:_
    --- FAIL: TestServerStartStopIdempotency/Attempt_#3:_ (0.00s)
        serverimpl_test.go:38:
                Error Trace:    serverimpl_test.go:38
                                                        serverimpl_test.go:64
                Error:          Received unexpected error:
                                listen tcp 127.0.0.1:41099: bind: address already in use
                Test:           TestServerStartStopIdempotency/Attempt_#3:_
```

Issue #256, opened 2024-02-28
Data race in example supervisor


echlebek:
There is a data race in the example supervisor implementation. It can be observed sporadically while the tests are running.

```
==================
WARNING: DATA RACE
Write at 0x00c0002b71b0 by goroutine 136:
  github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor/commander.(*Commander).Start()
      /home/eric/go/src/github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor/commander/commander.go:52 +0x226
  github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor.(*Supervisor).startAgent()
      /home/eric/go/src/github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor/supervisor.go:412 +0x64
  github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor.(*Supervisor).runAgentProcess()
      /home/eric/go/src/github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor/supervisor.go:509 +0x79a
  github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor.NewSupervisor.gowrap1()
      /home/eric/go/src/github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor/supervisor.go:112 +0x33

Previous read at 0x00c0002b71b0 by goroutine 133:
  github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor/commander.(*Commander).Stop()
      /home/eric/go/src/github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor/commander/commander.go:119 +0x5c
  github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor.(*Supervisor).Shutdown()
      /home/eric/go/src/github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor/supervisor.go:537 +0xad
  github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor.TestNewSupervisor()
      /home/eric/go/src/github.com/open-telemetry/opamp-go/internal/examples/supervisor/supervisor/supervisor_test.go:63 +0x175
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1689 +0x21e
  testing.(*T).Run.gowrap1()
      /usr/local/go/src/testing/testing.go:1742 +0x44

[...]

PR #348, opened 2025-02-11
fix: wait for HTTP server serve() termination

rogercoll:
Fixes https://github.com/open-telemetry/opamp-go/issues/347

### DIFF
--- a/server/serverimpl.go
+++ b/server/serverimpl.go
@@ -174,7 +174,12 @@ func (s *server) Stop(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
-		s.httpServerServeWg.Wait()
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			s.httpServerServeWg.Wait()
+		}
 	}
 	return nil
 }
--- a/server/serverimpl.go
+++ b/server/serverimpl.go
@@ -40,7 +40,8 @@ type server struct {
 
 	// The listening HTTP Server after successful Start() call. Nil if Start()
 	// is not called or was not successful.
-	httpServer *http.Server
+	httpServer        *http.Server
+	httpServerServeWg *sync.WaitGroup
 
 	// The network address Server is listening on. Nil if not started.
 	addr net.Addr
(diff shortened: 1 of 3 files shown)


server/serverimpl.go
```diff
@@ -159,7 +169,11 @@ func (s *server) Stop(ctx context.Context) error {
 		defer func() { s.httpServer = nil }()
 		// This stops accepting new connections. TODO: close existing
 		// connections and wait them to be terminated.
-		return s.httpServer.Shutdown(ctx)
+		err := s.httpServer.Shutdown(ctx)
+		if err != nil {
+			return err
+		}
+		s.httpServerServeWg.Wait()
```
tigrannajaryan, line 176: This needs to honour `ctx` cancellation. If context is Done(), Stop() needs to return with an error. I am not sure if this is going to be the behavior, given that Wait() can wait indefinitely long. Please add a test that verifies the cancellation case.

tigrannajaryan:
Please update doc-comment of OpAMPServer.Stop() to explain that it waits until the server socket is released.

server/serverimpl.go
rogercoll: Makes sense to me, added in https://github.com/open-telemetry/opamp-go/pull/348/commits/4e4986f743b387b30c1b3725125a22db18369c51

rogercoll:
looks like there is an unrelated flaky test on the examples: 
`2025/02/13 14:49:28.332879 [OPAMP] Could not load TLS config, working without TLS: open ../../certs/certs/ca.cert.pem: no such file or directory`

rogercoll:
related data race issue https://github.com/open-telemetry/opamp-go/issues/256

tigrannajaryan:
> related data race issue #256

I rerun the build a couple times and it fails. Perhaps this PR changed timings and the race is now encountered more consistently. Regardless of the reason I can't merge the PR if the build does not pass. We need to fix #256 first.

rogercoll:
@tigrannajaryan The race condition test is now green after merging main. There are some changes on the files related to formatting which I think it would be better to tackle in a separate PR: https://github.com/open-telemetry/opamp-go/pull/356

tigrannajaryan:
> @tigrannajaryan The race condition test is now green after merging main. There are some changes on the files related to formatting which I think it would be better to tackle in a separate PR: #356

Excellent! Let's get #356 merged and then we will merge this one. Thank you for your patience and for going the extra mile.

tigrannajaryan (approved):

PR #352, opened 2025-02-18
fix: add commander stopping sync mechanism

rogercoll:
Resolves https://github.com/open-telemetry/opamp-go/issues/256

Test that generates the data race:
```go
func TestNewSupervisor(t *testing.T) {
...

	supervisor, err := NewSupervisor(&internal.NopLogger{})
	assert.NoError(t, err)

	supervisor.Shutdown()
}
```

It looks like the commander struct is started in a go routine when the `NewSupervisor` is created. The data race happens when the commander is stopped (`supervisor.Shutdown()`) while being initalized (Start).

The fix adds a synchronization mechanism to prevent the commander being started if it is already being stopped.

### DIFF
--- a/internal/examples/supervisor/supervisor/commander/commander.go
+++ b/internal/examples/supervisor/supervisor/commander/commander.go
@@ -45,8 +45,8 @@ func NewCommander(logger types.Logger, cfg *config.Agent, args ...string) (*Comm
 // Start the Agent and begin watching the process.
 // Agent's stdout and stderr are written to a file.
 func (c *Commander) Start(ctx context.Context) error {
-	c.isStoppingMutex.Lock()
-	defer c.isStoppingMutex.Unlock()
+	c.startStopMutex.Lock()
+	defer c.startStopMutex.Unlock()
 
 	c.logger.Debugf(ctx, "Starting agent %s", c.cfg.Executable)
 
@@ -123,8 +123,8 @@ func (c *Commander) IsRunning() bool {
 // and if the process does not finish kills it forcedly by sending SIGKILL.
 // Returns after the process is terminated.
 func (c *Commander) Stop(ctx context.Context) error {
-	c.isStoppingMutex.Lock()
-	defer c.isStoppingMutex.Unlock()
+	c.startStopMutex.Lock()
+	defer c.startStopMutex.Unlock()
 
 	if c.cmd == nil || c.cmd.Process == nil {
 		// Not started, nothing to do.
--- a/internal/examples/supervisor/supervisor/commander/commander.go
+++ b/internal/examples/supervisor/supervisor/commander/commander.go
@@ -26,8 +26,7 @@ type Commander struct {
 	waitCh  chan struct{}
 	running int64
 
-	// True when stopping is in progress.
-	isStoppingFlag  bool
+	// Process should not be started while being stopped.
 	isStoppingMutex sync.RWMutex
 }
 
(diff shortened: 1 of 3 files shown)


internal/examples/supervisor/supervisor/commander/commander.go
```diff
@@ -116,6 +128,10 @@ func (c *Commander) IsRunning() bool {
 // and if the process does not finish kills it forcedly by sending SIGKILL.
 // Returns after the process is terminated.
 func (c *Commander) Stop(ctx context.Context) error {
+	c.isStoppingMutex.Lock()
+	c.isStoppingFlag = true
```
tigrannajaryan, line 132: We never set `isStoppingFlag` to false after this. It means `Restart()` (or `Start()` after `Stop()`) will not work correctly.

tigrannajaryan (changes requested):

internal/examples/supervisor/supervisor/commander/commander.go
tigrannajaryan: We don't have a test that verifies the scenario I describe, but it can happen in reality when the config is changed and the agent is restarted, see https://github.com/open-telemetry/opamp-go/blob/71fdce8d4b62fd3d2c3cbbc50f9765bc73a0fe15/internal/examples/supervisor/supervisor/supervisor.go#L494

tigrannajaryan: It may be worth testing the Supervisor behavior manually after this fix. The restarts happen when the Server sends a new config to the Supervisors.

rogercoll: Good catch, I end up removing the isStoppingFlag and rely on the mutex lock. https://github.com/open-telemetry/opamp-go/pull/352/commits/f71ff69b8806e1d9fde956da1d8a2d4607e1f188

This will prevent a race condition between the Start/Stop (manually tested with restarts), although I think it should be handled by the Supervisor and not the commander. The current Supervisor code can be shutdown, but not stopping the supervised processes. We will need to add a sync mechanism between the `runAgentProcess` routine and the Shutdown method.

internal/examples/supervisor/supervisor/commander/commander.go
```diff
@@ -24,6 +25,9 @@ type Commander struct {
 	doneCh  chan struct{}
 	waitCh  chan struct{}
 	running int64
+
+	// Process should not be started while being stopped.
+	isStoppingMutex sync.RWMutex
```
tigrannajaryan, line 30:
```suggestion
	startStopMutex sync.RWMutex
```

internal/examples/supervisor/supervisor/supervisor.go
```diff
@@ -331,7 +329,7 @@ service:
 // 1) the remote config from OpAMP Server, 2) the own metrics config section,
 // 3) the local override config that is hard-coded in the Supervisor.
 func (s *Supervisor) composeEffectiveConfig(ctx context.Context, config *protobufs.AgentRemoteConfig) (configChanged bool, err error) {
-	var k = koanf.New(".")
+	k := koanf.New(".")
```
tigrannajaryan, line 332: Please remove unrelated changes here and elsewhere. If we want to reformat the source it can be a separate PR, ideally supported by a linter so that we consistent in all source code.

rogercoll: sounds like a plan, removed https://github.com/open-telemetry/opamp-go/pull/352/commits/8f99ffa3a3ea13aa17e9825a8a2a24a23d5cec66

tigrannajaryan (approved):

tigrannajaryan:
Thanks for the fix @rogercoll

PR #354, opened 2025-02-21
Use optional headers from DownloadableFile messages

tpaschalis:
This PR wires in the optional headers available in the DownloadableFile message to the request, if they're present.

I think there's a couple different ways to test this, I believe that's the more straightforward one, but I can look at alternatives if you'd like to keep `createDownloadSrv` as-is.

### DIFF
--- a/client/internal/packagessyncer.go
+++ b/client/internal/packagessyncer.go
@@ -280,6 +280,12 @@ func (s *packagesSyncer) downloadFile(ctx context.Context, pkgName string, file 
 		return fmt.Errorf("cannot download file from %s: %v", file.DownloadUrl, err)
 	}
 
+	if file.Headers != nil && len(file.Headers.Headers) > 0 {
+		for _, h := range file.Headers.Headers {
+			req.Header.Add(h.GetKey(), h.GetValue())
+		}
+	}
+
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
 		return fmt.Errorf("cannot download file from %s: %v", file.DownloadUrl, err)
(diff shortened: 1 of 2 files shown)


client/internal/packagessyncer.go
```diff
@@ -280,6 +280,12 @@ func (s *packagesSyncer) downloadFile(ctx context.Context, pkgName string, file
 		return fmt.Errorf("cannot download file from %s: %v", file.DownloadUrl, err)
 	}
 
+	if file.Headers != nil && len(file.Headers.Headers) > 0 {
+		for _, h := range file.Headers.Headers {
+			req.Header.Add(h.GetKey(), h.GetValue())
```
tpaschalis, line 285: Using the `GetXXX` helpers that contain the nilness check to avoid doing it explicitly.

client/internal/packagessyncer.go
```diff
@@ -280,6 +280,12 @@ func (s *packagesSyncer) downloadFile(ctx context.Context, pkgName string, file
 		return fmt.Errorf("cannot download file from %s: %v", file.DownloadUrl, err)
 	}
 
+	if file.Headers != nil && len(file.Headers.Headers) > 0 {
```
tigrannajaryan, line 283: Nit, I think the second check is unnecessary?
```suggestion
	if file.Headers != nil {
```

tigrannajaryan (approved):

client/internal/packagessyncer.go
tpaschalis: Yeah, you're right, removed it on d88b449cb455b8541a4c17305e24e61515dbf7a7

Issue #346, opened 2025-02-10
[Feat]: websocket support does not return messages.


tttoad:
https://opentelemetry.io/docs/specs/opamp/#websocket-message-exchange
The specification does not require that we must respond to client messages in `WebSocket`, and we should support not responding to messages. 
This will reduce sending empty responses.

PR #355, opened 2025-02-25
feat: websocket: support no response.

tttoad:
see: #346.


server/serverimpl.go
```diff
@@ -264,6 +264,10 @@ func (s *server) handleWSConnection(reqCtx context.Context, wsConn *websocket.Co
 		}
 
 		response := connectionCallbacks.OnMessage(msgContext, agentConn, &request)
```
tigrannajaryan, line 266: Please also update doc-comment for OnMessage. It currently says "Must return a ServerToAgent message...".

tigrannajaryan (commented):
We also need to handle the case OnMessage() returning nil in handlePlainHTTPRequest. However, there we MUST send a response (because plain HTTP), so if nil, we must substitute it by empty message. And add a test for plain http too.

server/serverimpl.go
tttoad: Done.

server/types/callbacks.go
```diff
@@ -54,8 +54,9 @@ type ConnectionCallbacks struct {
 	OnConnected func(ctx context.Context, conn Connection)
 
 	// OnMessage is called when a message is received from the connection. Can happen
-	// only after OnConnected(). Must return a ServerToAgent message that will be sent
-	// as a response to the Agent.
+	// only after OnConnected(). 
+	// When the returned ServerToAgent message is nil, WebSocket will not send a 
+	// message to the Agent, and the HTTP request will respond to an empty message.
```
tigrannajaryan, line 59: I think we lost the bit that says if the return is not nil it will be sent as a response to the Agent. This now only talks about the nil return case.

tigrannajaryan (approved):
LGTM, just one minot nit on the doc-comment.

server/types/callbacks.go
tttoad: done.

tttoad:
@tigrannajaryan PTAL.

tigrannajaryan:
Thank you @tttoad

PR #356, opened 2025-02-26
chore: add fmt makefile target and go tools

rogercoll:
Go formatting based on Otel collector's approach: https://github.com/open-telemetry/opentelemetry-collector

The PR has been committed as:
- Add fmt go tools and refactor the corresponding make targets
- Add CI check
- Push formatted files

### DIFF
--- a/makefile
+++ b/makefile
@@ -97,16 +97,7 @@ tidy:
 	cd $(TOOLS_MOD_DIR) && rm -fr go.sum && $(GOCMD) mod tidy
 
 .PHONY: fmt
-fmt: common/gofmt common/goimports common/gofumpt
-
-.PHONY: common/gofmt
-common/gofmt:
+fmt: $(GOIMPORTS) $(GOFUMPT)
 	gofmt -w -s ./
-
-.PHONY: common/goimports
-common/goimports: $(GOIMPORTS)
-	$(GOIMPORTS) -w  -local go.opentelemetry.io/collector ./
-
-.PHONY: common/gofumpt
-common/gofumpt: $(GOFUMPT)
+	$(GOIMPORTS) -w  ./
 	$(GOFUMPT) -l -w .
--- a/makefile
+++ b/makefile
@@ -31,7 +31,7 @@ test:
 	cd internal/examples && $(GOCMD) test -race ./...
 
 .PHONY: test-with-cover
-test-with-cover:
+test-with-cover: $(GOACC)
 	$(GOACC) --output=coverage.out --ignore=protobufs ./...
 
 show-coverage: test-with-cover
@@ -74,6 +74,7 @@ gen-proto:
 
 	cp -R internal/proto/github.com/open-telemetry/opamp-go/protobufs/* protobufs/
 	rm -rf internal/proto/github.com/
+	$(MAKE) fmt
 
 .PHONY: gomoddownload
 gomoddownload:
--- a/makefile
+++ b/makefile
@@ -5,7 +5,20 @@ $(1)
 
 endef
 
-TOOLS_MOD_DIR := ./internal/tools
+GOCMD?= go
+
+# SRC_ROOT is the top of the source tree.
+SRC_ROOT := $(shell git rev-parse --show-toplevel)
+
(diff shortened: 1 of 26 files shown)


makefile
```diff
@@ -61,18 +74,39 @@ gen-proto:
...
-	cd internal/examples && rm -fr go.sum && go mod tidy
-	cd internal/tools && rm -fr go.sum && go mod tidy
+	$(GOCMD) mod tidy
+	cd internal/examples && rm -fr go.sum && $(GOCMD) mod tidy
+	cd $(TOOLS_MOD_DIR) && rm -fr go.sum && $(GOCMD) mod tidy
+
+.PHONY: fmt
+fmt: common/gofmt common/goimports common/gofumpt
+
+.PHONY: common/gofmt
```
tigrannajaryan, line 102: Unclear why this uses `common/` prefix in target names. I suggest we simplify and remove it.
```suggestion
.PHONY: gofmt
```

makefile
```diff
@@ -61,18 +74,39 @@ gen-proto:
...
+common/gofmt:
+	gofmt -w -s ./
+
+.PHONY: common/goimports
+common/goimports: $(GOIMPORTS)
+	$(GOIMPORTS) -w  -local go.opentelemetry.io/collector ./
```
tigrannajaryan, line 108: What does `-local` do and why does it refer to go.opentelemetry.io/collector?

makefile
```diff
@@ -61,18 +74,39 @@ gen-proto:
...
 	rm -fr go.sum
-	go mod tidy
-	cd internal/examples && rm -fr go.sum && go mod tidy
-	cd internal/tools && rm -fr go.sum && go mod tidy
+	$(GOCMD) mod tidy
+	cd internal/examples && rm -fr go.sum && $(GOCMD) mod tidy
+	cd $(TOOLS_MOD_DIR) && rm -fr go.sum && $(GOCMD) mod tidy
+
+.PHONY: fmt
+fmt: common/gofmt common/goimports common/gofumpt
```
tigrannajaryan, line 100: This looks unnecessarily complicated. Do we ever need to run them separately? Each subtarget runs just a single command, so perhaps simply list the 3 commands here and eliminate the subtargets:
```suggestion
fmt:
  gofmt -w -s ./
  $(GOIMPORTS) -w -local go.opentelemetry.io/collector ./
  $(GOFUMPT) -l -w .
```
These are actually supposed to be run sequentially so making them separate targets is strictly speaking incorrect (and will likely break with `make -j` when they try to concurrently modify the same file).

tigrannajaryan (commented):
Thanks a lot for working on this.

makefile
```diff
@@ -5,7 +5,20 @@ $(1)
 
 endef
 
-TOOLS_MOD_DIR := ./internal/tools
+GOCMD?= go
```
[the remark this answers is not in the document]
rogercoll: That's normally override externally.  For example, let's say you have multiple Go toolchains installed on your system:
```
GOCMD=/bin/go112 make
```

Basically, it allows to easily use another Go binary name/path.

makefile
rogercoll: Ups, good catch that's a copy/paste from the collector repo and it is used to override the `github.com/open-telemetry/opentelemetry-collector/*` imports with `go.opentelemetry.io/collector`

rogercoll: removed in https://github.com/open-telemetry/opamp-go/pull/356/commits/eed6e204d026cbba650126f92e3c0b978e5e2072

makefile
rogercoll: looks simpler, changed in https://github.com/open-telemetry/opamp-go/pull/356/commits/eed6e204d026cbba650126f92e3c0b978e5e2072

makefile
rogercoll: You are right, in most cases we end up just running the `fmt` target. Let's unify them for now https://github.com/open-telemetry/opamp-go/pull/356/commits/eed6e204d026cbba650126f92e3c0b978e5e2072

tigrannajaryan (approved):

Issue #358, opened 2025-02-27
Unnecessary error logs after calling wsConnection.Disconnect()


andykellr:
The following error messages appear when the server closes an Agent websocket connection via the `Disconnect()` method.

```
Cannot read a message from WebSocket: read tcp 127.0.0.1:3001->127.0.0.1:49552: use of closed network connection
```

```
error closing the WebSocket connection: close tcp 127.0.0.1:3001->127.0.0.1:49552: use of closed network connection
```

The issue is in `server.handleWSConnection` when `ReadMessage()` returns an error and then the `defer` tries to `Close()` the already closed websocket connection.

PR #359, opened 2025-03-04
fix: Add atomic.Bool to ensure wsConnection.Disconnect() is only called once

shazlehu:
Addresses this [issue](https://github.com/open-telemetry/opamp-go/issues/358) by addding an atomic.Bool so we only close the underlying websocket connection once. This PR also adds the test `TestDisconnectServerWSConnection` to ensure that that error log isn't generated when the connection is closed server-side. I renamed `TestDisconnectWSConnection` to `TestDisconnectClientWSConnection` since it tests closing the connection from the client-side, though the comments in the test implied it tested disconnecting from the server.

Perhaps it would be better to simply ignore the error when we close the connection twice? I'm unsure if there are error conditions where we'd want to be able to try to close the connection again.

Resolves https://github.com/open-telemetry/opamp-go/issues/358

### DIFF
--- a/server/serverimpl_test.go
+++ b/server/serverimpl_test.go
@@ -30,7 +30,6 @@ import (
 
 func startServer(t *testing.T, settings *StartSettings) *server {
 	return startServerWithLogger(t, settings, &sharedinternal.NopLogger{})
-
 }
 
 func startServerWithLogger(t *testing.T, settings *StartSettings, logger clienttypes.Logger) *server {
--- a/server/serverimpl_test.go
+++ b/server/serverimpl_test.go
@@ -29,19 +29,8 @@ import (
 )
 
 func startServer(t *testing.T, settings *StartSettings) *server {
-	srv := New(&sharedinternal.NopLogger{})
-	require.NotNil(t, srv)
-	if settings.ListenEndpoint == "" {
-		// Find an avaiable port to listen on.
-		settings.ListenEndpoint = testhelpers.GetAvailableLocalAddress()
-	}
-	if settings.ListenPath == "" {
-		settings.ListenPath = "/"
-	}
-	err := srv.Start(*settings)
-	require.NoError(t, err)
+	return startServerWithLogger(t, settings, &sharedinternal.NopLogger{})
 
-	return srv
 }
 
 func startServerWithLogger(t *testing.T, settings *StartSettings, logger clienttypes.Logger) *server {
@@ -321,14 +310,13 @@ func (o *testLogger) Errorf(_ context.Context, format string, v ...any) {
 func TestDisconnectServerWSConnection(t *testing.T) {
 	connectionCloseCalled := int32(0)
 	var serverConn types.Connection
-	mutex := sync.Mutex{}
+	connReady := make(chan struct{}) // Channel to signal when serverConn is assigned
 	callback := types.Callbacks{
 		OnConnecting: func(request *http.Request) types.ConnectionResponse {
 			return types.ConnectionResponse{Accept: true, ConnectionCallbacks: types.ConnectionCallbacks{
 				OnConnected: func(ctx context.Context, conn types.Connection) {
-					mutex.Lock()
 					serverConn = conn
-					mutex.Unlock()
+					close(connReady)
 				},
 				OnConnectionClose: func(conn types.Connection) {
 					atomic.StoreInt32(&connectionCloseCalled, 1)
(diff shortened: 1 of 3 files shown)


server/wsconnection.go
```diff
@@ -19,6 +20,7 @@ type wsConnection struct {
 	// For more: https://pkg.go.dev/github.com/gorilla/websocket#hdr-Concurrency
 	connMutex *sync.Mutex
 	wsConn    *websocket.Conn
+	closed    *atomic.Bool
```
tigrannajaryan, line 23: Does this need to be a pointer? Seems to be like unnecessary extra allocation.

server/wsconnection.go
```diff
@@ -35,5 +37,9 @@ func (c wsConnection) Send(_ context.Context, message *protobufs.ServerToAgent)
 }
 
 func (c wsConnection) Disconnect() error {
+	if c.closed.Load() {
+		return nil
+	}
+	c.closed.Store(true)
```
tigrannajaryan, line 43: To ensure this happens once, I think the correct way is to use the atomic `CompareAndSwap` instead of separate Load/Store which can race.

server/serverimpl_test.go
```diff
@@ -263,10 +281,77 @@ func TestDisconnectWSConnection(t *testing.T) {
...
+
+	// Connect to the Server.
+	conn, _, err := dialClient(settings)
+
+	// Verify that the connection is successful.
 	assert.NoError(t, err)
+	assert.True(t, atomic.LoadInt32(&connectionCloseCalled) == 0)
+
+	// Close connection from server side
+	mutex.Lock()
```
tigrannajaryan, line 352: What does this mutex protect against? Does `serverConn` value ever change (except from `nil` to the only connection we get)? Can `serverConn` be nil here?

server/serverimpl_test.go
```diff
@@ -275,6 +360,10 @@ func TestDisconnectWSConnection(t *testing.T) {
 		_, _, err := conn.ReadMessage()
 		return err != nil
 	})
+
+	require.Equal(t, 1, len(logger.logs))
+	require.Contains(t, logger.logs[0], "Errorf: Cannot read a message from WebSocket: read tcp")
+	require.Contains(t, logger.logs[0], "use of closed network connection")
```
tigrannajaryan, line 366: Do we care about the log messages produced? It looks fragile to make the tests dependent on the exact formatting of the messages.

tigrannajaryan:
@shazlehu  Is this still in progress?

shazlehu:
@tigrannajaryan I'll make a push to get this over the finish line in the next week

server/wsconnection.go
shazlehu: Good call

server/serverimpl_test.go
shazlehu: Agree we should avoid fragility. The original motivation for this change was to deal with superfluous error logs produced by the server-side Disconnect() calls. Open to suggestions for a better criteria for to test for.

server/wsconnection.go
tigrannajaryan: I think it is fine to change the receiver types (unless I am missing something). That should avoid extra allocation on every connection.

server/wsconnection.go
```diff
@@ -17,23 +18,31 @@ type wsConnection struct {
...
+	connMutex sync.Mutex
 	wsConn    *websocket.Conn
+	closed    atomic.Bool
 }
 
 var _ types.Connection = (*wsConnection)(nil)
 
-func (c wsConnection) Connection() net.Conn {
+func newWSConnection(wsConn *websocket.Conn) types.Connection {
+	return &wsConnection{wsConn: wsConn, connMutex: sync.Mutex{}, closed: atomic.Bool{}}
```
tigrannajaryan: I think zero-initialization for connMutex and closed is fine, e.g. this should work the same way and is less verbose:
```suggestion
	return &wsConnection{wsConn: wsConn}
```

server/serverimpl_test.go
tigrannajaryan: Perhaps as a compromise: in testLogger store error logs separate from debug logs and here only assert that there is an error log produced, but don't check the error message itself. This should be fairly stable since we always expect exactly one error log and don't really care about the message itself.

shazlehu: Sounds like a good compromise. All set.

shazlehu:
I think everything is set on this? Anything else needed @tigrannajaryan?

tigrannajaryan (approved):

tigrannajaryan:
Thank you @shazlehu

PR #366, opened 2025-03-24
Make it possible to set capabilities

### LINKED ISSUE #365: Client should be able to change capabilities while running
Right now the nextMessage sends capabilities on every message, but there's no way to actually change them after start. It would be great if the client could actually change them.

jaronoff97:
Closes #365

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -152,4 +152,17 @@ type OpAMPClient interface {
 	// This method is subject to agent status compression - if components is not
 	// different from the cached agent state, this method is a no-op.
 	SetAvailableComponents(components *protobufs.AvailableComponents) error
+
+	// SetCapabilities updates the set of capabilities that the client supports.
+	// These capabilities will be communicated to the server in the next message.
+	//
+	// This method can be called at any time after Start(), including from within
+	// an OnMessage handler, to dynamically update the set of supported capabilities.
+	// The updated capabilities will be sent to the server in the next outgoing message.
+	//
+	// The capabilities parameter must not be nil; passing a nil value will result in an error.
+	//
+	// For more details, refer to the OpAMP specification:
+	// https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agenttoservercapabilities
+	SetCapabilities(capabilities *protobufs.AgentCapabilities) error
 }
--- a/client/clientimpl_test.go
+++ b/client/clientimpl_test.go
@@ -1738,6 +1738,72 @@ func TestOfferUpdatedVersion(t *testing.T) {
 	})
 }
 
+func TestSetCapabilities(t *testing.T) {
+	testClients(t, func(t *testing.T, client OpAMPClient) {
+		// Start a Server.
+		srv := internal.StartMockServer(t)
+		srv.EnableExpectMode()
+
+		var clientRcvCustomMessage atomic.Value
+
+		// Start a client.
+		settings := types.StartSettings{
+			OpAMPServerURL: "ws://" + srv.Endpoint,
+			Callbacks: types.Callbacks{
+				OnMessage: func(ctx context.Context, msg *types.MessageData) {
+					clientRcvCustomMessage.Store(msg.CustomMessage)
+				},
+			},
+			Capabilities: protobufs.AgentCapabilities_AgentCapabilities_ReportsEffectiveConfig,
+		}
+		prepareClient(t, &settings, client)
+
+		// Client --->
+		assert.NoError(t, client.Start(context.Background(), settings))
+
+		// ---> Server
+		srv.Expect(func(msg *protobufs.AgentToServer) *protobufs.ServerToAgent {
+			assert.EqualValues(t, 0, msg.SequenceNum)
+			// The first status report after Start must have the ReportsStatus.
+			assert.True(t, protobufs.AgentCapabilities(msg.Capabilities)&protobufs.AgentCapabilities_AgentCapabilities_ReportsStatus != 0)
+			// The first status report after Start must have the ReportsEffectiveConfig.
+			assert.True(t, protobufs.AgentCapabilities(msg.Capabilities)&protobufs.AgentCapabilities_AgentCapabilities_ReportsEffectiveConfig != 0)
+			// The first status report after Start must not  have the AcceptsRemoteConfig.
+			assert.True(t, protobufs.AgentCapabilities(msg.Capabilities)&protobufs.AgentCapabilities_AgentCapabilities_AcceptsRemoteConfig == 0)
+			return &protobufs.ServerToAgent{
+				InstanceUid: msg.InstanceUid,
+			}
+		})
+
+		newCapabilities := protobufs.AgentCapabilities_AgentCapabilities_AcceptsRemoteConfig
+		err := client.SetCapabilities(&newCapabilities)
+		assert.NoError(t, err)
+
+		// ---> Server
+		srv.Expect(func(msg *protobufs.AgentToServer) *protobufs.ServerToAgent {
+			// Check ReportsStatus is still true.
+			assert.True(t, protobufs.AgentCapabilities(msg.Capabilities)&protobufs.AgentCapabilities_AgentCapabilities_ReportsStatus != 0)
+			// ReportsEffectiveConfig should no longer be present.
+			assert.True(t, protobufs.AgentCapabilities(msg.Capabilities)&protobufs.AgentCapabilities_AgentCapabilities_ReportsEffectiveConfig == 0)
+			// AcceptsRemoteConfig should now be present
+			assert.True(t, protobufs.AgentCapabilities(msg.Capabilities)&protobufs.AgentCapabilities_AgentCapabilities_AcceptsRemoteConfig != 0)
+
+			// Send a custom message response and ask client for full state again.
+			return &protobufs.ServerToAgent{
+				InstanceUid: msg.InstanceUid,
+				Flags:       uint64(protobufs.ServerToAgentFlags_ServerToAgentFlags_ReportFullState),
+			}
+		})
+
+		// Shutdown the Server.
+		srv.Close()
+
+		// Shutdown the client.
+		err = client.Stop(context.Background())
+		assert.NoError(t, err)
+	})
+}
+
 func TestReportCustomCapabilities(t *testing.T) {
 	testClients(t, func(t *testing.T, client OpAMPClient) {
 		// Start a Server.
--- a/client/httpclient.go
+++ b/client/httpclient.go
@@ -10,6 +10,8 @@ import (
 	"github.com/open-telemetry/opamp-go/protobufs"
 )
 
+var _ OpAMPClient = &httpClient{}
+
 // httpClient is an OpAMP Client implementation for plain HTTP transport.
 // See specification: https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#plain-http-transport
 type httpClient struct {
@@ -125,6 +127,11 @@ func (c *httpClient) SetAvailableComponents(components *protobufs.AvailableCompo
 	return c.common.SetAvailableComponents(components)
 }
 
+// SetCapabilities implements OpAMPClient.
+func (c *httpClient) SetCapabilities(capabilities *protobufs.AgentCapabilities) error {
+	return c.common.SetCapabilities(capabilities)
+}
+
 func (c *httpClient) runUntilStopped(ctx context.Context) {
 	// Start the HTTP sender. This will make request/responses with retries for
 	// failures and will wait with configured polling interval if there is nothing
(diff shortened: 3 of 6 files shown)


client/wsclient.go
```diff
@@ -22,6 +22,8 @@ const (
 	defaultShutdownTimeout = 5 * time.Second
 )
 
+var _ OpAMPClient = &wsClient{}
```
andykellr, line 25:
```suggestion
var _ OpAMPClient = (*wsClient)(nil)
```

andykellr (approved):
LGTM

tigrannajaryan:
The spec does not say whether this is a support operation (to change capabilities after they are first reported). Let me think about it.

tigrannajaryan (dismissed):
Blocking temporarily, give me a bit time to think about the implications of this.

andykellr:
> The spec does not say whether this is a support operation (to change capabilities after they are first reported). Let me think about it.

Sounds good. I held off on merging until you had a chance to look at it.

My reading of the spec was that since it was not specifically mentioned or prohibited that it would be ok to allow it. Some servers may expect this to stay the same for the life of the connection but since it is passed with every message I think the server should be able to adjust as needed.

tigrannajaryan:
I don't think we can make this blanket change.

There are capabilities which are checked at the Start() and the corresponding invariants for the capabilities are checked at the same time.

For example, in PrepareStart() we verify that if AcceptsPackages is set then PackagesStateProvider is also provided. Later in receivedProcessor we rely on this invariant. If we break the invariant (which you can easily do via SetCapabilities) then receivedProcessor will attempt to use a nil PackagesStateProvider (which will either crash or error, I didn't look further).

At least from implementation perspective this is not a change we can make.

I am also not sure conceptually it works for every single capability, to allow changing them on the fly.

Before we make this blanket change I would like to see the analysis which explain why it is OK to change each particular capability on the fly (I am not sure that is true). Then we will need to make sure the implementation is ready for that (it is currently not).

As an alternate, is there a particular capability that you need to change after Start()? we can look at supporting just that.

jaronoff97:
Yes, i did notice that. I think we could do instantiation on demand though based on the capabilities being set. Initially, I just need to be able to enable `AcceptsRemoteConfig`, and I'd be okay limiting the `SetCapabilities` to only allow flipping that on and off initially, adding in more allowed changes in the future?

tigrannajaryan:
I think AcceptsRemoteConfig is doable. I don't see any special invariants for this capability in the codebase.

We can do this:

1. Modify OpAMP spec to explicitly call out that the agent may send a different set of capabilities in subsequent AgentToServer messages after the initial one. Explain that only AcceptsRemoteConfig can change.
2. Restrict SetCapabilities to only allow certain capabilities to be changed after Start. For now it will only allow AcceptsRemoteConfig. We will need to vet all new capabilities one by one if we want to allow changing them.

client/client.go
```diff
@@ -152,4 +152,17 @@ type OpAMPClient interface {
...
+	//
+	// This method can be called at any time after Start(), including from within
+	// an OnMessage handler, to dynamically update the set of supported capabilities.
+	// The updated capabilities will be sent to the server in the next outgoing message.
+	//
+	// The capabilities parameter must not be nil; passing a nil value will result in an error.
+	//
+	// For more details, refer to the OpAMP specification:
+	// https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agenttoservercapabilities
+	SetCapabilities(capabilities *protobufs.AgentCapabilities) error
```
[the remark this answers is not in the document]
tigrannajaryan: @jaronoff97 I am not sure if you updated the PR to make this change. Is it ready for another review round?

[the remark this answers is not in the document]
jaronoff97: i was waiting on andy's review and when you said "we decided to go ahead with this feature" I thought it was dismissing this comment. I see now, i'll update the PR based on what you said above in this thread.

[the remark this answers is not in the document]
tigrannajaryan: Yeah, no problem, I just wasn't sure if I need to take another look.
I would like @andykellr's opinion too.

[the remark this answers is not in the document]
tigrannajaryan: @jaronoff97 in case I wasn't clear, I am waiting for you to tell me if this is ready for another round of review.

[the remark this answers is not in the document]
jaronoff97: i was waiting on andy's opinion here, but either way i can make your requested changes from above before its ready for more review. apologies.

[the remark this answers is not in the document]
tigrannajaryan: @jaronoff97 sounds good, thanks. Ping me when you update the PR, I will take a look.

client/internal/clientcommon.go
```diff
@@ -97,12 +98,17 @@ func (c *ClientCommon) PrepareStart(
...
 		return errAlreadyStarted
 	}
-	capabilities := settings.Capabilities
-
-	// According to OpAMP spec this capability MUST be set, since all Agents MUST report status.
-	capabilities |= protobufs.AgentCapabilities_AgentCapabilities_ReportsStatus
-	if err := c.ClientSyncedState.SetCapabilities(&capabilities); err != nil {
-		return err
+	// Deprecated: Use client.SetCapabilities() instead.
+	if settings.Capabilities != 0 {
```
jaronoff97, line 102: @tigrannajaryan this _technically_ could be a breaking change if someone isn't setting any capabilities, should I just default to setting reports status here and then in the future we remove this?

client/internal/clientcommon.go
```diff
@@ -492,3 +504,36 @@ func (c *ClientCommon) SetAvailableComponents(components *protobufs.AvailableCom
...
+			msg.Capabilities = uint64(c.ClientSyncedState.Capabilities())
+		},
+	)
+	c.sender.ScheduleSend()
+	return nil
+}
+
+// QUESTION: DO we want this?
+// SetPackagesStateProvider allows the confgiguration of the packages state provider after start.
+// func (c *ClientCommon) SetPackagesStateProvider(packagesStateProvider types.PackagesStateProvider) error {
```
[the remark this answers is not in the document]
tigrannajaryan: I am not sure I understand. Why would the client need to set the state provider after the Start? It can be set before Start, but can return a different set of packages anytime.

client/internal/clientcommon.go
tigrannajaryan: I think we should emit a warning in the log if SetCapabilities is not called before Start. Later, we can make it an error. Then finally some time after that we remove Settings.Capabilities field.

In any of these 3 states, ReportsStatus should always be automatically added, regardless of how the Capabilities are set (via Settings or via method call).

client/client.go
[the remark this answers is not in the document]
andykellr: This approach is fine with me.

andykellr (approved):

client/internal/clientcommon.go
jaronoff97: @tigrannajaryan i do make sure reportsstatus is always set (I do the add in this method as well in the clientcommon setcapabilities), but I'll add in a log right now.

jaronoff97: @tigrannajaryan should be all good now. PTAL 🙇

client/httpclient.go
```diff
@@ -125,6 +127,11 @@ func (c *httpClient) SetAvailableComponents(components *protobufs.AvailableCompo
 	return c.common.SetAvailableComponents(components)
 }
 
+// SetCapabilities implements OpAMPClient.
```
tigrannajaryan: Comment seems wrong.

client/internal/clientcommon.go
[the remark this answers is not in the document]
tigrannajaryan: I am still not sure about this part. If this is not needed please delete the commented code.

client/client.go
```diff
@@ -152,4 +152,17 @@ type OpAMPClient interface {
 	// This method is subject to agent status compression - if components is not
 	// different from the cached agent state, this method is a no-op.
 	SetAvailableComponents(components *protobufs.AvailableComponents) error
+
+	// SetCapabilities updates the set of capabilities that the client supports.
+	// These capabilities will be communicated to the server in the next message.
+	//
+	// This method can be called at any time after Start(), including from within
```
tigrannajaryan: This can be called before Start() too, right?

client/clientimpl_test.go
```diff
@@ -218,13 +226,27 @@ func TestStopCancellation(t *testing.T) {
...
 }
 
+func TestStartNoCapabilities(t *testing.T) {
+	testClients(t, func(t *testing.T, client OpAMPClient) {
+		setErr := client.SetAgentDescription(createAgentDescr())
+		require.NoError(t, setErr)
+		settings := createNoServerSettings()
+		prepareSettings(t, &settings, client)
+		err := client.Start(context.Background(), settings)
+		// assert.EqualValues(t, internal.ErrCapabilitiesNotSet, err)
```
tigrannajaryan: Remove commented line.

client/clientimpl_test.go
```diff
@@ -2407,6 +2521,119 @@ func TestSetAvailableComponents(t *testing.T) {
...
+		},
+		{
+			name:         "AcceptsPackages capability without PackagesStateProvider",
+			capabilities: protobufs.AgentCapabilities_AgentCapabilities_AcceptsPackages,
+			setupFunc: func(t *testing.T, client OpAMPClient) {
+				// Do not set PackagesStateProvider
+			},
+			expectedError: internal.ErrPackagesStateProviderNotSet,
+		},
+		// {
```
tigrannajaryan: Please remove commented code from everywhere (or make these case work if they are needed).

client/internal/clientcommon.go
```diff
@@ -67,30 +65,67 @@ func NewClientCommon(logger types.Logger, sender Sender) ClientCommon {
...
+			return err
+		}
+	}
+	if c.ClientSyncedState.Capabilities() == 0 {
+		c.Logger.Errorf(ctx, "you must call client.SetCapabilities before start.")
+		capabilities := protobufs.AgentCapabilities_AgentCapabilities_ReportsStatus
+		if err := c.ClientSyncedState.SetCapabilities(&capabilities); err != nil {
+			return err
+		}
+		// Eventually we will error here.
```
tigrannajaryan: What does "Eventually" mean? Do we need this to be an error?

jaronoff97: this was in reference to your comment [here](https://github.com/open-telemetry/opamp-go/pull/366#discussion_r2120874171). I can open an issue and just reference that instead.

tigrannajaryan:
@jaronoff97 I did a review round a couple weeks ago in case you missed it.

jaronoff97:
oh 🤦 i did miss it, i'll update this! thanks for the bump.

client/internal/clientcommon.go
```diff
@@ -72,30 +70,68 @@ func NewClientCommon(logger types.Logger, sender Sender) ClientCommon {
...
-	// According to OpAMP spec this capability MUST be set, since all Agents MUST report status.
-	c.Capabilities |= protobufs.AgentCapabilities_AgentCapabilities_ReportsStatus
+	// Deprecated: Use client.SetCapabilities() instead.
+	if settings.Capabilities != 0 {
+		c.Logger.Errorf(ctx, "settings.Capabilities is deprecated, use client.SetCapabilities() instead")
+		capabilities := settings.Capabilities
+		// According to OpAMP spec this capability MUST be set, since all Agents MUST report status.
+		capabilities |= protobufs.AgentCapabilities_AgentCapabilities_ReportsStatus
+		if err := c.ClientSyncedState.SetCapabilities(&capabilities); err != nil {
+			return err
```
jaronoff97, line 113: @tigrannajaryan logically, this state can never happen but codecov isn't smart enough to realize that... the only way setcapabilities could error is if the capabilities passed in here is nil, but capabilities is a reference to an instantiated variable in both this case and the next case which means this could never happen. How should I proceed here?

tigrannajaryan: That's fine, ignore codecov.

client/internal/clientcommon.go
```diff
@@ -72,30 +70,68 @@ func NewClientCommon(logger types.Logger, sender Sender) ClientCommon {
...
+		}
+	}
+	if c.ClientSyncedState.Capabilities() == 0 {
+		c.Logger.Errorf(ctx, "you must call client.SetCapabilities before start.")
+		capabilities := protobufs.AgentCapabilities_AgentCapabilities_ReportsStatus
+		if err := c.ClientSyncedState.SetCapabilities(&capabilities); err != nil {
+			return err
+		}
+		// Error here.
```
tigrannajaryan:
```suggestion
		// We allow this to succeed for now, but later this will become an error.
```

tigrannajaryan (commented):
Please update the PR description to explain the changes so that the CHANGELOG in the next release is clear about what's being changed/deprecated.

tigrannajaryan:
Merged, thanks for patience @jaronoff97 !

PR #380, opened 2025-05-06
fix: close connections & wait goroutine when server shutdown

### LINKED ISSUE #271: [server] Method of handling websocket connections leaks goroutines
**Context**

Hello, this was found as a result of investigation for https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/30931. Enabling [`goleak`](https://github.com/uber-go/goleak) in the `server` directory of `opamp-go` shows the `TestServerCallsHTTPMiddlewareOverWebsocket` test is leaking. Here's the call stack for the `goleak` failure:

**Stack trace**
```
=== RUN   TestServerCallsHTTPMiddlewareOverWebsocket
--- PASS: TestServerCallsHTTPMiddlewareOverWebsocket (0.01s)
PASS
goleak: Errors on successful test run: found unexpected goroutines:
[Goroutine 23 in state IO wait, with internal/poll.runtime_pollWait on top of the stack:
internal/poll.runtime_pollWait(0x4888de10, 0x72)
	/Users/crobert/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.0.darwin-amd64/src/runtime/netpoll.go:345 +0x85
internal/poll.(*pollDesc).wait(0xc0001e2780?, 0xc000236000?, 0x0)
	/Users/crobert/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.0.darwin-amd64/src/internal/poll/fd_poll_runtime.go:84 +0x27
internal/poll.(*pollDesc).waitRead(...)
	/Users/crobert/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.0.darwin-amd64/src/internal/poll/fd_poll_runtime.go:89
internal/poll.(*FD).Read(0xc0001e2780, {0xc000236000, 0x1000, 0x1000})
	/Users/crobert/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.0.darwin-amd64/src/internal/poll/fd_unix.go:164 +0x27a
net.(*netFD).Read(0xc0001e2780, {0xc000236000?, 0x0?, 0x0?})
	/Users/crobert/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.0.darwin-amd64/src/net/fd_posix.go:55 +0x25
net.(*conn).Read(0xc0001940c0, {0xc000236000?, 0x10?, 0x10?})
	/Users/crobert/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.0.darwin-amd64/src/net/net.go:179 +0x45
bufio.(*Reader).fill(0xc0001a8300)
	/Users/crobert/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.0.darwin-amd64/src/bufio/bufio.go:110 +0x103
bufio.(*Reader).Peek(0xc0001a8300, 0x2)
	/Users/crobert/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.0.darwin-amd64/src/bufio/bufio.go:148 +0x53
github.com/gorilla/websocket.(*Conn).read(0xc0001dc2c0, 0x0?)
	/Users/crobert/go/pkg/mod/github.com/gorilla/websocket@v1.5.1/conn.go:378 +0x26
github.com/gorilla/websocket.(*Conn).advanceFrame(0xc0001dc2c0)
	/Users/crobert/go/pkg/mod/github.com/gorilla/websocket@v1.5.1/conn.go:824 +0x6d
github.com/gorilla/websocket.(*Conn).NextReader(0xc0001dc2c0)
	/Users/crobert/go/pkg/mod/github.com/gorilla/websocket@v1.5.1/conn.go:1034 +0x13e
github.com/gorilla/websocket.(*Conn).ReadMessage(0x193d0a0?)
	/Users/crobert/go/pkg/mod/github.com/gorilla/websocket@v1.5.1/conn.go:1120 +0x13
github.com/open-telemetry/opamp-go/server.(*server).handleWSConnection(0xc0001f00c0, {0x197e350, 0xc00028e0a0}, 0xc0001dc2c0, {0x197d918, 0xc0002b2018})
	/Users/crobert/dev/opamp-go/server/serverimpl.go:240 +0x1a5
created by github.com/open-telemetry/opamp-go/server.(*server).httpHandler in goroutine 22
	/Users/crobert/dev/opamp-go/server/serverimpl.go:210 +0x350
]
```
This shows the test itself is passing, but `goleak` detects leaked goroutines.

**Analysis**

I found the culprit to be this section:
https://github.com/open-telemetry/opamp-go/blob/c7fc5850ffdf65e97a39b080d73f1f23346bf52a/server/serverimpl.go#L233-L234
As shown by the comment, this line will block until the connection can read a message or until it's closed. However, the close call is in the same goroutine, within a `defer` statement. This means the websocket will never be closed, and the goroutine is leaked.

The solution would be to have the `wsConn.Close` method be called when the `reqCtx` is cancelled, from what I can tell. This will allow the read call to stop blocking and properly finish running. Note that I don't have much opamp experience, so the end solution may be quite different than my suggestion.

minuk-dev:
fix #271 

- Add `serverCtx` and `serverCancel` to propagate server's shutdown event to each connection.
- Replace `context.Background` to `serverCtx` in `handleWSConnection()`'s websocket loop.
- Add `connWg` to wait all connections' goroutine.
- Make a goroutine for `wsConn.ReadMessage()` because the method does not support context.
- Add `TestServerStartStopCloseNoLeakGoroutine` test:
  - Please comment if you have any better idea to test this feature instead of `goleak`. 🙇


server/serverimpl.go
```diff
@@ -45,6 +45,11 @@ type server struct {
 
 	// The network address Server is listening on. Nil if not started.
 	addr net.Addr
+
+	// To gracefully stop the server.
+	serverCtx    context.Context
```
tigrannajaryan, line 50: http.Server has a BaseContext which if I remember correctly can be used for cancellations (I may be wrong on this). I think BaseContext is passed as the request context and if that is the case we can use it for cancellation in our handlers. 
Please check that approach, since that may be more idiomatic usage of http.Sever.

server/serverimpl.go
```diff
@@ -45,6 +45,11 @@ type server struct {
...
+	serverCancel context.CancelFunc
+	connWg       *sync.WaitGroup
```
tigrannajaryan, line 52: Do we need this new WaitGroup? I think `httpServerServeWg` should be ready after all connections are closed (unless Serve() doesn't wait for connections to close - I haven't checked).

server/serverimpl.go
```diff
@@ -243,13 +259,40 @@ func (s *server) handleWSConnection(reqCtx context.Context, wsConn *websocket.Co
...
+
 	// Loop until fail to read from the WebSocket connection.
 	for {
-		msgContext := context.Background()
+		msgContext := s.serverCtx
 		request := protobufs.AgentToServer{}
 
-		// Block until the next message can be read.
-		mt, msgBytes, err := wsConn.ReadMessage()
+		go func() {
```
tigrannajaryan, line 274: Creating a goroutine for each message seems excessive, this may have a significant cost for tiny messages (benchmark?). Alternatively, can we create one goroutine per connection and keep it running until the connection is closed?

server/serverimpl.go
minuk-dev: > Please check that approach, since that may be more idiomatic usage of http.Sever.

I really agree with you. I also want to implement it more idiomatically.
But, I think it's hard. Please see the below reasons.

> http.Server has a BaseContext which if I remember correctly can be used for cancellations (I may be wrong on this)

websocket library has an issue.
We should manually call `Close()`. https://github.com/gorilla/websocket/issues/448

The BaseContext is not conveyed to `websocket.Conn`.
(And I think the `BaseContext` is not also set to `net.Conn`. I cannot find the spec. So I checked [its code](https://cs.opensource.google/go/go/+/refs/tags/go1.24.3:src/net/http/server.go;l=3424))

---

> I think BaseContext is passed as the request context and if that is the case we can use it for cancellation in our handlers.

The request context is canceled while the connection is upgraded to websocket even if the connection is hijacked. So we cannot use the request's context in this case.

- ref. https://pkg.go.dev/net/http#Request.Context
> For incoming server requests, the context is canceled when the client's connection closes, the request is canceled (with HTTP/2), or when the ServeHTTP method returns.

<details> <summary> code level </summary>


- https://cs.opensource.google/go/go/+/refs/tags/go1.22.9:src/net/http/server.go;l=2046 
```go
// src/net/http/server.go
func (c *conn) readRequest(ctx context.Context) (w *response, err error) {
  // skip
  ctx, cancelCtx := context.WithCancel(ctx)
  req.ctx = ctx
  // skip
  w = &response{
     cancelCtx: cancelCtx,
  }
}

// skip
func (c *conn) serve(ctx context.Context) {
  // skip
  serverHandler{c.server}.ServerHTTP(w, w.req)
  inFlightResponse = nil
  w.cancelCtx()
  if c.hijacked() {
    return
  }
  // skip
}
```

server/serverimpl.go
minuk-dev: `httpServerServeWg` is to handle `httpServer.Serve`'s lifecycle.
But, we need to call all connection's `Close()` function to gracefully shutdown.
So, I think the purpose is different and cannot be shared.

tigrannajaryan:
@minuk-dev is this still in progress?

minuk-dev:
> @minuk-dev is this still in progress?

I forgot it. I'll fix the review of goroutine. Is there any other things should I apply?

server/serverimpl.go
minuk-dev: fixed 
https://github.com/open-telemetry/opamp-go/pull/380/commits/1e9f46a5a12808489b41f66dffcc97b8edc80e3f
https://github.com/open-telemetry/opamp-go/pull/380/commits/787910f6d3bf6a5ce56ad9032ed0899ba0ce5f10

server/serverimpl_test.go
```diff
@@ -72,6 +73,15 @@ func TestServerStartStopWithCancel(t *testing.T) {
 	assert.ErrorIs(t, err, context.Canceled)
 }
 
+func TestServerStartStopCloseNoLeakGoroutine(t *testing.T) {
+	goleak.VerifyNone(t)
```
michel-laterman:
```suggestion
	defer goleak.VerifyNone(t)
```
https://pkg.go.dev/go.uber.org/goleak#VerifyNone

server/serverimpl.go
michel-laterman: My understanding is that @minuk-dev is correct, once we upgrade/hijack the connection, the `req.Context` will not affect it.

minuk-dev:
I'm sorry for repeated Ci failure. I don't have enough time to check this PR, now. I'll re-open it in future. Sorry.

tigrannajaryan:
@minuk-dev please resolve the conflicts.

server/serverimpl.go
```diff
@@ -246,14 +263,36 @@ func (s *server) handleWSConnection(reqCtx context.Context, wsConn *websocket.Co
...
+		var mt int
+		var msgBytes []byte
+		var err error
+
+		// Wait for the message to be read or for the server to shut down.
+		select {
+		case <-serverCtx.Done():
+			s.logger.Debugf(msgContext, "Server is shutting down.: %v", s.serverCtx.Err())
+			return
+		case msg, ok := <-msgCh:
```
tigrannajaryan, line 289: @andykellr can you take a look at this? I wonder what the performance overhead is for communicating all messages via a channel. Do you anticipate a significant overhead at large scale / large number of connection and messages?

server/serverimpl.go
```diff
@@ -246,14 +263,36 @@ func (s *server) handleWSConnection(reqCtx context.Context, wsConn *websocket.Co
...
-		mt, msgBytes, err := wsConn.ReadMessage()
+		var mt int
+		var msgBytes []byte
+		var err error
+
+		// Wait for the message to be read or for the server to shut down.
+		select {
+		case <-serverCtx.Done():
+			s.logger.Debugf(msgContext, "Server is shutting down.: %v", s.serverCtx.Err())
+			return
```
michel-laterman, line 288: Can we send a control frame to close the connection here? Either 1000 (normal closure), or 1001 (going away)

minuk-dev:
I forgot this PR. I'll resolve conflicts and apply some reviews.

PR #383, opened 2025-05-08
feat: Define connectionType for server

minuk-dev:
- In the opamp-spec document, the connection type is http or websocket.
  - ref. https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#communication-model

> OpAMP protocol works over one of the 2 supported transports: plain HTTP connections and WebSocket connections.

- To define OpAMP server, a developer should be define `ConnectionCallbacks` and manage connections using `OnConnected` and `OnConnectionClose`. However, a way to handle the two type of connections are different.:
  - For example, http connection cannot send message, so if the server want to send `ServerToAgent` message, wait for the its next connection. (= next polling)

- So, IMO, the `Connection` provides a method to find connection type.

### DIFF
--- a/server/httpconnection.go
+++ b/server/httpconnection.go
@@ -35,6 +35,10 @@ func (c httpConnection) Send(_ context.Context, _ *protobufs.ServerToAgent) erro
 	return ErrInvalidHTTPConnection
 }
 
+func (c httpConnection) Type() types.ConnectionType {
+	return types.ConnectionTypeHTTP
+}
+
 func (c httpConnection) Disconnect() error {
 	// Disconnect() should not be called for plain HTTP connection.
 	return ErrInvalidHTTPConnection
--- a/server/types/connection.go
+++ b/server/types/connection.go
@@ -7,11 +7,26 @@ import (
 	"github.com/open-telemetry/opamp-go/protobufs"
 )
 
+const (
+	// ConnectionTypeHTTP is the type of a plain HTTP connection.
+	ConnectionTypeHTTP ConnectionType = "http"
+	// ConnectionTypeWebSocket is the type of a WebSocket connection.
+	ConnectionTypeWebSocket ConnectionType = "websocket"
+)
+
+// ConnectionType is the type of a connection.
+// It can be one of ConnectionTypeHTTP or ConnectionTypeWebSocket.
+type ConnectionType string
+
 // Connection represents one OpAMP connection.
 // The implementation MUST be a comparable type so that it can be used as a map key.
 type Connection interface {
 	// Connection returns the underlying net.Conn
 	Connection() net.Conn
+
+	// Type returns the type of the connection.
+	// The type is one of http, websocket, or unknown.
+	Type() ConnectionType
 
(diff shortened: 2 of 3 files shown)


tigrannajaryan:
Unclear why this is necessary. The callbacks are intended to abstract away the differences between transports. Can you provide a specific code example where knowing the transport type is necessary and cannot be avoided?

minuk-dev:
> Unclear why this is necessary. The callbacks are intended to abstract away the differences between transports. Can you provide a specific code example where knowing the transport type is necessary and cannot be avoided?

Honestly, because making a sample code needs more time, I try to explain the cases.
But, I understand the cases can be ambiguous. If you would like to explain each cases as a code, I'll make it.

## Case 1. To detect that an agent is died unexpectedly.
- In the websocket case, the server easily determine the agent is dead because of the connection is disconnected.
- However, in the http case, the server cannot determine the agent is dead because the connection is instant.
  - To determine its liveness, if the connection is http, the server should implements additional logic.:
     - e.g. if a last http connection is over 2 * (http polling time), the agent is dead unexpectedly.

## Case 2. To send `ServerToAgent` message.
- When a user want to send serverToAgent message using opamp server, it is easy in the case of websocket connection, because finding `connection` in connection list & directly invoke `Send()` method.
- However, http connection is instant, the server cannot save the connection and saving connection is meaningless. So, we need to implement additional logic for the next connection of http.

## Case 3. To handle connectionMap.
- As I mentioned in the Case 2, http connection is instant and whenever to be connected, it is created.
- It means when the server handles many agents, unnecessarily register http connections to connectionMap & be deleted.

tigrannajaryan:
>In the websocket case, the server easily determine the agent is dead because of the connection is disconnected.

I am not sure this assertion is correct. The connection may drop briefly for a variety of reasons, not because the agent is dead. The agent may reconnect quickly.

I would still prefer to compute "time since last seen" to determine liveliness, regardless of the transport used.

>However, http connection is instant, the server cannot save the connection and saving connection is meaningless. So, we need to implement additional logic for the next connection of http.

I do not see why it is necessary to save a connection. To deliver something to an agent the server should remember the agent's instanceid and next time the agent connects the server can send whatever is necessary in the response when using http connection. This can be determined by error returned by Send() method of the Connection. If it fails (which it will with http transport) the server is supposed to deliver whatever it was trying at the next opportunity. Note that this is also required for WebSocket connection, since Send() is not guaranteed to succeed, it is always possible for Send() to fail because of network errors.

>It means when the server handles many agents, unnecessarily register http connections to connectionMap & be deleted.

That is one possible approach. Another approach is to keep connections in the map and only delete them after the particular agent is not seen for a period of time.

To summarize: I am not convinced any of these cases are better served by knowing the connection type.

minuk-dev:
> This can be determined by error returned by Send() method of the Connection. If it fails (which it will with http transport) the server is supposed to deliver whatever it was trying at the next opportunity.

Thank you. I understand. 

> To summarize: I am not convinced any of these cases are better served by knowing the connection type.

I understand you. I close this PR. Thank you for your detailed description.

Issue #179, opened 2023-06-27
Set `agent_disconnect` on disconnect


srikanthccv:
Related to https://github.com/open-telemetry/opamp-go/pull/174. Client initiated disconnect should set the `agent_disconnect` and send the `AgentToServer` message.

PR #384, opened 2025-05-08
fix: Send AgentToServer.agent_disconnect when client is stopped.

minuk-dev:
- In the [opamp-spec document](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agenttoserveragent_disconnect), 
  > AgentDisconnect MUST be set in the last AgentToServer message sent from the Client to the Server.

- But, currently there is no implementation for it.
  - Because of it, the below issue occurs.
    - https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/39676

- To resolve it, ensure to send `AgentToServer.agent_disconnect` when the client's `Stop()`.

[Edit]
related to #179 

[Edit - 2]
- To ensure to send last scheduled message, when client is stopped.

### DIFF
--- a/client/wsclient.go
+++ b/client/wsclient.go
@@ -116,6 +116,13 @@ func (c *wsClient) Start(ctx context.Context, settings types.StartSettings) erro
 }
 
 func (c *wsClient) Stop(ctx context.Context) error {
+	// AgentDisconnect MUST be set in the last AgentToServer message sent from the Client to the Server.
+	c.sender.NextMessage().Update(
+		func(msg *protobufs.AgentToServer) {
+			msg.AgentDisconnect = &protobufs.AgentDisconnect{}
+		},
+	)
+	c.sender.ScheduleSend()
 	return c.common.Stop(ctx)
 }
 
--- a/client/clientimpl_test.go
+++ b/client/clientimpl_test.go
@@ -516,15 +516,23 @@ func TestFirstStatusReport(t *testing.T) {
 
 		// Start a Server.
 		srv := internal.StartMockServer(t)
+		var isFirstSrvMessage atomic.Bool
+		isFirstSrvMessage.Store(true)
 		srv.OnMessage = func(msg *protobufs.AgentToServer) *protobufs.ServerToAgent {
-			assert.EqualValues(t, 0, msg.SequenceNum)
-			return &protobufs.ServerToAgent{
-				InstanceUid:  msg.InstanceUid,
-				RemoteConfig: remoteConfig,
+			if isFirstSrvMessage.Load() {
+				isFirstSrvMessage.Store(false)
+				assert.EqualValues(t, 0, msg.SequenceNum)
+				return &protobufs.ServerToAgent{
+					InstanceUid:  msg.InstanceUid,
+					RemoteConfig: remoteConfig,
+				}
 			}
+			return &protobufs.ServerToAgent{InstanceUid: msg.InstanceUid}
 		}
 
 		// Start a client.
+		var isFirstClientMessage atomic.Bool
+		isFirstClientMessage.Store(true)
 		var connected, remoteConfigReceived int64
 		settings := types.StartSettings{
 			Callbacks: types.Callbacks{
@@ -532,10 +540,13 @@ func TestFirstStatusReport(t *testing.T) {
 					atomic.AddInt64(&connected, 1)
 				},
 				OnMessage: func(ctx context.Context, msg *types.MessageData) {
-					// Verify that the client received exactly the remote config that
-					// the Server sent.
-					assert.True(t, proto.Equal(remoteConfig, msg.RemoteConfig))
-					atomic.AddInt64(&remoteConfigReceived, 1)
+					if isFirstClientMessage.Load() {
+						isFirstClientMessage.Store(false)
+						// Verify that the client received exactly the remote config that
+						// the Server sent.
+						assert.True(t, proto.Equal(remoteConfig, msg.RemoteConfig))
+						atomic.AddInt64(&remoteConfigReceived, 1)
+					}
 				},
 			},
 			Capabilities: protobufs.AgentCapabilities_AgentCapabilities_AcceptsRemoteConfig,
--- a/client/internal/mockserver.go
+++ b/client/internal/mockserver.go
@@ -63,7 +63,7 @@ func newMockServer(t *testing.T) (*MockServer, *http.ServeMux) {
 				return
 			}
 
-			srv.handleWebSocket(t, w, r)
+			srv.handleWebSocket(w, r)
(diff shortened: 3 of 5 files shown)


client/internal/clientcommon.go
```diff
@@ -169,11 +175,24 @@ func (c *ClientCommon) Stop(ctx context.Context) error {
...
+	lastMsgCtx, lastMsgCancel := context.WithTimeout(ctx, defaultClientAgentDisconnectTimeout)
+	defer lastMsgCancel()
+
 	c.isStoppingMutex.Lock()
 	cancelFunc := c.runCancel
 	c.isStoppingFlag = true
 	c.isStoppingMutex.Unlock()
 
+	// Wait until the last message is sent.
+	<-lastMsgCtx.Done()
```
tigrannajaryan, line 194: Doesn't this unconditionally wait for `defaultClientAgentDisconnectTimeout`? What's the point of having this as a separate context?

tigrannajaryan: I think the desirable behavior is to wait until the message is sent. For http transport we should wait until an http response is received. For WS transport I think we should send the message and initiate a graceful socket shutdown.

minuk-dev: > For WS transport I think we should send the message and initiate a graceful socket shutdown. The spec defines the behavior for WS:
> To close a connection the Client MUST first send an AgentToServer message with agent_disconnect field set. The Client MUST then send a WebSocket [Close](https://datatracker.ietf.org/doc/html/rfc6455#section-5.5.1) control frame and follow the procedure defined by WebSocket standard.

Oh, I missed the spec. Sorry for my mistake. Do you think it is better not to send "agent_disconnect" in http case?

minuk-dev: > it honours the input context. Still, I don't think this is what we want to do.

(If I'm wrong, please feel free to talk. If there is an error, I apologize 🙇 ).

## In the websocket case

```
ClientCommon.StartConnectAndRun() // create runCtx & runCancel
|- wsClient.runUntilStopped(ctx context.Context)
|-- wsClient.runOneCycle(ctx context.Context) // create senderCtx & stopSender
|--- WSSender.Start(ctx context.Context, conn *websocket.Conn)
|---- WSSender.run(ctx context.Context)
```

- `WSSender.run`'s ctx is `senderCtx` which is a child context of `runCtx`.:
  - It means when we call `cancelFunc` (=`c.runCancel`), we cannot guarantee to send the last `AgentToServer` message.
- And because we set `isStoppingFlag` as true, `wsClient.runUntilStopped()`'s for loop will be end. and also `ClientCommon.StartConnectAndRun()` also will be end.

- So, we can wait `<-lastMsgCtx.Done()` for its graceful shutdown before cancel when we set `isStoppingFlag` as true.

## In the http case

```
ClientCommon.StartConnectAndRun() // create runCtx & runCancel
|- httpClient.runUntilStopped(ctx context.Context)
|-- HTTPSender.Run(ctx context.Context, ...)
```

- Hmm... I'm sorry. It looks always wait for `defaultClientAgentDisconnectTimeout`.  I think I should fix it to gracefully shutdown.

tigrannajaryan:
Sorry for the delay. I suggest that for now we implement agent_disconnect for WebSocket only. If necessary we can add it for http transport later.

client/internal/clientcommon.go
minuk-dev: I think this conversation was resolved because only websocket is supported.
ref.  https://github.com/open-telemetry/opamp-go/pull/384#issuecomment-2922502252

client/wsclient.go
```diff
@@ -114,6 +114,13 @@ func (c *wsClient) Start(ctx context.Context, settings types.StartSettings) erro
 }
 
 func (c *wsClient) Stop(ctx context.Context) error {
+	// AgentDisconnect MUST be set in the last AgentToServer message sent from the Client to the Server.
+	c.sender.NextMessage().Update(
+		func(msg *protobufs.AgentToServer) {
+			msg.AgentDisconnect = &protobufs.AgentDisconnect{}
+		},
+	)
+	c.sender.ScheduleSend()
```
tigrannajaryan, line 125: This schedules the sending but does not wait for sending to be done. The `Stop()` below cancels the context of the runner which checks for pending sending messages and sends them. I think this means that there is no guarantee that the scheduled message will be sent. It may or may not be sent depending on who wins the race.

I may be missing something in my analysis, please correct me if so.

minuk-dev: Condition

goroutine 1:
wsSender.Start(startCtx, conn)
|-wsSender.run()

goroutine 2:
wsClient.Stop(stopCtx)
|-wsSender.Stop(stopCtx)

---
Case 1. if startCtx is done:

```go
		case <-ctx.Done():
			select {
			// If there is a pending message, we will try to send it before closing the connection.
			case <-s.hasPendingMessage:
				stopCtx, cancel := context.WithTimeout(context.Background(), defaultSendCloseMessageTimeout)
				defer cancel()
				s.sendNextMessage(stopCtx)
			default:
			}
```
- sendNextMessage()

Case 2. Stop() is called:
- wait for stoppedSignal after trigger `runCancel()`.:
  - `runCancel()` triggers case 1.

```go
	// Wait until stopping is finished.
	select {
	case <-ctx.Done():
		return ctx.Err()
	case <-c.stoppedSignal:
	}
```

agardnerIT:
I believe this logic should be added to HTTP too in order for the collector to be spec-compliant.

tigrannajaryan:
The build is failing because of a bug that will be fixed by https://github.com/open-telemetry/opamp-go/pull/425
Need to wait for that PR to be merged first.

minuk-dev:
> I believe this logic should be added to HTTP too in order for the collector to be spec-compliant.

@agardnerIT I agree with you. But, I think the spec is ambiguous.
But, if we try to send `agent_disconnect`, we should one more `AgentToServer` message http.

https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#websocket-transport-opamp-client-initiated

> To close a connection the Client MUST first send an AgentToServer message with agent_disconnect field set. The Client MUST then send a WebSocket [Close](https://datatracker.ietf.org/doc/html/rfc6455#section-5.5.1) control frame and follow the procedure defined by WebSocket standard.

Should we suggest to add the spec like
"The Client SHOULD send the AgentToServer message with agent_disconnect field set, if the Client is HTTP"?
Honestly, I'm not sure about it. Because sometimes, the http client is hard to send additional message.
So I agree https://github.com/open-telemetry/opamp-go/pull/384#discussion_r2080829073 this review.

Could you please open new issue to discuss it, if you need?
Honestly, this PR is too longly opened... I'm not enough time to do it.

PR #389, opened 2025-05-26
Demo: Feat/proxy connection settings

michel-laterman:
Add ProxyURL an ProxyHeaders support to start settings as well as the example agent.

This PR will remain as a draft until the opamp-spec for the feature is released.
- Relates: https://github.com/open-telemetry/opamp-spec/pull/221

### DIFF
--- a/internal/examples/agent/agent/agent.go
+++ b/internal/examples/agent/agent/agent.go
@@ -116,7 +116,6 @@ func NewAgent(logger types.Logger, agentType string, agentVersion string, initia
 		}
 		agent.tlsConfig = tlsConfig
 	}
-	agent.tlsConfig = tlsConfig
 	if err := agent.connect(withTLSConfig(agent.tlsConfig)); err != nil {
 		agent.logger.Errorf(context.Background(), "Cannot connect OpAMP client: %v", err)
 		return nil


go.mod
```diff
@@ -1,11 +1,12 @@
 module github.com/open-telemetry/opamp-go
 
-go 1.22
+go 1.23.0
```
michel-laterman: Requires: https://github.com/open-telemetry/opamp-go/pull/423

client/wsclient.go
```diff
@@ -413,3 +426,46 @@ func (c *wsClient) runUntilStopped(ctx context.Context) {
...
+		if err != nil {
+			return err
+		}
+		c.dialer.NetDialContext = dialer.DialContext
+	case "https":
+		dialer, err := dialer.NewProxyConnectDialer(proxyURL, &net.Dialer{}, dialer.WithTLS(cfg), dialer.WithProxyConnectHeaders(headers))
+		if err != nil {
+			return err
+		}
+		c.dialer.NetDialTLSContext = dialer.DialContext
```
michel-laterman, line 464: This is a little messy (as we configure the `websockets.Dialer` to behave slightly differently for HTTP(S) proxies vs non-HTTP(S) proxies; but it's much easer to do at this point in time then it is to replace the entire websockets library.

If https://github.com/gorilla/websocket/pull/988 is ever merged and released we can clean this up a lot.

client/wsclient.go
```diff
@@ -419,3 +432,54 @@ func (c *wsClient) runUntilStopped(ctx context.Context) {
...
+	case "https":
+		if len(headers) > 0 {
+			dialer, err := dialer.NewProxyConnectDialer(proxyURL, &net.Dialer{}, dialer.WithTLS(cfg), dialer.WithProxyConnectHeaders(headers))
+			if err != nil {
+				return err
+			}
+			c.dialer.NetDialTLSContext = dialer.DialContext
+			return nil
+		}
+		c.dialer.Proxy = http.ProxyURL(proxyURL) // No connect headers, use a regular proxy
```
michel-laterman, line 482: This is a little messy (as we mix setting the dialer.Proxy and dialer.NetDial*Context methods); but it's much easer to do at this point in time then it is to replace the entire websockets library.

If https://github.com/gorilla/websocket/pull/988 is ever merged and released we can clean this up.

I've tested this locally using [tinyproxy](https://tinyproxy.github.io/).
I've confirmed that when the ProxyURL and ProxyHeaders are specified, the dialer is used and the proxy is used via `CONNECT`.
If I only specify the ProxyURL, the `c.dialer.Proxy = http.ProxyURL(proxyURL)` approach is used; one thing that is interesting to note is that this still results in `CONNECT` request to the proxy but that may be more to do with the use of websockets.

Steps to test:
1) install and start tinyproxy - no additional config is needed for basic testing
2) start the example opamp server
3) Add the `ProxyURL` and `ProxyHeaders` to the example agent start settings: https://github.com/open-telemetry/opamp-go/blob/b5f732b2cedc05090c1f49dfe4e3874c40c64a7e/internal/examples/agent/agent/agent.go#L110-L112
For example
```go
ProxyURL: "http://localhost:8888", // tinyproxy's default address
ProxyHeaders: http.Header{"test-proxy-key": []string{"test-val"}},
```
4) Start the example agent
5) Agent should start and connect to the server; check the proxy logs to ensure that the connection has been made through the proxy.

michel-laterman:
@andykellr @tigrannajaryan we're able to use a custom dialer func in order to get proxy connect headers working with gorilla/websockets.
I've created a custom dialer and have included it as a dependency.

The implementation is now ready for review.

client/wsclient.go
```diff
@@ -79,7 +83,16 @@ func (c *wsClient) Start(ctx context.Context, settings types.StartSettings) erro
 	}
 
 	// Prepare connection settings.
-	c.dialer = *websocket.DefaultDialer
+	c.dialer = websocket.Dialer{
+		Proxy:            http.ProxyFromEnvironment,
+		HandshakeTimeout: 45 * time.Second,
+	}
```
tigrannajaryan: Any particular reason to do this instead of what was previously?

client/wsclient.go
```diff
@@ -419,3 +432,54 @@ func (c *wsClient) runUntilStopped(ctx context.Context) {
...
+			dialer, err := dialer.NewProxyConnectDialer(proxyURL, &net.Dialer{}, dialer.WithTLS(cfg), dialer.WithProxyConnectHeaders(headers))
+			if err != nil {
+				return err
+			}
+			c.dialer.NetDialTLSContext = dialer.DialContext
+			return nil
+		}
+		c.dialer.Proxy = http.ProxyURL(proxyURL) // No connect headers, use a regular proxy
+	case "socks5":
+		c.dialer.Proxy = http.ProxyURL(proxyURL)
```
tigrannajaryan: Since it is the same as the "default" case perhaps eliminate it.

client/internal/httpsender_test.go
```diff
@@ -356,3 +357,57 @@ func TestPackageUpdatesWithError(t *testing.T) {
 
 	cancel()
 }
+
+func TestHTTPSenderSetProxy(t *testing.T) {
```
tigrannajaryan, line 361: Is there any way we can also add tests that use an actual proxy and connect through it? I don't know if there is a premade proxy library in Go we can use for that.

client/wsclient.go
michel-laterman: The change is so that proxy env vars are used by default (I made a similar change to the HTTP client). Should we change the default to be aware of proxies?

client/internal/httpsender_test.go
michel-laterman: For HTTP it's easy to use a testserver with a custom handler, i have not seen if that approach would work for websockets tests or if we would need a library

client/wsclient.go
tigrannajaryan: Sorry, I wasn't clear. `websocket.DefaultDialer` is defined like this already:
```go
var DefaultDialer = &Dialer{
	Proxy:            http.ProxyFromEnvironment,
	HandshakeTimeout: 45 * time.Second,
}
```
The change in this PR doesn't seem to change anything in the resulting behavior. Am I missing something?

client/internal/httpsender_test.go
tigrannajaryan: OK, would be great to add it for http transport for now and skip for websockets.

client/wsclient.go
michel-laterman: I think I may have avoided using the default directly as we modify the dialer (we'll avoid needing something like #425 in the future)

client/internal/httpsender_test.go
michel-laterman: added below

client/wsclient.go
tigrannajaryan: I don't understand. The line `c.dialer = *websocket.DefaultDialer` makes a copy of websocket.DefaultDialer. Any subsequent modification of c.dialer will not impact websocket.DefaultDialer.

The reason https://github.com/open-telemetry/opamp-go/pull/425 was needed because we were incorrectly taking a pointer to http.DefaultClient and were modifying it through that pointer. That is not the case here, we are not taking a pointer, we are making a copy.

michel-laterman: got it, reverted

PR #390, opened 2025-05-27
Demo: Feat/connection settings status

michel-laterman:
Add support to automatically send connection settings status from the
client to the server. When a client receives a message with offered
connection settings it will immediately send a message stating that it is
applying the settings. If the offered connection settings contain OpAMP
settings the success/failure message will automatically be communicted.
For other connection types the developer will need to call
SetConnectionStatus within the OnMessage callback hook.

- Relates: https://github.com/open-telemetry/opamp-spec/pull/220


client/httpclient.go
```diff
@@ -125,6 +125,11 @@ func (c *httpClient) SetAvailableComponents(components *protobufs.AvailableCompo
 	return c.common.SetAvailableComponents(components)
 }
 
+// SetConnectionStatus implements OpAMPClient.SetConnectionStatus
+func (c *httpClient) SetConnectionStatus(status *protobufs.ConnectionSettingsStatus, scheduleSend bool) error {
+	return c.common.SetConnectionStatus(status, scheduleSend)
+}
```
andykellr, line 131: Let just use `SetConnectionSettingsStatus` to be consistent with the `ConnectionSettingsStatus` message.

client/types/callbacks.go
```diff
@@ -18,6 +18,7 @@ type MessageData struct {
 
 	// Connection settings are offered by the Server. These fields should be processed
 	// as described in the ConnectionSettingsOffers message.
+	OfferedConnectionsHash []byte
```
tigrannajaryan:
```suggestion
	OfferedConnectionsSettingsHash []byte
```

client/internal/receivedprocessor.go
```diff
@@ -92,6 +94,20 @@ func (r *receivedProcessor) ProcessReceivedMessage(ctx context.Context, msg *pro
 	}
 
 	if msg.ConnectionSettings != nil {
+		msgData.OfferedConnectionsHash = msg.ConnectionSettings.Hash
+		if r.hasCapability(protobufs.AgentCapabilities_AgentCapabilities_ReportsConnectionSettingsStatus) {
+			connectionStatus := &protobufs.ConnectionSettingsStatus{
+				LastConnectionSettingsHash: msg.ConnectionSettings.Hash,
+				Status:                     protobufs.RemoteConfigStatuses_RemoteConfigStatuses_APPLYING,
```
tigrannajaryan: I think we should only set status to APPLYING if corresponding Accepts*ConnectionSettings agent capability is defined. Otherwise we should probably report FAILED status and set an appropriate error message saying capability not supported.

client/internal/receivedprocessor.go
```diff
@@ -92,6 +94,20 @@ func (r *receivedProcessor) ProcessReceivedMessage(ctx context.Context, msg *pro
...
+			}
+			if err := r.clientSyncedState.SetConnectionSettingsStatus(connectionStatus); err != nil {
+				r.logger.Errorf(ctx, "Unable to persist connection settings status applying state: %v", err)
+			}
+			r.sender.NextMessage().Update(func(sendMsg *protobufs.AgentToServer) {
+				sendMsg.ConnectionSettingsStatus = connectionStatus
+			})
+			r.sender.ScheduleSend() // TODO: this might be noisy
```
tigrannajaryan: Just set `scheduled = true` instead and it will schedule a send once at the end of this function.

internal/examples/agent/agent/agent.go
```diff
@@ -633,3 +631,9 @@ func (agent *Agent) getCertFromSettings(certificate *protobufs.TLSCertificate) (
 
 	return &cert, nil
 }
+
+func (agent *Agent) onConnectionSettings(ctx context.Context, settings *protobufs.ConnectionSettingsOffers) error {
+	agent.logger.Debugf(context.Background(), "Received connection settings offers from server, hash=%x.", settings.Hash)
+	// TODO handle traces, logs, and other connection settings
+	return agent.initMeter(settings.OwnMetrics)
```
tigrannajaryan: We should do this only if settings.OwnMetrics != nil. Otherwise it is going to panic when trying to create the Meter.

client/internal/receivedprocessor.go
michel-laterman: The callbacks all apply before schedule is checked. This means that the connection settings can be applied before APPLYING is sent, so unless we change the order of the check and callbacks I don't think that's what we want.

tigrannajaryan: Is that a problem? The callbacks may apply the settings asynchronously, in which case APPLYING will be appropriate. Even if the callback apply the settings synchronously, that's still fine, since in that case they are supposed to set the status to APPLIED, which will overwrite this APPLYING state, so we will never send APPLYING state.

michel-laterman: Ah, yeah I was thinking too much about the example agent we have; i'll make the change.

client/internal/receivedprocessor.go
```diff
@@ -281,3 +402,16 @@ func (r *receivedProcessor) rcvCommand(ctx context.Context, command *protobufs.S
...
 	}
 }
+
+// updateStoredConnectionSettingsStatus returns a bool of if status should replace oldStatus.
+// It's true if:
+// - no oldStatus
+// - hash changes
+// - status changes from APPLYING or UNSET
+// - status changes to FAILED
+func updateStoredConnectionSettingsStatus(oldStatus, status *protobufs.ConnectionSettingsStatus) bool {
```
tigrannajaryan, line 412: Should this return true if oldStatus.Status == status.Status?

client/internal/receivedprocessor.go
```diff
@@ -281,3 +402,16 @@ func (r *receivedProcessor) rcvCommand(ctx context.Context, command *protobufs.S
 		r.callbacks.OnCommand(ctx, command)
 	}
 }
+
+// updateStoredConnectionSettingsStatus returns a bool of if status should replace oldStatus.
+// It's true if:
+// - no oldStatus
+// - hash changes
+// - status changes from APPLYING or UNSET
```
tigrannajaryan, line 410: I am not sure I understand why this particular set of conditions was chosen. For example if the previous Status for FAILED and the current status is APPLYING or APPLIED, why shouldn't we update it?

tigrannajaryan (commented):
Sorry for the delay, I was out for 2 weeks.

client/internal/receivedprocessor.go
michel-laterman: no, if the statuses are the same there is no change and no need to update

client/internal/receivedprocessor.go
michel-laterman: To reword a bit:
We will always update if there is no old status
We will always update if the old status, and new status have a different hash

If the hash between the old/new status' are the same, we will allow updates from
- `UNSET|APPLYING -> APPLIED|FAILED`
- `ANY -> FAILED`

The reason why `FAILED` if preferred over `APPLIED` is because we have two seperate callbacks that can execute independently; if the server offers connection settings for OpAMP, and something like traces we want to signal a failure if either has an error.

michel-laterman:
@tigrannajaryan can you please re-review?

andykellr (approved):

andykellr:
@tigrannajaryan I can merge and release if you think it is ready.

tigrannajaryan:
@andykellr sounds good. Please also update the CHANGELOG for opamp-spec v0.13 release.

PR #398, opened 2025-06-22
Replace staticcheck with golangci-lint

### LINKED ISSUE #321: Consolidate lint stage checks with golangci-lint
We agreed to take this approach https://github.com/open-telemetry/opamp-go/pull/318#issuecomment-2489540935.

tylerauerbeck:
As mentioned in #321 and associated issues/PR's, this PR adds `golangci-lint` configs (mostly adapted from the [opentelemetry-collector config](https://github.com/open-telemetry/opentelemetry-collector/blob/main/.golangci.yml). This enables just the `staticcheck` linter for now since this is what it's replacing and then the standard formatters. The existing lints that needed fixed were then addressed in separate commits.

This also updates the appropriate workflow (now named `lint.yaml`) and runs it using the latest `golangci-lint` v2 version.

Resolves #321


.github/workflows/lint.yml
```diff
@@ -0,0 +1,27 @@
+name: static-check
```
michel-laterman, line 1: (nitpick): Should this be renamed to lint?

.github/workflows/lint.yml
michel-laterman, line 1: Should we enable annotations from golangci-lint: https://github.com/golangci/golangci-lint-action?tab=readme-ov-file#annotations?

PR #401, opened 2025-06-23
chore(ci): add govulncheck workflow

### LINKED ISSUE #294: Add govulncheck to the CI
We should check for vulnerabilities in each push: https://pkg.go.dev/golang.org/x/vuln/cmd/govulnchec

For example, the current Server (Go 1.20) has the following vulnerabilities: 

```
Vulnerability #2: GO-2024-2887
    Unexpected behavior from Is methods for IPv4-mapped IPv6 addresses in
    net/netip
  More info: https://pkg.go.dev/vuln/GO-2024-2887
  Standard library
    Found in: net/netip@go1.21.10
    Fixed in: net/netip@go1.21.11
    Example traces found:
Error:       #1:  server.server.Start, which eventually calls netip.Addr.IsLoopback
Error:       #2:  server.server.Start, which eventually calls netip.Addr.IsMulticast
```

tylerauerbeck:
Per #294, adds govulncheck to a workflow and additionally to the makefile so it can be run during CI and locally.

Resolves #294


.github/workflows/govuln.yml
```diff
@@ -0,0 +1,28 @@
...
+    name: govulncheck (project)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+      - name: Setup Go
+        uses: actions/setup-go@v5.5.0
+        with:
+          go-version: "1.22"
```
michel-laterman, line 23:
```suggestion
          go-version-file: "go.mod"
```

michel-laterman (commented):
I've changed other workflows to use `go.mod` as the source for the go version

.github/workflows/govuln.yml
```diff
@@ -0,0 +1,28 @@
...
+      - id: govulncheck
+        uses: golang/govulncheck-action@v1
+        with:
+           go-version-input: 1.22
```
michel-laterman, line 27:
```suggestion
           go-version-file: "go.mod"
```

PR #403, opened 2025-06-27
Plumb TLS Config to Package Syncer

owais:
Package Syncer uses the default HTTP Client from stdlib (http.DefaultClient) and ignores the TLS configuration provided by the user. This means opamp client can correctly uses TLS configuration for everything but fails when it comes to downloading packages. This is important in situtations where a server needs custom TLS config to work correctly.

This commit ensures Package Syncer uses the same TLS config that HTTP and WebSocker clients use to talk to the server.

This is technically a bug fix. Users passing TLS config to StartSettings would expect opamp client to use the settings for all connections which is not true. This PR ensures the client works as configured for all outgoing connections.

### DIFF
--- a/client/internal/httpsender.go
+++ b/client/internal/httpsender.go
@@ -58,6 +58,7 @@ type HTTPSender struct {
 	url                string
 	logger             types.Logger
 	client             *http.Client
+	tlsConfig          *tls.Config
 	callbacks          types.Callbacks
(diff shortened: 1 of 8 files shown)


tigrannajaryan:
Closed in favour of https://github.com/open-telemetry/opamp-go/pull/405

PR #405, opened 2025-06-27
Add DownloadHTTPClient callback

owais:
Package Syncer uses the http.DefaultClient and gives users not chance of configuring an HTTP client for the purpose of downloading packages. This means users cannot configure TLS, timeouts, keep-alive, use a custom transport, a proxy, etc to faciliate package file downloads.

This commit add a new DownloadHTTPClient callback that allows users to provide a custom HTTP Client to use with package downloads.

The callback will be called for every DownloadableFile. This is to allow users to tweak the client based on the file URL. Users who do not need to tweak the client for each download can always return the same client. We could call the callback just once per agent and use the same client for all downloads but I felt this gives users more flexibility and makes it very clear that the client is only for downloading files. Happy to change it to a "global" download client though.

### DIFF
--- a/client/types/callbacks.go
+++ b/client/types/callbacks.go
@@ -130,6 +130,11 @@ type Callbacks struct {
 	//
 	// The responses in the via parameter are passed with their bodies closed.
 	CheckRedirect func(req *http.Request, viaReq []*http.Request, via []*http.Response) error
+
+	// DownloadHTTPClient is called to create an HTTP client that is used to download files by the package syncer.
+	// If the callback is not set, a default HTTP client will be created with the default transport settings.
+	// The callback must return a non-nil HTTP client or an error.
+	DownloadHTTPClient func(ctx context.Context, file *protobufs.DownloadableFile) (*http.Client, error)
 }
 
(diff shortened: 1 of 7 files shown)


client/types/callbacks.go
```diff
@@ -130,6 +130,9 @@ type Callbacks struct {
 	//
 	// The responses in the via parameter are passed with their bodies closed.
 	CheckRedirect func(req *http.Request, viaReq []*http.Request, via []*http.Response) error
+
+	// DownloadHTTPClient is called to create an HTTP client that is used to download files by the package syncer.
+	DownloadHTTPClient func(ctx context.Context, file *protobufs.DownloadableFile) (*http.Client, error)
```
tigrannajaryan, line 137: Document that when this is nil a clone of the DefaultClient is used.

tigrannajaryan:
Please fix the build.

client/types/callbacks.go
owais: Done.

tigrannajaryan (approved):

Issue #422, opened 2025-07-23
TestRedirectHTTP is unstable


tigrannajaryan:
See https://github.com/open-telemetry/opamp-go/actions/runs/16477833514/job/46584067723?pr=421
```
Run make
go test -race ./...
2025/07/23 17:41:07 http: TLS handshake error from 127.0.0.1:51264: EOF
2025/07/23 17:41:07 http: TLS handshake error from 127.0.0.1:37120: EOF
==================
WARNING: DATA RACE
Write at 0x000000ff59b0 by goroutine 1290:
  github.com/open-telemetry/opamp-go/client/internal.(*HTTPSender).Run()
      /home/runner/work/opamp-go/opamp-go/client/internal/httpsender.go:107 +0x388
  github.com/open-telemetry/opamp-go/client.(*httpClient).runUntilStopped()
      /home/runner/work/opamp-go/opamp-go/client/httpclient.go:139 +0x224
  github.com/open-telemetry/opamp-go/client.(*httpClient).runUntilStopped-fm()
      <autogenerated>:1 +0x2a
  github.com/open-telemetry/opamp-go/client/internal.(*ClientCommon).StartConnectAndRun.func1()
      /home/runner/work/opamp-go/opamp-go/client/internal/clientcommon.go:245 +0xa1

Previous read at 0x000000ff59b0 by goroutine 899:
  net/http.(*Client).do()
      /opt/hostedtoolcache/go/1.23.11/x64/src/net/http/client.go:606 +0x13e
  net/http.(*Client).Do()
      /opt/hostedtoolcache/go/1.23.11/x64/src/net/http/client.go:590 +0x454
  github.com/open-telemetry/opamp-go/client/internal.(*HTTPSender).sendRequestWithRetries()
      /home/runner/work/opamp-go/opamp-go/client/internal/httpsender.go:204 +0x406
  github.com/open-telemetry/opamp-go/client/internal.(*HTTPSender).makeOneRequestRoundtrip()
      /home/runner/work/opamp-go/opamp-go/client/internal/httpsender.go:162 +0x44
  github.com/open-telemetry/opamp-go/client/internal.(*HTTPSender).Run()
      /home/runner/work/opamp-go/opamp-go/client/internal/httpsender.go:119 +0x584
  github.com/open-telemetry/opamp-go/client.(*httpClient).runUntilStopped()
      /home/runner/work/opamp-go/opamp-go/client/httpclient.go:139 +0x224
[...]

PR #423, opened 2025-07-23
Update go version to v1.23.0

michel-laterman:
Update the required min go version to `v1.23.0`; alter workflows to use the `go.mod` when determining what go version to use.

### DIFF
--- a/client/internal/wssender.go
+++ b/client/internal/wssender.go
@@ -91,8 +91,7 @@ func (s *WSSender) SetHeartbeatInterval(d time.Duration) error {
 func (s *WSSender) shouldSendHeartbeat() <-chan time.Time {
 	t := s.heartbeatTimer
 
-	// Before Go 1.23, the only safe way to use Reset was to [Stop] and
-	// explicitly drain the timer first.
+	// Handle both GODEBUG=asynctimerchan=[0|1] properly.
 	// ref: https://pkg.go.dev/time#Timer.Reset
 	if !t.Stop() {
 		select {


.github/workflows/static-check.yml
```diff
@@ -26,9 +26,9 @@ jobs:
       - id: install_go
         uses: WillAbides/setup-go-faster@v1.14.0
         with:
-          go-version: "1.22.x"
+          go-version-file: "go.mod"
       - uses: dominikh/staticcheck-action@v1.4.0
         with:
           install-go: false
-          version: "2023.1.7"
```
michel-laterman, line 33: @tigrannajaryan, this version of staticcheck didn't support v1.23+, I've updated it to the latest released version and now all the checks execute.

michel-laterman:
It looks like `TestRedirectHTTP/check_redirect` may be flaky, it succeeded in a build before I [changed a comment](https://github.com/open-telemetry/opamp-go/actions/runs/16479529940/job/46589902464), but has failed on the [latest attempt](https://github.com/open-telemetry/opamp-go/actions/runs/16479682244/job/46590438479?pr=423)

tigrannajaryan:
Looks like one of the tests failed again, the same as here https://github.com/open-telemetry/opamp-go/issues/422

I will rerun for now.

.github/workflows/static-check.yml
tigrannajaryan: Thanks. That's what I was missing.

tigrannajaryan (approved):

client/internal/wssender.go
```diff
@@ -91,8 +91,7 @@ func (s *WSSender) SetHeartbeatInterval(d time.Duration) error {
 func (s *WSSender) shouldSendHeartbeat() <-chan time.Time {
 	t := s.heartbeatTimer
 
-	// Before Go 1.23, the only safe way to use Reset was to [Stop] and
-	// explicitly drain the timer first.
+	// Handle both GODEBUG=asynctimerchan=[0|1] properly.
```
tigrannajaryan, line 94: I am not sure I understand this comment.

michel-laterman: The update to go v1.23 changed the default behaviour of the timer. However a user can still enable the old behaviour with the GODEBUG env var.
The comment change is to make the drain behaviour reference the env var instead of a version number.

PR #425, opened 2025-07-31
Fix incorrect usage of DefaultHttpClient causing data race

tigrannajaryan:
DefaultHttpClient is a global variable and should not be modified concurrently.

We now clone it before modifying.

Resolves https://github.com/open-telemetry/opamp-go/issues/422

### DIFF
--- a/client/internal/utils/httpclient.go
+++ b/client/internal/utils/httpclient.go
@@ -1,0 +1,13 @@
+package utils
+
+import "net/http"
+
+func NewHttpClient() *http.Client {
+	client := &http.Client{}
+	// Clone the default transport to avoid being affected by global state changes.
+	if tr, ok := http.DefaultTransport.(*http.Transport); ok {
+		tr = tr.Clone()
+		client.Transport = tr
+	}
+	return client
+}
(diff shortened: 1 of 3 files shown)


client/internal/utils/httpclient.go
```diff
@@ -0,0 +1,13 @@
+package utils
+
+import "net/http"
+
+func NewHttpClient() *http.Client {
+	client := &http.Client{}
+	// Clone the default transport to avoid being affected by global state changes.
+	if tr, ok := http.DefaultTransport.(*http.Transport); ok {
```
michel-laterman, line 8: Are we doing this specifically so that `http.DefaultTransport` is not nil? or are we able to just use `&http.Client{}` instead of a function to create a client?

tigrannajaryan: This is a pre-existing function, I just moved it from client/types/callbacks.go to a separate package to reuse.

I think the code is written defensively just in case the DefaultTransport is changed to be a different type in the future (I don't think it will, but still). I agree, probably unnecessary.

andykellr (approved):

evan-bradley (approved):
Makes sense to me.

PR #434, opened 2025-08-12
fix: Package Download race condition and close idle connections

dpaasman00:
Was running into a couple issues here while working on the supervisor to support upgrading the collector.

1. Data race
When syncing the package, we spin off a go routine to provide periodic updates on the progress. TLDR this go routine doesn't clean up by the time we update the package state with a successful install. This change adds a wait group so we make sure this routine exits before we continue.

2. Idle go routines
After resolving the data race, was also running into the following error. This was resolved by calling "CloseIdleConnections" on the http client we get to download the file. 

If you have any questions about these changes, let me know!

### DIFF
--- a/client/internal/mocksender.go
+++ b/client/internal/mocksender.go
@@ -1,0 +1,38 @@
+package internal
+
+import (
+	"time"
+
+	"github.com/open-telemetry/opamp-go/client/types"
+)
+
+// MockSender is a mock implementation of the Sender interface for testing purposes.
+type MockSender struct {
+	nextMessage *NextMessage
+}
+
+// NewMockSender creates a new MockSender with default values.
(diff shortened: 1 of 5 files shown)


tigrannajaryan:
Thanks for the fix. Can you add a test that verifies the fix? Ideally try to trigger the race and see if it fails before the fix with `-race` flag enabled.

tigrannajaryan:
> I added the test but I wasn't able to replicate the race condition in this test environment. Let me know your thoughts. 

The test still looks useful so let's keep it even if it doesn't help hit the particular race.

>Also the codecov bot is alerting on a mock I added. I'm not sure if there's a way to ignore that file?

We can ignore it, it does not block merging.

Issue #443, opened 2025-08-28
Feature Request: Official Docker image for OpAMP server


kilian-goetz:
### Problem
Currently, the `opamp-go` repository provides an example OpAMP server implementation (`internal/examples/server`), but there is **no official Docker image** available.  
This makes it harder to try out the server in environments where Docker is the primary way to deploy observability components.

Users who want to test OpAMP alongside the OpenTelemetry Collector and the OpAMP Supervisor must currently:
- clone the repo,
- build the server binary with Go,
- optionally create their own container image.

### Proposal
It would be very helpful to provide an **official Docker image** for the OpAMP server, published to a trusted registry (e.g. `ghcr.io/open-telemetry/opamp-server` or on the open telemetry repository on the docker hub).

This image would allow users to easily spin up a minimal control-plane with:
- the OpAMP server listening on `4320` (OpAMP protocol),
- the UI exposed on `4321`,
- suitable defaults for trying out the protocol with the Supervisor.

### Alternatives
- Community members can build their own images (e.g. with a simple Dockerfile wrapping the example server), but this leads to fragmentation and unverified builds.
- There are some unofficial images on Docker Hub, but they are not maintained by OpenTelemetry.

### Use Cases
- Developers who want to test OpAMP locally with Docker Compose (`opamp-server + opampsupervisor`).
- Simplifying demos, tutorials, and documentation around OpAMP.

### Additional Context
The OpenTelemetry ecosystem already provides official images for the Collector and OpAMP Supervisor.  
Having an official image for the OpAMP server would complete the stack and make it much easier for new users to experiment with agent management.

Saraah910:
Hi,
Is someone already working on this?

juandemanjon:
Please provide feedback before starting the integration with the build/publish process

juandemanjon:
I had created two images for the server and agent

```bash
docker run --rm -p 4320:4320 -p 4321:4321  ghcr.io/juandemanjon/opamp-server-example:issue_443-7be09e8
```
```
2025/12/09 17:57:55.477638 [MAIN] OpAMP Server starting...
2025/12/09 17:57:55.478677 [MAIN] OpAMP Server running...
[...]

PR #447, opened 2025-09-14
Dockerfile for creating an image to test the opamp-server on localhost with supervisor

Saraah910:
Issue: [https://github.com/open-telemetry/opamp-go/issues/443](https://github.com/open-telemetry/opamp-go/issues/443)

This PR adds a clean Docker-based setup for running the OpAMP server, its UI, and the Supervisor together for local testing and demos.

Changes included:

1. Added Dockerfile and entrypoint.sh to build and run both the OpAMP server and Supervisor in a single container.
2. Configured ports:
       - 4320 for the OpAMP server (protocol)
       - 4321 for the web UI

3. The entry point script ensures the Supervisor connects automatically to the local OpAMP server.

Please suggest changes or improvements if the solution is not as desired.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+
+
+sleep 2
+
+/bin/supervisor --server-url ws://127.0.0.1:4320/v1/opamp &
+
+wait
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -2,7 +2,6 @@
 
 
 sleep 2
-
 /bin/supervisor --server-url ws://127.0.0.1:4320/v1/opamp &
 
 wait
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,8 @@
-# syntax=docker/dockerfile:1
 FROM golang:1.23 AS builder
 
 WORKDIR /src
-
-# Copy all source code
 COPY . /src
-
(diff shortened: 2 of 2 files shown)


entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+
+sleep 2
+/bin/supervisor --server-url ws://127.0.0.1:4320/v1/opamp &
```
tigrannajaryan, line 5: What does --server-url parameter do?

Saraah910: That parameter actually does not exist, I was thinking of hardcoding the URL by creating that particular flag in the main.go file of agent dir. But was not sure if the approach is correct or not.

Thought process: running the container, the opamp-server will start listening on port 4320 & 4321, then with some flag or something, providing the endpoint where agents will connect. One container will do both.

Saraah910: Lil new to opamp. I'd love insights from you to make necessary changes and understand your thought process.

tigrannajaryan: Since you are new to the project I suggest starting with simpler issues.

Also, is this an AI-generated PR? If yes see [the policy](https://github.com/open-telemetry/community/blob/main/guides/contributor/genai.md#frequently-asked-questions---contributors).

PR #450, opened 2025-09-18
Use client.SetCapabilities() instead of deprecated Capabilities in the Agent example

juandemanjon:
This minor PR, remove the deprecated message in the agent example log

```
2025/09/17 22:22:22 Starting OpAMP client...
2025/09/17 22:22:22 settings.Capabilities is deprecated, use client.SetCapabilities() instead
2025/09/17 22:22:22 OpAMP Client started.
```

### DIFF
--- a/internal/examples/agent/agent/agent.go
+++ b/internal/examples/agent/agent/agent.go
@@ -171,12 +171,6 @@ func (agent *Agent) connect(ops ...settingsOp) error {
 			OnConnectionSettings:      agent.onConnectionSettings,
 		},
 		RemoteConfigStatus: agent.remoteConfigStatus,
-		Capabilities: protobufs.AgentCapabilities_AgentCapabilities_AcceptsRemoteConfig |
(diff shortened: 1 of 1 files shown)


internal/certs.go
```diff
@@ -46,7 +41,32 @@ func CreateClientTLSConfig(clientCert *tls.Certificate, caCertPath string) (*tls
 	return cfg, nil
 }
 
-func CreateServerTLSConfig(caCertPath, serverCertPath, serverKeyPath string) (*tls.Config, error) {
+func CreateServerTLSConfig(caCertBytes, serverCertBytes, serverKeyBytes []byte) (*tls.Config, error) {
```
michel-laterman: I think the certificate stuff should be moved to a separate PR

juandemanjon: My bad the second commit was intended for another PR for the issue #451. I did revert it

tigrannajaryan (approved):

Issue #460, opened 2025-11-10
Allow Custom Endpoint in the Agent Example


juandemanjon:
# Allow Custom Endpoint in the Agent Example

The current Agent example always attempts to connect to the OpAMP server at localhost, making it impossible to use the client example to test remote OpAMP servers.

To make the example more flexible and consistent with OpenTelemetry Collector conventions, I propose introducing a new argument: 

`--endpoint wss://opamp.myserver.com:4320/v1/opamp`

If `--endpoint` is not provided, it will default to `wss://localhost:4320/...` like it does now.

To further align with existing Collector configuration patterns, Instead of `--initial-insecure-connection` we can adopt `--tls-insecure`, and `--tls-insecure_skip_verify` to enable TLS but not verify the certificate. This change would make the Agent example more intuitive for users familiar with OpenTelemetry components.

Following the same pattern, we can also support additional TLS configuration options consistent with the Collector:

```
--tls-cert_file /etc/ssl/certs/client.crt           \
--tls-key_file /etc/ssl/private/client.key          \
--tls-ca_file /etc/ssl/certs/client-ca.crt 
```

By default, the agent can still use the embedded self-signed certificates like it does now.

PR #461, opened 2025-11-13
Add configurable TLS settings and server endpoint to agent example

juandemanjon:
This PR is related to issue #460 

 * Introduces configurable TLS options for the example agent.
 * Adds support for specifying a server endpoint.
 * Relays on ```go.opentelemetry.io/collector/config/configtls``` to get  the agent ```*tls.Config```
 * Server example is listening for incoming connections on all available network interfaces

Manual Testing
 * ```make run-examples``` 
 * ```cd internal/examples; go run agent/main.go --tls-insecure_skip_verify --endpoint wss://example:4320/v1/opamp```

### DIFF
--- a/internal/examples/server/opampsrv/opampsrv.go
+++ b/internal/examples/server/opampsrv/opampsrv.go
@@ -56,7 +56,7 @@ func (srv *Server) Start() {
 				},
 			},
 		},
-		ListenEndpoint: "127.0.0.1:4320",
+		ListenEndpoint: "0.0.0.0:4320",
 		HTTPMiddleware: otelhttp.NewMiddleware("/v1/opamp"),
 	}
 	tlsConfig, err := certs.CreateServerTLSConfig(


internal/examples/config/agent.go
```diff
@@ -0,0 +1,35 @@
...
+	"net/url"
+
+	"github.com/open-telemetry/opamp-go/internal/examples/certs"
+	"go.opentelemetry.io/collector/config/configopaque"
+	"go.opentelemetry.io/collector/config/configtls"
+)
+
+type AgentConfig struct {
+	Endpoint   string                 `mapstructure:"endpoint"`
+	TLSSetting configtls.ClientConfig `mapstructure:"tls,omitempty"`
```
michel-laterman, line 16: 👍

michel-laterman (approved):
The consistency with the collector is a good step; thanks for doing the work!

internal/examples/agent/main.go
```diff
@@ -7,6 +7,8 @@ import (
 	"os/signal"
 
 	"github.com/open-telemetry/opamp-go/internal/examples/agent/agent"
+	"github.com/open-telemetry/opamp-go/internal/examples/config"
+	"go.opentelemetry.io/collector/config/configtls"
```
tigrannajaryan, line 11: I am a bit reluctant to bring the collector as a dependency to opamp. Can we instead copy `configtls.ClientConfig` into this example and avoid adding the dependency? If the amount of copying is small I would prefer that.

juandemanjon: Thanks for the feedback! 

I wanted to clarify that the dependency is only added to the agent example (internal/examples/agent), not to the core opamp-go library itself, so it won't affect any consumers of the opamp-go module.

The agent example is already instrumented with OpenTelemetry and has several dependencies on other go.opentelemetry.io modules, so adding go.opentelemetry.io/collector/config/configtls fits naturally with the existing pattern.

Since this is example code showing best practices, I think using the standard OpenTelemetry Collector patterns makes sense rather than maintaining a fork of security-sensitive TLS code

That said, I'm happy to explore alternatives if you have strong concerns about the dependency!

PR #388, opened 2025-05-21
Explicitly close resp.Body in retry loop

rohit-sonawane:
https://github.com/open-telemetry/opamp-go/issues/387

Without fix :
As following it will show  number of BOUND connections for retrial after getting 503's

After Fix :
we see BOUND connections ports are being released

Issue #387, opened 2025-05-21
Too many BOUND connections in case of 503's and 429's on Windows Environments


rosonawa-cisco:
We have noticed too many ports with BOUND state leaking as a  result of  429's and 503's from server
to clients running on Windows environments 

After debugging a bit, we have  uncovered an issue with opamp-go library's [httpSender.sendRequestWithRetries](https://github.com/open-telemetry/opamp-go/blob/main/client/internal/httpsender.go#L215) method, where response body is not getting closed explicitly in case of 429, 503 before going on for retry -
We tried a fix and it looks encouraging (no port leaks).  
Currently observing a fix and will submit a patch in couple of days , after verifying there are no further leaks 

This happens in windows as explicit call to resp.Body.Close() is necessary to cleanup tcp sockets. In linux because of connection reuse, old connections will get used up

CC : @pavankrish123 , @kelseyma

PR #463, opened 2025-11-18
Explicitly close resp.Body in retry loop 

pavankrish123:
#387 

Without fix 

446803871-3c4e75ac-c29b-4c8e-8ce1-5418ddb7d380

With fix 
446804112-651f633e-8a35-4291-bdcc-35635201a2f6

Redid the PR #388  but with tests this time :) 

@rohit-sonawane @kelseyma


client/internal/httpsender.go
```diff
@@ -243,6 +243,7 @@ func (h *HTTPSender) sendRequestWithRetries(ctx context.Context) (*http.Response
 					case http.StatusTooManyRequests, http.StatusServiceUnavailable:
 						interval = recalculateInterval(interval, resp)
 						err = fmt.Errorf("server response code=%d", resp.StatusCode)
+						_ = resp.Body.Close()
 
 					default:
 						return nil, fmt.Errorf("invalid response from server: %d", resp.StatusCode)
```
tigrannajaryan: Do we need a Close() here too?

client/internal/httpsender.go
tigrannajaryan: Is the caller responsible for Close() in this case?

client/internal/httpsender.go
pavankrish123: Technically it is returning from the function and no more retires will be done so no fd leak.

Let me try to restructure the logic a bit and see if we can reason about the code properly

client/internal/httpsender.go
pavankrish123: @tigrannajaryan  let me know if the new structure makes sense now? 

 If I follow you correctly, the caller should be responsible for closing the resp - however in case when client is forced to do a retry  we will have temporary responses before we give up control to caller. I did a bit of refactor, see if it helps

client/internal/httpsender.go
```diff
@@ -230,39 +238,66 @@ func (h *HTTPSender) sendRequestWithRetries(ctx context.Context) (*http.Response
...
+			if result.interval > 0 {
+				interval = result.interval
+			}
+
+			// Log and notify about the retryable failure.
+			h.logger.Errorf(ctx, "Failed to do HTTP request (%v), will retry", result.err)
+			h.callbacks.OnConnectFailed(ctx, result.err)
+
 		case <-ctx.Done():
+			timer.Stop()
```
tigrannajaryan: Is this necessary? Docs say Stop() call is not needed for garbage collection.

client/internal/httpsender.go
```diff
@@ -230,39 +238,66 @@ func (h *HTTPSender) sendRequestWithRetries(ctx context.Context) (*http.Response
...
+
+	// Handle HTTP response status codes.
+	switch resp.StatusCode {
+	case http.StatusOK:
+		h.callbacks.OnConnect(ctx)
+		return requestResult{resp: resp, err: nil, retry: false}
+
+	case http.StatusTooManyRequests, http.StatusServiceUnavailable:
+		retryInterval := recalculateInterval(currentInterval, resp)
+		_ = resp.Body.Close()
```
tigrannajaryan, line 287: For completeness I think we also need for read the body before closing to free the connection for reuse. Something like `io.Copy(ioutil.Discard, resp.Body)`.

client/internal/httpsender.go
tigrannajaryan: Yes, we already Close() it for the success case in `HTTPSender.receiveResponse`.

client/internal/httpsender.go
pavankrish123: +1

pavankrish123: done.

client/internal/httpsender.go
pavankrish123: +1

client/internal/httpsender_test.go
```diff
@@ -62,6 +63,49 @@ func TestHTTPSenderRetryForStatusTooManyRequests(t *testing.T) {
 	srv.Close()
 }
 
+func TestHTTPSenderRetryForStatusServiceUnavailable(t *testing.T) {
```
tigrannajaryan: Does this test case have anything to do with the new functionality added to Close() the body? It seems to verify unrelated functionality. If that is the case let's not add in the same PR. We want the PR to be about one subject - making Close() happen and having tests that verify that Close() happens.

client/internal/httpsender_test.go
```diff
@@ -501,3 +545,607 @@ func TestHTTPSenderSetProxy(t *testing.T) {
 		assert.True(t, connected.Load(), "test request did not use proxy")
 	})
 }
+
+func TestHTTPSenderClosesBodyOnStatusTooManyRequests(t *testing.T) {
```
tigrannajaryan: This is a good test but it seem TestAttemptRequest already covers it (unless I am missing something).
Can we simplify and reduce the number of test cases to what's really necessary. I don't want to overdo and have too many overlapping tests. There is maintenance cost to tests.

client/internal/httpsender_test.go
```diff
@@ -501,3 +545,607 @@ func TestHTTPSenderSetProxy(t *testing.T) {
...
+		setupContext        func() context.Context
+		currentInterval     time.Duration
+		wantRetry           bool
+		wantErr             bool
+		wantErrContains     string
+		wantStatusCode      int
+		wantIntervalGreater bool
+		wantOnConnectCalled bool
+		wantBodyClosed      bool
+		checkBodyClosed     bool
```
tigrannajaryan: Where is this used?

tigrannajaryan (commented):
Are the tests AI generated? I am getting a feeling of overwhelming amount of tests for no good reason.

client/internal/httpsender_test.go
pavankrish123: Sure @tigrannajaryan . I was making sure the refactored code didn't break the existing functionality

pavankrish123: removed.

client/internal/httpsender_test.go
pavankrish123: The test actually runs through the loop (excercising the time.NewTimer(modifiedInterval) vs attemptRequest which is basically just one time execution and testing the return value>

Further more the scenario of  429 → 200 (eventual success) is also tested here. It verifies if the Body is closed between 429 and 200. The TestAttemptRequest hits 429 once and validates the Body is closed. 

We can treat this more like an integration test and that one as unit test 

Let me know if you want me to remove the other 

- I am fine either way - The TestAttemptRequest is AI generated one to exercise the helper method

pavankrish123: removed these and simplified tests :)

client/internal/httpsender_test.go
```diff
@@ -501,3 +502,186 @@ func TestHTTPSenderSetProxy(t *testing.T) {
 		assert.True(t, connected.Load(), "test request did not use proxy")
 	})
 }
+
+func TestHTTPSenderClosesBody(t *testing.T) {
```
tigrannajaryan, line 506: 👍

client/internal/httpsender_test.go
```diff
@@ -501,3 +502,186 @@ func TestHTTPSenderSetProxy(t *testing.T) {
...
+					Value: &protobufs.AnyValue_StringValue{StringValue: "test-service"},
+				},
+			}},
+		}
+	})
+	sender.url = url
+	return sender
+}
+
+func TestSendRequestWithRetriesPrepareRequestError(t *testing.T) {
```
tigrannajaryan: What exactly is this test verifying? Is it something that was changed in this PR? If yes please add some comments to explain the purpose of the test.

pavankrish123: removed the test case. it wasn't testing anything from the changes that went into PR. But these lines doesn't have any coverage 

```go
	req, err := h.prepareRequest(ctx)
	if err != nil {
		if errors.Is(err, context.Canceled) {
			h.logger.Debugf(ctx, "Client is stopped, will not try anymore.")
		} else {
			h.logger.Errorf(ctx, "Failed prepare request (%v), will not try anymore.", err)
		}
		return nil, err
	}
	if req == nil {
		// Nothing to send.
		return nil, nil
	}
	
```

tigrannajaryan (approved):

PR #471, opened 2025-12-05
Clone existing Transport before updating TLSConfig

kelseyma:
Found this issue while doing some testing: When updating the TLS config, it will overwrite the existing client's Transport and all existing values are lost. We should instead clone the transport and only update the TLS config portion, similar to how SetProxy and NewHttpClient methods handle it.

### DIFF
--- a/client/internal/httpsender.go
+++ b/client/internal/httpsender.go
@@ -400,8 +400,13 @@ func (h *HTTPSender) EnableCompression() {
 
 func (h *HTTPSender) AddTLSConfig(config *tls.Config) {
 	if config != nil {
-		h.client.Transport = &http.Transport{
-			TLSClientConfig: config,
+		tlsTransport := &http.Transport{}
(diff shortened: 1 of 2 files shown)


client/internal/httpsender.go
```diff
@@ -400,8 +400,13 @@ func (h *HTTPSender) EnableCompression() {
 
 func (h *HTTPSender) AddTLSConfig(config *tls.Config) {
 	if config != nil {
-		h.client.Transport = &http.Transport{
-			TLSClientConfig: config,
+		tlsTransport := &http.Transport{}
```
tigrannajaryan, line 403: Nit: unnecessary operation if we end up executing line 406.

kelseyma: I added this to ensure transport is initialized and maintain some parity with prior behavior, but let me know if the preference is to do a hard fail if the clone doesn't work and I can update the logic.

PR #476, opened 2025-12-15
Use correct Go version for go-mod-tidy action

tigrannajaryan:
The action is failing and I think it is because it tries to use an old Go version, see https://github.com/open-telemetry/opamp-go/pull/475

### DIFF
--- a/.github/workflows/tidy.yml
+++ b/.github/workflows/tidy.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v6.0.0
         with:
-          go-version: "1.23"
+          go-version-file: go.mod
       - name: Setup Go Environment
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/tidy.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v6.0.0
         with:
-          go-version: "1.22"
+          go-version: "1.23"
```
michel-laterman:
```suggestion
          go-version-file: go.mod
```

If we use the `go.mod` file to get the go version from, we only need to change one version number.

tigrannajaryan: Replaced, thanks.

michel-laterman (approved):

PR #481, opened 2026-01-06
Add scale test mode to example agent

michel-laterman:
Add a scale test mode to the example agent that will start a lot of agents (1 goroutine per agent) that connect to a server.

This allows us to leverage all capabilities in the spec that have been added to the example agent.

- Closes https://github.com/open-telemetry/opamp-go/issues/485

### DIFF
--- a/internal/examples/agent/agent/agent.go
+++ b/internal/examples/agent/agent/agent.go
@@ -238,7 +238,7 @@ func (agent *Agent) connect(ops ...settingsOp) error {
 			customCapability_Health,
 		},
 	}
-	err = agent.opampClient.SetCustomCapabilities(customCapabilities)
+	err = agent.client.SetCustomCapabilities(customCapabilities)
 	if err != nil {
 		return err
 	}
@@ -635,7 +635,7 @@ func (agent *Agent) processCustomMessage(ctx context.Context, customMessage *pro
 
 func (agent *Agent) sendCustomMessage(ctx context.Context, message *protobufs.CustomMessage) error {
 	for {
-		sendingChan, err := agent.opampClient.SendCustomMessage(message)
+		sendingChan, err := agent.client.SendCustomMessage(message)
 
 		switch {
 		case err == nil:


internal/examples/scale/main.go
```diff
@@ -0,0 +1,111 @@
...
+				CAFile: tlsCAFile,
+			},
+		},
+	}
+
+	logger.Printf("Starting %d agents", agentCount)
+	// Create a slice to track agents so we can safely stop them later.
+	// Use of slice instead of a concurrent goroutine to reduce memory usage.
+	agents := make([]*agent.Agent, 0, agentCount)
+	for range agentCount {
```
juandemanjon: To be able to reuse this code, I recommend to move it out of main.go

tigrannajaryan:
The direction of this PR looks good to me. Modify the existing example agent to be reusable for scale testing.

internal/examples/makefile
```diff
@@ -57,21 +61,21 @@ run-docker: build-docker-server build-docker-agent
...
 docker-compose-scale:
-	@echo "Usage: make docker-compose-scale AGENTS=<number>"
+	@echo "Usage: make docker-compose-scale [AGENTS=<number>]"
 	@if [ -z "$(AGENTS)" ]; then \
-		echo "Error: AGENTS variable not set"; \
-		exit 1; \
+	  cd $(SRC_ROOT)/internal/examples && docker compose --profile scale up -d --build; \
+	else \
+	  cd $(SRC_ROOT)/internal/examples && SCALE_AGENT_COUNT=$(AGENTS) docker compose --profile scale up -d --build; \
 	fi
```
michel-laterman, line 76: @juandemanjon, I've added an example dockerfile, and have adjusted the `make docker-compose-scale` target to use it instead.
By default it starts 1000 agents, or `$AGENTS` if that is defined.
What do you think?

juandemanjon: We should keep separate processes with different PIDs. It will allow the Supervisor's use cases to restart the Agent on demand, which means that we should keep running separate agents in the Docker Compose.

You can add a new `make docker-compose-scale-test` or similar

I think this scale feature can be integrated with the agent instead of a separate tool. 

https://github.com/open-telemetry/opamp-go/issues/485#issuecomment-3730649894

juandemanjon: Please revert the `docker-compose-scale` changes

michel-laterman: reverted

internal/examples/agent/main.go
```diff
@@ -1,60 +1,312 @@
...
 	"flag"
+	"fmt"
 	"log"
+	"net/url"
 	"os"
 	"os/signal"
+	"strconv"
+	"time"
 
+	opampinternal "github.com/open-telemetry/opamp-go/internal"
```
juandemanjon, line 15: This will break if the examples are moved to another repository. Currently there’s no direct dependency on opamp-go/internal even though the examples live under internal/examples.

I recommend duplicating `NopLogger` in internal/examples/agent/agent/logger.go:
```go
type NopLogger struct{}

func (l *NopLogger) Debugf(ctx context.Context, format string, v ...interface{}) {}
func (l *NopLogger) Errorf(ctx context.Context, format string, v ...interface{}) {}
```

internal/examples/server/opampsrv/opampsrv.go
```diff
@@ -30,10 +44,41 @@ func NewServer(agents *data.Agents) *Server {
...
+		panic(err)
+	}
 
 	srv := &Server{
-		agents: agents,
-		logger: logger,
+		agents:     agents,
+		logger:     logger,
+		counter:    counter,
+		errCounter: errCounter,
```
juandemanjon: I suggest encapsulating the telemetry-related fields into a dedicated struct, e.g.:

```go
type metricsState struct {
    counter    metric.Int64Counter
    errCounter metric.Int64Counter
    meter      metric.Meter
    resource   *resource.Resource
    exporter   *myExporterType
}
```
```go
	srv := &Server{
		agents:     agents,
		logger:     logger,
		metricsState: metricsState,
```
And encapsulate the above resource, exporter, meter initialization in a `func (*metricsState) initialize() `

internal/examples/server/opampsrv/opampsrv.go
```diff
@@ -49,8 +94,15 @@ func (srv *Server) Start() {
 					return types.ConnectionResponse{
 						Accept: true,
 						ConnectionCallbacks: types.ConnectionCallbacks{
+							OnConnected:       func(ctx context.Context, _ types.Connection) { srv.counter.Add(ctx, 1) },
```
juandemanjon: Following the previous abstraction feedback, 
```go
OnConnected: func(ctx context.Context, _ types.Connection) { srv.metricsState.OnConnect(ctx) }
```

juandemanjon (commented):
Thank you for adding this feature to the agent example

internal/examples/agent/README.md
```diff
@@ -0,0 +1,42 @@
+# agent
+
+Agent provides provides and example agent implementation for the OpAMP protocol.
+
+Both HTTP and Websocket connections are supported.
```
juandemanjon: Add this to the readme: But the server example only supports Websocket connections

internal/examples/agent/main.go
michel-laterman: I used the existing NopLogger in order to avoid code duplication

michel-laterman:
Running a docker container running with `memory: 256M`, and `cpus: '2'` on a 2023 MacBook Pro with an Apple M3 Pro chip:
```
docker version
Client:
 Version:           29.1.3
 API version:       1.52
 Go version:        go1.25.5
 Git commit:        f52814d
 Built:             Fri Dec 12 14:48:46 2025
 OS/Arch:           darwin/arm64
 Context:           desktop-linux

Server: Docker Desktop 4.55.0 (213807)
 Engine:
  Version:          29.1.3
  API version:      1.52 (minimum version 1.44)
  Go version:       go1.25.5
  Git commit:       fbf3ed2
  Built:            Fri Dec 12 14:50:40 2025
  OS/Arch:          linux/arm64
  Experimental:     false
 containerd:
  Version:          v2.2.0
  GitCommit:        1c4457e00facac03ce1d75f7b6777a7a851e5c41
 runc:
  Version:          1.3.4
  GitCommit:        v1.3.4-0-gd6d73eb8
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

I destroy and recreate the the opamp-server for each test, `docker container stats` for the server shows::
```
scale-count: 4000
CONTAINER ID   NAME           CPU %     MEM USAGE / LIMIT   MEM %     NET I/O           BLOCK I/O   PIDS
dd0e2dd653ec   opamp-server   1.45%     200.7MiB / 256MiB   78.41%    22.9MB / 26.5MB   0B / 0B     18

scale-count: 5000
CONTAINER ID   NAME           CPU %     MEM USAGE / LIMIT   MEM %     NET I/O           BLOCK I/O         PIDS
b3e90a9afd5c   opamp-server   1.87%     247MiB / 256MiB     96.49%    25.1MB / 30.9MB   54.7MB / 89.8MB   18

scale-count: 6000
CONTAINER ID   NAME           CPU %     MEM USAGE / LIMIT   MEM %     NET I/O           BLOCK I/O       PIDS
c90eed3d64bd   opamp-server   5.82%     250.2MiB / 256MiB   97.75%    30.8MB / 37.4MB   442MB / 509MB   18
```

The above stats are what generally occur once the server is "at rest" (few messages are exchanged). No server error callbacks are triggered in the above tests. However the CPU usage goes over 200% during the initial message exchanges.

When the scale count is set to 10k we see:
```
CONTAINER ID   NAME           CPU %     MEM USAGE / LIMIT   MEM %     NET I/O           BLOCK I/O         PIDS
20843ec87456   opamp-server   6.15%     255.6MiB / 256MiB   99.85%    46.9MB / 57.7MB   1.57GB / 1.82GB   19
```
However (read) errors are present in the opamp-server

internal/examples/agent/main.go
```diff
@@ -1,60 +1,312 @@
...
+	agentType             string
+	agentVersion          string
+	tlsInsecure           bool
+	tlsInsecureSkipVerify bool
+	tlsCertFile           string
+	tlsKeyFile            string
+	tlsCAFile             string
+	endpoint              string
+	heartbeat             time.Duration
+	quiteAgent            bool
```
tigrannajaryan:
```suggestion
	quietAgent            bool
```

internal/examples/agent/main.go
```diff
@@ -1,60 +1,312 @@
...
+
+	// scale test options
+	runScale   bool
+	scaleCount uint64
```
tigrannajaryan: Do we need both? What if we remove runScale and if scaleCount==1 (by default) we run one agent, otherwise we run multiple.

internal/examples/agent/main.go
```diff
@@ -1,60 +1,312 @@
...
+	if cfg.runScale {
+		if cfg.scaleCount == 0 {
+			return errors.New("scale count must not be zero")
+		}
+	}
+	return nil
+}
+
+// loadEnv will attempt to load config options from environment variables.
+func loadEnv(cfg *flagConfig) {
```
tigrannajaryan, line 101: Why do we need both command line options and env variable control?

michel-laterman: Supporting env vars makes it easier to run from a container

internal/examples/agent/main.go
michel-laterman: good point

tigrannajaryan (approved):

PR #482, opened 2026-01-07
client will only store state in PrepareStart for enabled capabilities

michel-laterman:
Prepare start will now only set `RemoteConfigStatus` as state if `ReportsRemoteConfig` is set.
Ditto for `PackageStatus`.

### DIFF
--- a/client/internal/clientcommon_test.go
+++ b/client/internal/clientcommon_test.go
@@ -1,27 +1,15 @@
 package internal
 
 import (
-	"context"
 	"testing"
 	"time"
 
 	"github.com/open-telemetry/opamp-go/client/types"
+	opampinternal "github.com/open-telemetry/opamp-go/internal"
 	"github.com/open-telemetry/opamp-go/protobufs"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
-
-type testLogger struct {
-	testing.TB
-}
-
-func (t *testLogger) Debugf(_ context.Context, format string, v ...any) {
-	t.Logf("DEBUG: "+format, v...)
-}
-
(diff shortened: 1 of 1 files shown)


juandemanjon (approved):

client/internal/clientcommon.go
```diff
@@ -87,10 +87,15 @@ func (c *ClientCommon) validateCapabilities(capabilities protobufs.AgentCapabili
...
 		}
 	} else {
-		if capabilities&protobufs.AgentCapabilities_AgentCapabilities_AcceptsPackages != 0 ||
-			capabilities&protobufs.AgentCapabilities_AgentCapabilities_ReportsPackageStatuses != 0 {
-			return ErrPackagesStateProviderNotSet
-		}
+		// FIXME This behaviour introduces a bug:
+		// If we create a new client, then set capabilities the AcceptsPackages or ReportsPackageStatus
+		// it will return an error because the PackagesStateProvider is not set.
+		// PackagesStateProvider is set in PrepareStart, which expects SetCapabilities to be set before.
```
michel-laterman, line 93: I'm not sure what the intended behaviour here is, but to be explicit this occurs when one does:

```go
c := NewClientCommon(logger, sender)
err := c.SetCapabilities(protobufs.AgentCapabilities_AgentCapabilities_AcceptsPackages)
if err != nil {
    panic(err) // will panic
}
c.PrepareStart(ctx, types.StartSettings{
    PackagesStateProvider: NewInMemPackagesStore(),
})
```

@tigrannajaryan @andykellr, should I remove this else block? Or is this an issue with my unit test?

michel-laterman: To be specific; this is the same issue as https://github.com/open-telemetry/opamp-go/pull/449

PR #483, opened 2026-01-07
Example server will ensure client is capable of handling response

michel-laterman:
Fix example server to only set RemoteConfig, and ConnectionSettings if the agent capability flag is set. This removes excess logging from the client library.

### DIFF
--- a/internal/examples/server/data/agent.go
+++ b/internal/examples/server/data/agent.go
@@ -263,9 +263,9 @@ func (agent *Agent) processStatusUpdate(
 	}
 
 	configChanged := false
+	// Agent description is changed.
 	if agentDescrChanged {
-		// Agent description is changed.
-
+		// Only need to check if config has changed if the agent is able to accept a new config
 		if agent.hasCapability(protobufs.AgentCapabilities_AgentCapabilities_AcceptsRemoteConfig) {
 			// We need to recalculate the config.
 			configChanged = agent.calcRemoteConfig()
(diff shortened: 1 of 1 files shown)


internal/examples/server/data/agent.go
```diff
@@ -266,18 +266,23 @@ func (agent *Agent) processStatusUpdate(
 	if agentDescrChanged {
 		// Agent description is changed.
 
-		// We need to recalculate the config.
-		configChanged = agent.calcRemoteConfig()
+		if agent.hasCapability(protobufs.AgentCapabilities_AgentCapabilities_ReportsRemoteConfig) {
+			// We need to recalculate the config.
+			configChanged = agent.calcRemoteConfig()
```
tigrannajaryan, line 271: calcRemoteConfig() has nothing to do with AgentCapabilities_ReportsRemoteConfig. calcRemoteConfig() calculates the configuration _for_ the agent and that needs to happen regardless of whether the agent reports or doesn't report remote config status (which is what ReportsRemoteConfig indicates).

calcRemoteConfig() can be probably conditional to AcceptsRemoteConfig capability.

michel-laterman: Changed the check

internal/examples/server/data/agent.go
```diff
@@ -266,18 +266,23 @@ func (agent *Agent) processStatusUpdate(
...
+			if agent.Status.RemoteConfigStatus != nil {
+				configChanged = configChanged || bytes.Equal(agent.Status.RemoteConfigStatus.LastRemoteConfigHash, agent.remoteConfig.ConfigHash)
+			}
+		}
 
-		// And set connection settings that are appropriate for the Agent description.
-		agent.calcConnectionSettings(response)
+		if agent.hasCapability(protobufs.AgentCapabilities_AgentCapabilities_AcceptsOtherConnectionSettings) {
```
tigrannajaryan: AcceptsOtherConnectionSettings is for "other_connections" field. However calcConnectionSettings doesn't set other_connections, it sets "own_metrics". The capability corresponding to that is ReportsOwnMetrics.

tigrannajaryan (approved):

PR #484, opened 2026-01-07
Server uses sync.Pool for gzip.Writer allocations

michel-laterman:
Use of gzip.Writer pool for plain HTTP responses - gzip.Writers have a large memory footprint, if many agent clients clients were to respond at once a server could have an OOM crash. Use of the pool also reduces allocs.

```
go test -v -benchmem -benchtime 5s -bench=BenchmarkCompressGzip
...
goos: darwin
goarch: arm64
pkg: github.com/open-telemetry/opamp-go/server
cpu: Apple M3 Pro
BenchmarkCompressGzip
BenchmarkCompressGzip/with_pool
BenchmarkCompressGzip/with_pool-12         	  483432	     12155 ns/op	     113 B/op	       2 allocs/op
BenchmarkCompressGzip/no_pool
BenchmarkCompressGzip/no_pool-12           	   96456	     60972 ns/op	  813988 B/op	      19 allocs/op
```

### DIFF
--- a/server/serverimpl.go
+++ b/server/serverimpl.go
@@ -362,20 +362,6 @@ func (s *server) compressGzip(data []byte) ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
-func compressGzip(data []byte) ([]byte, error) {
-	var buf bytes.Buffer
-	w := gzip.NewWriter(&buf)
-	_, err := w.Write(data)
-	if err != nil {
-		return nil, err
-	}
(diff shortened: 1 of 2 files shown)


michel-laterman:
non gwPool changes: https://github.com/open-telemetry/opamp-go/pull/489

I'm going to rewrite this PR's history so that the 1st commit has both compress methods, and a benchmark for them, and the 2nd commit only has the (new) pool method.

PR #538, opened 2026-03-26
Restrict SetCapabilities after Start()

tigrannajaryan:
Only certain capabilities can be changed after Start(), see spec change: https://github.com/open-telemetry/opamp-spec/pull/306

### DIFF
--- a/client/clientimpl_test.go
+++ b/client/clientimpl_test.go
@@ -1928,9 +1928,8 @@ func TestSetCapabilities(t *testing.T) {
 			}
 		})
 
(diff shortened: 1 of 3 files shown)


tigrannajaryan:
I have closed the spec PR that attempted to describe this change in a formal way. I don't think we need that in the spec. See my reasoning here: https://github.com/open-telemetry/opamp-spec/pull/306#issuecomment-4360380653

I still believe having this restriction in the implementation is right. As new scenarios require other capability changes to be allowed opamp-go should work on supporting them, with corresponding testing verifying the implementation done.

Marking this as ready for review.

PR #544, opened 2026-04-01
Allow connection settings to be applied asynchronously

michel-laterman:
Expose SetConnectionsSettingsStatus method in client interface to allow implementations that use async approaches to connection management to succeed instead of automatically setting APPLIED.

This change technically breaks an assumption we had in the initial implementation and passes the responsibility for setting the connectionsettingsstatus to the callback implementation.

The `client/internal/receivedprocessor.go` still sets the status to APPLYING if any connection settings are sent (and the reports connections settings status capability is set),

- Relates: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46157

### DIFF
--- a/client/internal/receivedprocessor.go
+++ b/client/internal/receivedprocessor.go
@@ -1,7 +1,6 @@
 package internal
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"sync"
@@ -345,17 +344,4 @@ func (r *receivedProcessor) rcvCommand(ctx context.Context, command *protobufs.S
 	if command != nil {
 		r.callbacks.OnCommand(ctx, command)
 	}
-}
-
-// updateStoredConnectionSettingsStatus returns a bool of if status should replace oldStatus.
-// It's true if:
-// - no oldStatus
-// - hash changes
-// - status changes from APPLYING or UNSET
-// - status changes to FAILED
-func updateStoredConnectionSettingsStatus(oldStatus, status *protobufs.ConnectionSettingsStatus) bool {
-	return oldStatus == nil || !bytes.Equal(oldStatus.LastConnectionSettingsHash, status.LastConnectionSettingsHash) ||
-		oldStatus.Status == protobufs.ConnectionSettingsStatuses_ConnectionSettingsStatuses_APPLYING ||
-		oldStatus.Status == protobufs.ConnectionSettingsStatuses_ConnectionSettingsStatuses_UNSET ||
-		status.Status == protobufs.ConnectionSettingsStatuses_ConnectionSettingsStatuses_FAILED
 }
--- a/client/internal/receivedprocessor.go
+++ b/client/internal/receivedprocessor.go
@@ -275,34 +275,6 @@ func (r *receivedProcessor) rcvOpampConnectionSettings(ctx context.Context, sett
 		if err != nil {
 			r.logger.Errorf(ctx, "Failed to process OpAMPConnectionSettings: %v", err)
 		}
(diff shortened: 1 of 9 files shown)


michel-laterman:
@tigrannajaryan @evan-bradley, I'm addressing the changes needed in opamp-go for https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46157

This is a pretty fundamental change as it changes what sets the status from an internal detail (that assumes the callback is synchronous), to something that the connection settings callbacks needs to explicitly set

client/internal/receivedprocessor.go
```diff
@@ -275,34 +274,6 @@ func (r *receivedProcessor) rcvOpampConnectionSettings(ctx context.Context, sett
 		if err != nil {
 			r.logger.Errorf(ctx, "Failed to process OpAMPConnectionSettings: %v", err)
```
tigrannajaryan, line 275: I think we should report `ConnectionSettingsStatuses_FAILED` here if the `OnOpampConnectionSettings` returns an error.

If you look at the example the callback returns errors synchronously in some cases. I think reporting `ConnectionSettingsStatuses_FAILED` in such situation is the right thing to do.

internal/examples/agent/agent/agent.go
```diff
@@ -674,13 +682,26 @@ func (agent *Agent) tryChangeOpAMP(ctx context.Context, cert *tls.Certificate, t
 
 	if err := agent.connect(withTLSConfig(tlsConfig), withProxy(proxy)); err != nil {
 		agent.logger.Errorf(ctx, "Cannot connect after using new tls config: %s. Ignoring the offer\n", err)
+		if statusErr := agent.client.SetConnectionSettingsStatus(&protobufs.ConnectionSettingsStatus{
+			LastConnectionSettingsHash: hash,
+			Status:                     protobufs.ConnectionSettingsStatuses_ConnectionSettingsStatuses_FAILED,
+			ErrorMessage:               err.Error(),
+		}); statusErr != nil {
+			agent.logger.Errorf(ctx, "Failed to report connection settings status: %v", statusErr)
+		}
```
tigrannajaryan: Nit: this is a bit hard to read. Do you mind putting `protobufs.ConnectionSettingsStatus` message creation separately before `if`?

client/types/callbacks.go
```diff
@@ -92,7 +92,11 @@ type Callbacks struct {
 	//
 	// The Agent should process the offer by reconnecting the client using the new
 	// settings or return an error if the Agent does not want to accept the settings
-	// (e.g. if the TSL certificate in the settings cannot be verified).
+	// (e.g. if the TLS certificate in the settings cannot be verified).
+	//
+	// The caller is responsible for reporting the outcome via
+	// OpAMPClient.SetConnectionSettingsStatus with APPLIED or FAILED status.
+	// The client automatically sets APPLYING when the offer is received.
```
tigrannajaryan, line 99: I don't see where in implementation the client is automatically setting APPLYING. Can you tell which file/line numbers?

tigrannajaryan (commented):
@michel-laterman sorry for late review.

I think the change makes sense. I left a few comments/questions.

client/types/callbacks.go
michel-laterman: https://github.com/open-telemetry/opamp-go/blob/453b6b8fc612a1f6bea840b3ceb63e63f2881a6a/client/internal/receivedprocessor.go#L99-L108

tigrannajaryan (approved):

Issue #553, opened 2026-04-27
HTTP 401 responses could be retryable so `OnConnectFailed` fires


etgraylog:
## Expected Behavior

HTTP 401 (Unauthorized) responses could be treated as retryable with exponential backoff, similar to 429 and 503. The `OnConnectFailed` callback would then fire on each failed attempt, giving the application a chance to handle the authentication failure (e.g., refresh credentials, re-enroll, or exit for orchestrator restart).

This is a policy choice rather than a specified behavior. Stock opamp-go's treatment of 401 as non-retryable is a defensible position — it prevents a client from hammering an endpoint that has explicitly said "you're not authorized" — and leaves 401 handling entirely to the caller. Making 401 retryable with backoff is one reasonable alternative, best suited to deployments where the caller's `OnConnectFailed` logic actively drives recovery (credential refresh, re-enrollment, etc.) rather than treating the failure as terminal.

## Current Behavior

In `attemptRequest()` (`httpsender.go:280-304`), the response status code handling is:

```go
switch resp.StatusCode {
case http.StatusOK:
    // success
case http.StatusTooManyRequests, http.StatusServiceUnavailable:
    // retry with backoff
default:
    return requestResult{
        resp:  nil,
        err:   fmt.Errorf("invalid response from server: %d", resp.StatusCode),
        retry: false,  // <-- permanent failure
    }
}
```

HTTP 401 falls into the `default` case and returns `retry: false`. The current request attempt gives up, the polling loop moves on to the next cycle, sends the same request, gets another 401, and this pattern repeats indefinitely. The agent is alive and polling but has no automatic path to recovery.

## Possible Solution

Add `http.StatusUnauthorized` as a retryable status code:

```go
case http.StatusUnauthorized:
    _, _ = io.Copy(io.Discard, resp.Body)
    _ = resp.Body.Close()
    return requestResult{
        resp:  nil,
        err:   fmt.Errorf("server response code=%d (unauthorized)", resp.StatusCode),
        retry: true,
    }
```

This causes the retry loop to apply exponential backoff and fire `OnConnectFailed` on each attempt, giving the application a chance to take corrective action.

An alternative would be a new callback like `OnAuthFailure(ctx, error) bool` that lets the application decide whether to retry, but this is a larger API change.

### Tested via unit test

[...]

PR #556, opened 2026-04-29
Fix/401 retryable on connect failed

etgraylog:
Opening this PR to resolve the issue reported in https://github.com/open-telemetry/opamp-go/issues/553 .

### DIFF
--- a/client/internal/httpsender.go
+++ b/client/internal/httpsender.go
@@ -70,20 +70,42 @@ type HTTPSender struct {
 
 	// Processor to handle received messages.
 	receiveProcessor receivedProcessor
+
+	// HTTP response status codes that trigger a retry with backoff.
+	retryStatusCodes []int
+}
+
+// defaultRetryStatusCodes returns the HTTP response status codes that the
+// plain-HTTP transport retries by default.
+func defaultRetryStatusCodes() []int {
+	return []int{http.StatusTooManyRequests, http.StatusServiceUnavailable}
 }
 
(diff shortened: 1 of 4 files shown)


client/internal/httpsender.go
```diff
@@ -293,6 +293,17 @@ func (h *HTTPSender) attemptRequest(ctx context.Context, req *requestWrapper, cu
 			interval: retryInterval,
 		}
 
+	case http.StatusUnauthorized:
```
michel-laterman: I really don't think that this should be default behaviour for the agent.
I think adding a setting in StartSettings such as `RetryStatusCodes: []int{http.StatusTooManyRequests, http.StatusServiceUnavailable}` would be a better approach

etgraylog: Thanks for the review @michel-laterman. I went with your suggestion and made the retryable status codes configurable via StartSettings rather than hardcoding 401.

- New field: StartSettings.RetryStatusCodes []int
- When nil, the default []int{http.StatusTooManyRequests, http.StatusServiceUnavailable} is used — matches prior library behavior, so 401 is no longer retryable by default.
- A non-nil value (including an empty slice) fully replaces the default. Callers wanting the 401-retry behavior from the original PR opt in explicitly:
    ```go
      RetryStatusCodes: []int{
          http.StatusTooManyRequests,
          http.StatusServiceUnavailable,
          http.StatusUnauthorized,
      }
    ```
Also I merged upstream main to resolve conflicts with #566 and #570. Please take a look when you get a chance, thanks!

PR #560, opened 2026-05-11
Add a timeout setting for client HTTP requests

agoallikmaa:
This gives the ability to properly recover from cases where an intermediate proxy or server misbehaves by spuriously stopping responding on a connection without actually closing it, causing an indefinite hang. With a request timeout set, this can be recovered from, as it would just cause a failed request and then attempt to reconnect.

This was encountered in a real case, specifically due to a misbehaving proxy in that case. I've created a synthetic reproduction case demo that can shows the hang with request timeout unset, and the recovery with it set:
https://gist.github.com/agoallikmaa/fb8a8eb0e33ab8fab995132d2ce76250

The `receivedprocessor.go` change is just to fix current main branch build issue.

### DIFF
--- a/client/httpclient.go
+++ b/client/httpclient.go
@@ -51,6 +51,10 @@ func (c *httpClient) Start(ctx context.Context, settings types.StartSettings) er
 	// Add TLS configuration into httpClient
 	c.sender.AddTLSConfig(settings.TLSConfig)
 
+	if settings.HTTPRequestTimeout > 0 {
+		c.sender.SetRequestTimeout(settings.HTTPRequestTimeout)
+	}
+
 	if settings.EnableCompression {
 		c.sender.EnableCompression()
 	}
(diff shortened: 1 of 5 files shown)


tigrannajaryan (changes requested):
I don't think we should keep adding http client settings. A more manageable solution is to allow the Client to accept an instance of http.Client to use. The caller is then free to configure the http.Client as needed. We can probably add `Client *http.Client` setting to `StartSettings`.

One question to resolve is what we do for WebSocket Client. Do we ignore the `Client` setting?

@open-telemetry/opamp-go-approvers what do you think?

agoallikmaa:
That makes sense. Will change it to accept an instance of `http.Client` in settings, and leave WebSocket behavior untouched for now until further feedback.

tigrannajaryan:
> That makes sense. Will change it to accept an instance of `http.Client` in settings, and leave WebSocket behavior untouched for now until further feedback.

@agoallikmaa to avoid investing too much time prematurely, please feel free to post a draft initially (skip the tests, etc) so that we can review and decide if that's the desirable approach. 

cc @open-telemetry/opamp-go-approvers

PR #565, opened 2026-05-12
chore: Pass `downloadURL` to `UpdateContent`

dpaasman00:
This change is based on feedback left on this [PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47300#discussion_r3197221917) adding Package upgrades to the OpAMP Supervisor.

This adds a new parameter, `downloadURL`, to the `PacakgesStateProvider.UpdateContent` function. When the OpAMP client is handling a PackagesAvailable message, it retrieves the data found at `downloadURL` and places it in the `data io.Reader` parameter that is passed to `UpdateContent`. The issue with this is this function doens't know what kind of data is in this reader (e.g. if it's compressed or what archive type it is). Without this information, `UpdateContent` can't properly handle the data. By passing `downloadURL` we can inspect the url's suffix to infer the package type.

### DIFF
--- a/client/internal/inmempackagestore.go
+++ b/client/internal/inmempackagestore.go
@@ -68,7 +68,7 @@ func (l *InMemPackagesStore) FileContentHash(packageName string) ([]byte, error)
 	return l.fileHashes[packageName], nil
 }
 
-func (l *InMemPackagesStore) UpdateContent(_ context.Context, packageName string, data io.Reader, contentHash, signature []byte) error {
+func (l *InMemPackagesStore) UpdateContent(_ context.Context, packageName, _ string, data io.Reader, contentHash, signature []byte) error {
 	b, err := io.ReadAll(data)
 	if err != nil {
 		return err
--- a/client/internal/packagessyncer.go
+++ b/client/internal/packagessyncer.go
@@ -329,7 +329,7 @@ func (s *packagesSyncer) downloadFile(ctx context.Context, pkgName string, file 
 	defer detailsReporter.stop()
 
 	tr := io.TeeReader(resp.Body, detailsReporter)
-	err = s.localState.UpdateContent(ctx, pkgName, tr, file.ContentHash, file.Signature)
+	err = s.localState.UpdateContent(ctx, pkgName, file.GetDownloadUrl(), tr, file.ContentHash, file.Signature)
 	if err != nil {
 		return fmt.Errorf("failed to install/update the package %s downloaded from %s: %v", pkgName, file.DownloadUrl, err)
 	}
--- a/client/types/packagessyncer.go
+++ b/client/types/packagessyncer.go
@@ -77,6 +77,9 @@ type PackagesStateProvider interface {
 	// an error.
 	// Content hash must be updated if the data is updated without failure.
 	// The function must cancel and return an error if the context is cancelled.
+	// downloadURL is the URL the content was downloaded from. Implementations may
+	// inspect its path suffix (e.g. ".tar.gz", ".zip") to decide how to interpret or
+	// unpack the content.
 	UpdateContent(ctx context.Context, packageName, downloadURL string, data io.Reader, contentHash, signature []byte) error
 
 	// DeletePackage deletes the package from the Agent's local storage.
--- a/client/types/packagessyncer.go
+++ b/client/types/packagessyncer.go
@@ -77,7 +77,7 @@ type PackagesStateProvider interface {
 	// an error.
 	// Content hash must be updated if the data is updated without failure.
 	// The function must cancel and return an error if the context is cancelled.
-	UpdateContent(ctx context.Context, packageName string, data io.Reader, contentHash, signature []byte) error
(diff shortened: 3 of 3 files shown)


dpaasman00:
@tigrannajaryan Because the URL the package is hosted at can effectively be anything, what's the solution for if a URL doesn't contain the package type? I have limited knowledge of them, but if a user is using a signed CDN URL (seems like a valid option for an upgrade request), this solution breaks down.

douglascamata:
> what's the solution for if a URL doesn't contain the package type

@dpaasman00 one idea I had is to add one more option to figure out the format of the file: the `Content-Type` header from the request. Note that it won't solve everything because the header could be just `application/octet-stream` or even not present, depending on the server/cdn implementation.

A different and more complete idea, if adding a 3rd party dependency for this is acceptable, could be to use https://github.com/gabriel-vasile/mimetype/.

In both cases we would have to change more the signature of `UpdateContent` because the mimetype of the downloaded file is independent of the filename.

We could also combine these ideas into a tiered system where we first try the filename, then content-type header, then the `mimetype` package.

tigrannajaryan:
> The issue with this is this function doens't know what kind of data is in this reader (e.g. if it's compressed or what archive type it is).

@dpaasman00 why is it necessary to detect the kind of data? Why can't Collector Supervisor _require_ that the data is .tar.gz?

douglascamata:
@tigrannajaryan that's a big limitation, no? Why create this limitation if we have plenty of options to avoid it?

douglascamata:
The accepted file formats are not part of the OpAMP spec, so I don't think it's a good idea to allow only `.tar.gz`. This is not only for the Supervisor, by the way. 

I think there should be some flexibility here, specially when thinking about 3rd party packages and forward compatibility with format changes.

dpaasman00:
> why is it necessary to detect the kind of data? Why can't Collector Supervisor _require_ that the data is .tar.gz?

@tigrannajaryan I think it's valuable for this to be flexible.

The current happy path we're solving for is handling releases generated by the [OTel releases repo](https://github.com/open-telemetry/opentelemetry-collector-releases) which offers a `.tar.gz` format for all versions of the collector. However not every collector managed by the supervisor will come from this repository. As a result I think it's very important that this be flexible so upgrades are compatible with a variety of different ways a user could package their collector.

My original approach required configuring the expected archive format in the supervisor configuration. If dynamic detection becomes too convoluted, I think this is always an option to go back to.

Regardless of how the package format is detected, each potential packaging format will require specific "un-packaging" logic. My current plan/implementation for upgrades in the supervisor will allow this to be extendable without major refactoring. However it will still require a PR to add the new logic. Maybe instead, in conjunction with my work adding collector extensions to the supervisor, we can offload this to a new class of extensions specific to the supervisor that handle unarching?

dpaasman00:
> > what's the solution for if a URL doesn't contain the package type
> 
> @dpaasman00 one idea I had is to add one more option to figure out the format of the file: the `Content-Type` header from the request. Note that it won't solve everything because the header could be just `application/octet-stream` or even not present, depending on the server/cdn implementation.
> 
> A different and more complete idea, if adding a 3rd party dependency for this is acceptable, could be to use https://github.com/gabriel-vasile/mimetype/.
> 
> In both cases we would have to change more the signature of `UpdateContent` because the mimetype of the downloaded file is independent of the filename.
> 
> We could also combine these ideas into a tiered system where we first try the filename, then content-type header, then the `mimetype` package.

@douglascamata I think this is definitely worth exploring. I would be curious what other's opinions are about adding the 3rd party dependency.

tigrannajaryan:
> However not every collector managed by the supervisor will come from this repository. As a result I think it's very important that this be flexible so upgrades are compatible with a variety of different ways a user could package their collector.

@dpaasman00 do you have a specific example of a package that Otel Supervisor needs to support that is not .tar.gz currently or this is a hypothetical possibility?

dpaasman00:
@tigrannajaryan  `.zip` packages are one example, for users packaging their collector distributions on Windows in a `.zip` archive rather than `.tar.gz`. A specific example of this is the windows package we have at Bindplane for our version of the OCB collector. 

Another possible example is users not compressing their packages, for whatever reason that may be, such as just `.tar`. This would just be a hypothetical, like the CDN URL example, but I think it's important to consider them and make it easy in the future to support new formats.

tigrannajaryan:
> `.zip` packages are one example, for users packaging their collector distributions on Windows in a `.zip` archive rather than `.tar.gz`. A specific example of this is the windows package we have at Bindplane for our version of the OCB collector.

@dpaasman00 Thanks. This is useful. From what I see the default release on Window is `.msi`. This means for you particular situation you do need to be able to tell which one is downloaded, the msi or zip.

I think that gives me the datapoint I was looking for (that the Supervisor needs to make this call at runtime).

I think we can expose the URL and the response headers (Content-Type and other) for Supervisor to use if you think there is a good heuristic to use.

Detection based on content using https://github.com/gabriel-vasile/mimetype/ I think is a step too far.

tigrannajaryan (approved):

PR #566, opened 2026-05-18
Add HTTP client to StartSettings

agoallikmaa:
In https://github.com/open-telemetry/opamp-go/pull/560 it was proposed that HTTP client be added to StartSettings instead of adding any further individual HTTP client settings to there.

This is a draft for doing that, where I've made the minimal amount of changes to achieve that for now (also haven't added additional tests yet). There are several approaches that could be taken from here, but I'd like feedback on the following points before doing so:

* Should the specified client also be used for WebSocket?
* Should existing settings that the library user can apply to the client be removed, or kept for backwards compatibility/convenience?
* Would it make sense to create separate StartSettings types for WebSocket and HTTP request based client? (with common part for use by clientcommon)

### DIFF
--- a/client/internal/httpsender_test.go
+++ b/client/internal/httpsender_test.go
@@ -23,6 +23,12 @@ import (
 	"github.com/open-telemetry/opamp-go/protobufs"
 )
 
+func newTestHTTPSender() *HTTPSender {
+	sender := NewHTTPSender(&sharedinternal.NopLogger{})
+	sender.SetHTTPClient(&http.Client{})
+	return sender
+}
+
 func TestHTTPSenderRetryForStatusTooManyRequests(t *testing.T) {
 	var connectionAttempts int64
 	srv := StartMockServer(t)
@@ -38,7 +44,7 @@ func TestHTTPSenderRetryForStatusTooManyRequests(t *testing.T) {
 	}
 	ctx, cancel := context.WithCancel(context.Background())
 	url := "http://" + srv.Endpoint
-	sender := NewHTTPSender(&sharedinternal.NopLogger{})
+	sender := newTestHTTPSender()
 	sender.NextMessage().Update(func(msg *protobufs.AgentToServer) {
 		msg.AgentDescription = &protobufs.AgentDescription{
 			IdentifyingAttributes: []*protobufs.KeyValue{{
@@ -66,7 +72,7 @@ func TestHTTPSenderRetryForStatusTooManyRequests(t *testing.T) {
 }
 
 func TestHTTPSenderSetHeartbeatInterval(t *testing.T) {
-	sender := NewHTTPSender(&sharedinternal.NopLogger{})
+	sender := newTestHTTPSender()
 
 	// Default interval should be 30s as per OpAMP Specification
 	assert.Equal(t, (30 * time.Second).Milliseconds(), sender.pollingIntervalMs.Load())
@@ -169,7 +175,7 @@ func TestAddTLSConfig(t *testing.T) {
 
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
-			sender := NewHTTPSender(&sharedinternal.NopLogger{})
+			sender := newTestHTTPSender()
 			sender.client.Transport = tc.existingTransport
 
 			sender.AddTLSConfig(tc.tlsConfig)
@@ -233,7 +239,7 @@ func TestHTTPSenderRetryForFailedRequests(t *testing.T) {
 	}
 	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
 	url := "http://" + address
-	sender := NewHTTPSender(&sharedinternal.NopLogger{})
+	sender := newTestHTTPSender()
 	sender.NextMessage().Update(func(msg *protobufs.AgentToServer) {
 		msg.AgentDescription = &protobufs.AgentDescription{
 			IdentifyingAttributes: []*protobufs.KeyValue{{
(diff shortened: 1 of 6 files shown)


tigrannajaryan:
> * Should the specified client also be used for WebSocket?

Ideally we would want it, but I think WebSocket library we use doesn't allow supplying a client. Can you please check it? If we cannot do it then this setting will be only applicable to plain http implementation.

> * Should existing settings that the library user can apply to the client be removed, or kept for backwards compatibility/convenience?

I think we should keep them for now. We can declare them deprecated and can probably remove them after a while, since we are not at 1.0 yet.

> * Would it make sense to create separate StartSettings types for WebSocket and HTTP request based client? (with common part for use by clientcommon)

This is worth exploring. It may be interesting if we are anticipating more settings to be different in the future. There is benefit in having one "settings" struct but if the functionalities are going to diverge then that commonness becomes more fictional.

agoallikmaa:
> > * Should the specified client also be used for WebSocket?
> 
> Ideally we would want it, but I think WebSocket library we use doesn't allow supplying a client. Can you please check it? If we cannot do it then this setting will be only applicable to plain http implementation.

Checked and it doesn't seem like passing http client is applicable there. In theory a `websocket.Dialer` instance could be passed that is specific to the gorilla/websocket library.

> > * Should existing settings that the library user can apply to the client be removed, or kept for backwards compatibility/convenience?
> 
> I think we should keep them for now. We can declare them deprecated and can probably remove them after a while, since we are not at 1.0 yet.
> 
> > * Would it make sense to create separate StartSettings types for WebSocket and HTTP request based client? (with common part for use by clientcommon)
> 
> This is worth exploring. It may be interesting if we are anticipating more settings to be different in the future. There is benefit in having one "settings" struct but if the functionalities are going to diverge then that commonness becomes more fictional.

This would require some code changes for users of the library (to pass the protocol specific type instead), but is a more simple change to adapt to compared to potentially removing fields. Is it acceptable to make this change right now without regard for which version it lands in? Doing this now would help make it more clear that the deprecation markers for some fields are only applicable for HTTP as the relevant fields would be split into the protocol-specific structs, and not kept in the embedded common struct.

michel-laterman:
I think it makes sense to introduce the http.Client and mark the other attributes as deprecated.

@tigrannajaryan, one thing we can consider is changing the websocket library. (gorilla/websocket has had a new pr merged in over a year, and the latest release was in 2024)

The coder/websocket library contains the option to provide an http.Client when dialing: https://pkg.go.dev/github.com/coder/websocket#DialOptions

This would let us keep the same start settings for http + websocket connections.

agoallikmaa:
@michel-laterman That sounds reasonable, but perhaps it could be done in stages to make it apply for just the HTTP server for now, and a separate PR later for swapping out websocket? In this case, not splitting the start settings would make sense and this PR could just be solely about adding the HTTP client, basically what the draft already does.

@tigrannajaryan Do you see any issues with going ahead with practically just the current changes for this specific PR (I already made some doc and test changes to make it more complete)? The ability to specify the client gives the ability to fix some actual issues already, for example the one in https://github.com/open-telemetry/opamp-go/pull/560 which this draft was created as an alternative to.

I have converted this to a PR since I think the bigger potential changes (either websocket library replace, or split of settings) would make sense as a separate change later.

tigrannajaryan:
@agoallikmaa yes, we can go ahead with this PR. Change of WebSocket library and depreciation should be separate PRs.

Can you please create an issue to record which StartSettings options we want to deprecate after adding the `Client` option, so that we don't forget.

tigrannajaryan:
Please fix the build.

michel-laterman:
I've created a draft just to see the changes required to swap the client code: https://github.com/open-telemetry/opamp-go/pull/576

michel-laterman (approved):

client/internal/httpsender_test.go
```diff
@@ -23,6 +23,12 @@ import (
 	"github.com/open-telemetry/opamp-go/protobufs"
 )
 
+func newTestHTTPSender() *HTTPSender {
+	sender := NewHTTPSender(&sharedinternal.NopLogger{})
+	sender.SetHTTPClient(&http.Client{})
```
tigrannajaryan:
```suggestion
	sender.SetHTTPClient(utils.NewHttpClient())
```
I think this better preserves the previous behavior of the tests.

client/types/startsettings.go
```diff
@@ -18,18 +18,28 @@ type StartSettings struct {
...
+	Client *http.Client
+
 	// Optional function that can be used to modify the HTTP headers
 	// before each HTTP request.
 	// Can modify and return the argument or return the argument without modifying.
 	HeaderFunc func(http.Header) http.Header
 
 	// Optional TLS config for HTTP connection.
+	//
+	// Deprecated for HTTP transport: configure TLS on Client instead.
```
tigrannajaryan: I suggest that we don't do deprecations for now, since `Client` only is used for plain http. The "Deprecated" notice essentially only applies to plain http, but it will appear unconditionally, even when one is using WebSocket implementation and will be confusing.

I suggest that we first swap the underying WebSocket implementation, which will allow us to use `Client` option with WebSocket as well, and then after that we can deprecate these 3 options.

@michel-laterman what do you think?

client/internal/httpsender_test.go
agoallikmaa: Done.

client/types/startsettings.go
agoallikmaa: Removed the deprecations.

tigrannajaryan (approved):

PR #570, opened 2026-05-27
Add transport message size limits

pellared:
Per https://github.com/open-telemetry/opamp-spec/pull/346

### DIFF
--- a/client/internal/httpsender.go
+++ b/client/internal/httpsender.go
@@ -397,6 +397,9 @@ func (h *HTTPSender) readResponseBody(resp *http.Response) ([]byte, error) {
 	}
 	defer closeBody()
 
+	// Do not drain oversized responses after the limit is hit. Reading to EOF
+	// would preserve HTTP/1 keep-alive, but would also let a peer force
+	// unbounded network and decompression work after MaxMessageSize is exceeded.
 	return internal.ReadAllLimited(body, h.maxMessageSize, "response body")
 }
 
--- a/client/internal/httpsender.go
+++ b/client/internal/httpsender.go
@@ -64,6 +64,7 @@ type HTTPSender struct {
 	callbacks          types.Callbacks
 	pollingIntervalMs  atomic.Int64
 	compressionEnabled bool
+	maxMessageSize     int64
 
 	// Headers to send with all requests.
 	getHeader func() http.Header
@@ -81,6 +82,7 @@ func NewHTTPSender(logger types.Logger) *HTTPSender {
 		client:       utils.NewHttpClient(),
 	}
 	h.pollingIntervalMs.Store(defaultPollingIntervalMs)
+	h.maxMessageSize = internal.DefaultMaxMessageSize
 	// initialize the headers with no additional headers
 	h.SetRequestHeader(nil, nil)
 	return h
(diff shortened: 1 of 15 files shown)


client/internal/httpsender.go
```diff
@@ -363,14 +371,51 @@ func (h *HTTPSender) prepareRequest(ctx context.Context) (*requestWrapper, error
...
+}
+
+func (h *HTTPSender) readResponseBody(resp *http.Response) ([]byte, error) {
+	body, closeBody, err := h.responseBodyReader(resp)
+	if err != nil {
+		return nil, err
+	}
+	defer closeBody()
+
+	return internal.ReadAllLimited(body, h.maxMessageSize, "response body")
```
tigrannajaryan, line 403: I think we need to read the entire response to enable connection reuse. Unclear to me if that's the desirable behavior in this case or we want to avoid reading past the limit and instead prefer the connection to be dropped and a new one created. Thoughts?
The same question applies to reading of the request body by the server.

pellared: This applies mostly for `http.Client` (reading the response). But it should not be a problem doing it for the server as well.

Some references:
- https://pkg.go.dev/net/http#Client.Do
- https://goperf.dev/02-networking/efficient-net-use/?h=body#the-hidden-complexity-behind-a-simple-http-call

pellared: Actually, after a few iteration with Codex I think it may make sense to actually not drain the Body. 

>  full draining would let a peer force unbounded network/decompression work

I added some comments https://github.com/open-telemetry/opamp-go/pull/570/commits/ad1e48a9b9cb2a10df9a6ee3007e3caa117edab1

The good think is that these are internals and we can change it later if we think this is a wrong decision (mostly for the client).

internal/limits.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	return data, nil
+}
+
+func CopyDiscardLimited(r io.Reader, limit int64, kind string) error {
+	if limit < 0 {
+		_, err := io.Copy(io.Discard, r)
+		return err
+	}
+	n, err := io.Copy(io.Discard, io.LimitReader(r, limit+1))
```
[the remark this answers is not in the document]
pellared: https://github.com/open-telemetry/opamp-go/pull/570/commits/e2995597f410172f19b31ad583e5781007738df2

internal/limits.go
```diff
@@ -0,0 +1,72 @@
...
+		Kind:  kind,
+		Limit: limit,
+	}
+}
+
+func ReadAllLimited(r io.Reader, limit int64, kind string) ([]byte, error) {
+	if limit < 0 {
+		return io.ReadAll(r)
+	}
+	data, err := io.ReadAll(io.LimitReader(r, limit+1))
```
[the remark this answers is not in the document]
pellared: https://github.com/open-telemetry/opamp-go/pull/570/commits/e2995597f410172f19b31ad583e5781007738df2

server/serverimpl.go
```diff
@@ -421,6 +440,12 @@ func (s *server) handlePlainHTTPRequest(req *http.Request, w http.ResponseWriter
 		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
+	if err := internal.CheckSizeLimit(int64(len(bodyBytes)), s.settings.MaxMessageSize, "response body"); err != nil {
+		s.logger.Errorf(req.Context(), "Cannot send HTTP response: %v", err)
+		connectionCallbacks.OnMessageResponseError(agentConn, response, err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
```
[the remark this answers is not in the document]
pellared: In `server/serverimpl.go` (line 438), `bodyBytes` is reassigned from `proto.Marshal(response)` before the response limit check at `server/serverimpl.go` (line 443), so the check is using the marshaled response size, not the request body size.

client/internal/httpsender.go
tigrannajaryan: You are right. Somehow I thought this is client side.

pellared:
@open-telemetry/opamp-go-maintainers , PTAL

tigrannajaryan (approved):

tigrannajaryan:
Thanks @pellared

PR #578, opened 2026-06-08
examples: Add `ConfigCalculator` interface and example

ArthurSens:
## Summary

This adds a configurable remote config calculation layer to the example OpAMP server and uses it to support Prometheus receiver sharding across connected example agents.

Agents can now report additional non-identifying attributes via the example agent CLI/env config, and implementors of `ConfigCalculator` can use them to apply any custom logic they need. The example also adds an implementation of `ConfigCalculator` that groups Collectors by the `prometheus-cluster` attribute and manages the sharding configuration for the whole fleet when an agent connects or disconnects.

### DIFF
--- a/internal/examples/agent/agent/agent.go
+++ b/internal/examples/agent/agent/agent.go
@@ -60,9 +60,10 @@ type Agent struct {
 	client client.OpAMPClient
 	logger types.Logger
 
-	agentType    string
-	agentVersion string
-	instanceId   uuid.UUID
+	agentType       string
+	agentVersion    string
+	instanceId      uuid.UUID
+	extraAttributes map[string]string
 
 	agentConfig *config.AgentConfig
 
@@ -128,6 +129,13 @@ func WithAgentVersion(s string) Option {
 func WithInstanceID(id uuid.UUID) Option {
 	return func(agent *Agent) {
 		agent.instanceId = id
+	}
+}
+
+// WithExtraAttributes sets additional non-identifying attributes reported by the agent.
+func WithExtraAttributes(attrs map[string]string) Option {
+	return func(agent *Agent) {
+		agent.extraAttributes = attrs
 	}
 }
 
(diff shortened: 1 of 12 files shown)


michel-laterman:
Hi @ArthurSens, 

The point of the example agent/server is to show a minimal poc for the opamp protocol. I don't think that prometheus specific belongs here.

ArthurSens:
Hi Michel, yes, I understand the server is supposed to be an example. I opened this PR after a [brief discussion in Slack](https://cloud-native.slack.com/archives/C02J58HR58R/p1780768182085439) with @tigrannajaryan. The code added is definitely not production-ready; it is bare-minimum code to showcase more advanced features that can be done with an OpAMP, e.g., keeping tabs on the number of instances in the fleet and assigning stable numbers to each instance, changing received config before making it effective, reconfiguring the whole fleet once an instance joins/disconnect etc etc.

That said, happy to make adjustments to the PR to surface these advanced capabilities more and don't make this PR look like "This is how you support Prometheus".

tigrannajaryan:
@ArthurSens thanks for creating the PR. For the demo I aim for educational code and I was envisioning some sort of decoupled code that shows a case of special business logic for config, with Prometheus-specific config being an example of it. 

AI's attempt is not exactly what I was imagining, however I am not entirely sure how to structure it myself.

I don't want to waste any more of your time. I will think about it to see if I can come up with a way good way to show the idea.

Please feel free to close the PR, my understanding is it is not blocking you.

ArthurSens:
Yes, this isn't blocking us. 

The example server will be super useful, though, since we're building a blueprint dedicated to managing a Prometheus-based collector architecture(https://github.com/open-telemetry/sig-end-user/issues/339). Our current suggestion is to use an opamp-server capable of the features in this PR.

I'm happy to collaborate here so the blueprint has a nice example to show :)

tigrannajaryan:
I think a possible approach can be following:

- On the Agent side in `internal/examples/agent` package instead of hard-coding support for AGENT_PROMETHEUS_CLUSTER env var and "prometheus.cluster" attribute, implement ability to accept any additional non-identifying attribute via a more generic env var (e.g. AGENT_EXTRA_ATTRIBUTES or something like that to contain key/value pairs)
- On the Server side introduce a concept of `ConfigCalculator` interface that is pluggable at runtime. Any number of `ConfigCalculator` implementations can be registered at the server startup and `ConfigCalculator.Calculate()` method is called in `Agent.calcRemoteConfig()` to additionally mutate the `configBody`. `ConfigCalculator.AgentsChanged()` or similar is called when a new agent is added or removed so that total number can be tracked.
- An implementation of `ConfigCalculator` called `PrometheusShardingCalculator` contains the special logic.

tigrannajaryan:
Thanks @ArthurSens for updating the PR. I like this approach, nicely decoupled the Prometheus-specific logic from the core functionality.

@open-telemetry/opamp-go-approvers what do you think? Worth having this in the example?

andykellr:
In the abstract I think it is great to have an example server that does something useful and interesting, but `prometheus_sharding.go` could use a comment at the top explaining what it does and some comments throughout the code would be really helpful for someone trying to learn from this example implementation.

tigrannajaryan:
@ArthurSens looks like we have a consensus this is a good thing to have in the demo. If you could add the comments Andy is asking about and resolve the conflicts, we can get move this forward.

juandemanjon:
This is another example of the benefits of having an opamp-contrib repo where users can extend the server/agent features as they need

juandemanjon:
If we want to follow the OpenTelemetry model, I suggest naming ConfigCalculator as ConfigProcessor

ConfigProcessors transform, filter, and enrich config data as it flows through the OpAMP pipeline

PR #587, opened 2026-06-18
Fix build broken by #570

tigrannajaryan:
Fixed httpSender creation in the tests and OnRequest callback setting.

Github action didn't catch this because we didn't rebase from latest main.


tigrannajaryan:
FYI @pellared

andykellr (approved):

PR #588, opened 2026-06-20
Expose BackoffPolicy in StartSettings

lokesh-balla:
Exposing BackoffPolicy in StartSettings in to allow control over reconnects and retries to the user.

### DIFF
--- a/client/clientimpl_test.go
+++ b/client/clientimpl_test.go
@@ -12,6 +12,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -93,6 +94,18 @@ func TestConnectInvalidURL(t *testing.T) {
 
 func eventually(t *testing.T, f func() bool) {
 	assert.Eventually(t, f, 5*time.Second, 10*time.Millisecond)
+}
+
+// mockBackoffPolicy is a BackoffPolicy that returns a fixed interval and records
+// how many times NextBackOff has been called. Used in tests.
+type mockBackoffPolicy struct {
+	interval time.Duration
+	calls    atomic.Int64
+}
+
+func (p *mockBackoffPolicy) NextBackOff() time.Duration {
+	p.calls.Add(1)
+	return p.interval
 }
 
 type mockServerMessageHandler func(msg *protobufs.AgentToServer) *protobufs.ServerToAgent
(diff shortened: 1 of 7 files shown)


client/types/startsettings.go
```diff
@@ -89,4 +90,7 @@ type StartSettings struct {
 	// If nil, the default reporter interval (10s) will be used.
 	// If specified a minimum value of 1s will be enforced.
 	DownloadReporterInterval *time.Duration
+
+	// Optional ExponentialBackOffOpts to pass options to exponential backoff retries
+	ExponentialBackOffOpts []backoff.ExponentialBackOffOpts
```
tigrannajaryan: wsClient.ensureConnected is an internal, undocumented function. This setting controls that undocumented function and it is not clear what exactly is the expected behavior here, how ExponentialBackOffOpts impact the behavior of the Client.

We should not have this option unless the behavior of the Client is clearly documented.

Blocking, this won't be accepted in the current form.

tigrannajaryan (dismissed):

client/types/startsettings.go
lokesh-balla: [image](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#restoring-websocket-connection)

I believe we have backoff documented in the spec.

Is the issue that I am exposing the backoff library options directly? We could have a struct that takes initialInterval, RandomizationFactor, multiplier, maxInterval and maxElapsed, would that work?

tigrannajaryan: >I believe we have backoff documented in the spec.

We should not expect users of opamp-go to read the spec and know it in details. Settings should be documented in a way that reading only opamp-go docs is enough to understand what a particular setting controls.

>Is the issue that I am exposing the backoff library options directly? We could have a struct that takes initialInterval, RandomizationFactor, multiplier, maxInterval and maxElapsed, would that work?

Yes, that's one concern, since it locks opamp-go's API into using the specific version of specific backoff library. Instead of leaking cenkalti/backoff API, one approach is to have a setting that allows to provide a backoff behavior via an interface with NextBackOff() method that can be satisfied by a user-constructed instance of cenkalti/backoff. (however, need to figure out what to do with MaxElapsedTime).

Additionally, I would like to see 2 things:
- A concise description of what exactly the backoff setting controls.
- We should use the same setting for http transport. I do not see a reason to only allow this control for WebSocket transport.

lokesh-balla: I exposed a Interface now for BackoffPolicy for both http and ws with NextBackOff() and without any Stop() since we want it to run forever.


> (however, need to figure out what to do with MaxElapsedTime)

This for default case where the user does not set anything we will do that same thing and set it to 0 so its infinite and in case of user configured if the NextBackOff ever gives any value which is less than 0 then we can default to defaultMaxInterval.

client/types/startsettings.go
```diff
@@ -8,6 +8,19 @@ import (
 	"github.com/open-telemetry/opamp-go/protobufs"
 )
 
+// BackoffPolicy controls the delay between consecutive connection or request
+// retry attempts. The client calls NextBackOff after each failed attempt to
+// determine how long to wait before the next one.
+//
+// A nil BackoffPolicy causes the client to use a default exponential backoff
+// that retries indefinitely (initial interval 500ms, multiplier 1.5, max
+// interval 60s, no elapsed-time limit).
```
tigrannajaryan: Let's remove the specific numbers of what the defaults are. I don't want to make the defaults a documented public API contract.
```suggestion
// that retries indefinitely.
```

tigrannajaryan (commented):
I like this approach.

lokesh-balla:
@tigrannajaryan I missed a case where the BackOffPolicy object wont reset for every call. 

we can solve this either by exposing the Reset() method in interface or we can change the struct to take a func() which returns the BackOffPolicy. 

I am leaning towards the func() approach, what would you suggest?

client/types/startsettings.go
```diff
@@ -8,6 +8,19 @@ import (
 	"github.com/open-telemetry/opamp-go/protobufs"
 )
 
+// BackoffPolicy controls the delay between consecutive connection or request
+// retry attempts. The client calls NextBackOff to determine how long to wait
+// before the next one.
+type BackoffPolicy interface {
+	// NextBackOff returns the duration to wait before the next retry.
+	// A negative return value causes the client to use the defaultMaxInterval interval.
```
tigrannajaryan: Why is this necessary? Is there some situation where we want to supply a policy that ends up using DefaultMaxInterval, but we don't know what the value of DefaultMaxInterval is, so that we can supply a correct implementation of policy ourselves?

This seems unnecessary to me, but I may be missing something.

Unless there is a good reason I think a better approach is to make this an error situation and fail/stop the retry loop and log an error.

tigrannajaryan:
> I am leaning towards the func() approach, what would you suggest?

Yes, I think this works.

client/clientimpl_test.go
```diff
@@ -3207,3 +3220,67 @@ func TestSetConnectionSettingsStatusAsync(t *testing.T) {
 		require.NoError(t, err)
 	})
 }
+
+// TestClientCenkaltiBackoffPolicy verifies that a cenkalti/backoff
```
tigrannajaryan, line 3224: We are currently on an older version of cenkalti/backoff. Can you also verify that the latest version of cenkalti/backoff also satisfies BackoffPolicy interface?

No need to write a test, just confirm by reading cenkalti/backoff API docs to make sure our interface is future-proof, so that it works fine if someone uses a new version to implement their policy.

lokesh-balla: I checked for v7 (https://github.com/cenkalti/backoff/blob/v7/exponential.go#L83) and i think we are good here.

client/types/startsettings.go
lokesh-balla: There was no good reason for it. I just handled it because negative values are a possibility based on the return type time.Duration which is a int64. I changed it to return a error now for such cases.

client/wsclient.go
```diff
@@ -305,16 +311,24 @@ func (c *wsClient) tryConnectOnce(ctx context.Context) (retryAfter sharedinterna
...
-	// Make ticker run forever.
-	infiniteBackoff.MaxElapsedTime = 0
+	var bpolicy types.BackoffPolicy
+	if c.backoffPolicy != nil {
+		bpolicy = c.backoffPolicy()
+	} else {
+		b := backoff.NewExponentialBackOff()
+		b.MaxElapsedTime = 0
+		bpolicy = b
+	}
```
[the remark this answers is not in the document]
lokesh-balla: @tigrannajaryan should we handle this or just let it panic assuming its the user reponsibity to have a proper func here? If we default this case to use our exponentialBackoff when the function returns nil then that behaviour would be hidden to the user. 

may be the best we can do here would be check if the backoffPolicy and return of backoffPolicy() are not nil in Start method and error out directly there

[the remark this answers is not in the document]
tigrannajaryan: I think it is fine to panic on nil.

tigrannajaryan (approved):

tigrannajaryan:
Thank for your contribution and for your patience @lokesh-balla

PR #599, opened 2026-07-13
Message Attestation: Client + Server prototype

liustanley:
## Overview

> **Draft for early feedback.** This implements the Message Attestation proposal in [opamp-spec#333](https://github.com/open-telemetry/opamp-spec/pull/333), which is still under review. Opening as a draft so the implementation can be reviewed alongside the spec; details may change as the spec is finalized.

Implements Message Attestation. Each substantive `ServerToAgent` is wrapped in a `SignedServerToAgent` envelope carrying a detached X.509 signature (empty heartbeat responses are exempt — see below). Agents that opt in can verify that responses originate from an authorised distribution server and have not been tampered with in transit — even when TLS is terminated at a load balancer.

**Fully opt-in and backward-compatible.** Agents and servers that do not configure attestation use the existing wire format unchanged.

This PR contains the core implementation plus focused unit tests for the heartbeat exemption — no examples. See #600 for a runnable end-to-end example.

### Recent additions

- **Heartbeat response exemption.** On an attested connection the Server MAY send an empty heartbeat response (only `instance_uid` set) unsigned; the Agent accepts that exact shape via a structural, default-deny check and rejects any other unsigned message. This avoids a signing round-trip for an empty keepalive on every HTTP poll.
- **Multi-SAN signing certificates (OpAMP Gateway).** A signing leaf MAY carry multiple SAN entries (e.g. a gateway/proxy host plus the origin server host), so signatures from one key are accepted by Agents connecting through either host. The Agent's hostname match is never relaxed — a host absent from the SAN set still fails.

---

### Server — `server.Settings`

```go
type Settings struct {
    // ... existing fields ...

    // PayloadSigner produces detached signatures over every outbound
    // ServerToAgent when an Agent has declared
    // AgentCapabilities_RequiresPayloadTrustVerification. When set, the
    // server automatically advertises
    // ServerCapabilities_OffersPayloadTrustVerification on all responses.
    // Nil disables attestation entirely (default, backward-compatible).
    PayloadSigner signing.Signer
}
```

### Client — `client/types.StartSettings`

```go
type StartSettings struct {
    // ... existing fields ...

    // PayloadTrustProvider opts the Agent in to payload trust verification
    // and selects the trust mode. Nil (default) keeps the standard OpAMP
    // wire path with no verification. When set, the client automatically
    // sets AgentCapabilities_RequiresPayloadTrustVerification.
    //
    // Construct one with:
    //   signing.FixedAnchor(verifier) — a fixed, pre-provisioned trust anchor
    //   signing.TOFUAnchor(store)     — Trust On First Use enrollment
    //
    // Additional trust modes can be added as new PayloadTrustProvider
    // implementations without changing this field.
    PayloadTrustProvider signing.PayloadTrustProvider
}
```

---

## `signing` package — interfaces

```go
// Signer is implemented by the OpAMP server (or its delegate).
type Signer interface {
    // Sign returns the signature, the exact bytes signed, and the chain that
    // anchors them as one SignResult. Bundling them in a single call means the
    // chain provably anchors the certificate that produced the signature (a
    // separate lookup could race a rotation), and lets signers that mint a
    // certificate only as a side effect of signing implement the interface —
    // they have no standing chain to report ahead of time.
    Sign(ctx context.Context, payload []byte) (SignResult, error)
}

// SignResult is the output of one signing operation. Bundling all three fields
// in a single return keeps the chain and signature consistent, and is the
// extension point for future signing metadata (cert id, leaf NotAfter, ...).
type SignResult struct {
    Payload   []byte   // exact bytes signed, transmitted as SignedServerToAgent.payload
    Signature []byte   // detached signature over Payload
    ChainDER  [][]byte // signing chain in DER, first intermediate -> leaf (root excluded)
}

// TrustAnchorProvider is an optional extension of Signer for TOFU enrollment.
// The server checks for this interface via type assertion.
type TrustAnchorProvider interface {
    TrustAnchorPEM(ctx context.Context) ([]byte, error)
}

// PayloadTrustProvider is the single client-side entry point for opting in to
// payload trust verification. FixedAnchor and TOFUAnchor are the provided
// implementations; new trust modes are added as new implementations without
// changing the StartSettings surface.
type PayloadTrustProvider interface {
    // Verifier returns the Verifier the Agent uses to validate the server's
    // trust chain and per-message signatures. It returns (nil, nil) to signal
    // TOFU mode, in which case the provider MUST also implement TOFUEnroller.
    Verifier() (Verifier, error)
}

// TOFUEnroller is an optional interface a PayloadTrustProvider MAY satisfy to
// support Trust On First Use enrollment. The client detects it via a type
// assertion when Verifier returns no verifier.
type TOFUEnroller interface {
    // Enroll persists the trust anchor delivered on first connection and
    // returns the Verifier built from it. MUST be idempotent (write-once) so a
    // reconnecting agent cannot replace a valid anchor.
    Enroll(anchorPEM []byte) (Verifier, error)
}

// Verifier is implemented by the OpAMP agent.
type Verifier interface {
    // ValidateChain performs RFC 5280 path validation of the delivered chain
    // against the pre-configured trust anchor, and binds the signing identity
    // to the connected server by requiring the leaf SAN to cover dnsName.
    // dnsName MUST be non-empty — hostname verification is centralized here
    // (an empty name makes crypto/x509 skip the SAN check), so callers do not
    // perform a separate hostname check. On success it returns a
    // VerifiedCertificate.
    ValidateChain(ctx context.Context, chainDER [][]byte, now time.Time, dnsName string) (*VerifiedCertificate, error)

    // Verify checks a detached signature using the VerifiedCertificate's leaf.
    // It first re-confirms the chain is still valid at the current time — a
    // chain valid at handshake may have since expired — before checking the
    // signature.
    Verify(ctx context.Context, payload, signature []byte, cert *VerifiedCertificate) error
}

// VerifiedCertificate is a chain that has passed ValidateChain. Its fields are
// unexported, so it can only be produced by ValidateChain — the compiler then
// guarantees Verify is never called with an unvalidated chain. It retains the
// chain, trust anchors, and hostname so validity can be re-confirmed at
// time-of-use via ValidAt.
type VerifiedCertificate struct { /* unexported */ }
func (c *VerifiedCertificate) Leaf() *x509.Certificate  // the validated leaf (read-only)
func (c *VerifiedCertificate) ValidAt(now time.Time) error

// TOFUStore persists the enrolled root CA across restarts.
type TOFUStore interface {
    Load() ([]byte, error)   // nil if no anchor stored yet
    Save(pem []byte) error   // idempotent — O_EXCL write, subsequent saves are no-ops
}
```

### Provided implementations

| Type | Usage |
|---|---|
| `signing.LocalSigner` | In-process signing from a `crypto.Signer` (file-loaded key) |
| `signing.LocalVerifier` | In-process verification against an `*x509.CertPool` |
| `signing.RemoteSigner` | Delegates `Sign` / `TrustAnchorPEM` to an HTTP policy server |
| `signing.FileTOFUStore` | Persists the enrolled root CA to a file |

### Constructor / provider helpers

```go
signing.LocalSignerFromFiles(keyPath, chainPath string) (*LocalSigner, error)
signing.VerifierFromFile(caPath string) (*LocalVerifier, error)
signing.VerifierFromPEM(pemBytes []byte) (*LocalVerifier, error)
signing.NewFileTOFUStore(path string) *FileTOFUStore
signing.NewRemoteSigner(baseURL string) *RemoteSigner

// Client-side PayloadTrustProvider constructors:
signing.FixedAnchor(v Verifier) PayloadTrustProvider      // fixed, pre-provisioned anchor
signing.TOFUAnchor(store TOFUStore) PayloadTrustProvider  // Trust On First Use enrollment
```

---

## Callbacks

No new callbacks are introduced. Existing callbacks now receive attestation-related errors through the existing surfaces:

```go
// ConnectionCallbacks (set inside OnConnecting)

// OnReadMessageError fires when the server cannot read or deserialise
// an inbound message. Now also covers oversized-message rejections.
OnReadMessageError func(conn Connection, mt int, msgByte []byte, err error)

// OnMessageResponseError fires when the server fails to send a response.
// Now also fires when a response exceeds MaxMessageSize.
OnMessageResponseError func(conn Connection, message *protobufs.ServerToAgent, err error)
```

On the client side, any attestation failure (missing trust chain, chain validation failure, SAN mismatch, bad signature) terminates the connection immediately. The WS client applies exponential backoff before reconnecting. The HTTP polling client resets attestation state and applies the same backoff before the next poll.

---

## Signing certificate rotation

The server's signing chain may rotate over the lifetime of a long-lived connection — the root (the agent's trust anchor) never changes, but the chain to it does. The server obtains the chain from each `Signer.Sign` call (returned alongside the signature in `SignResult`, so the two are always consistent) and re-sends `trust_chain_response` whenever the chain changes; the client re-validates the new chain against its pinned trust anchor and re-pins the new leaf, so rotation happens **without dropping the connection**. An unchanged chain is a no-op (the client compares against the pinned chain and skips re-validation) — this matters for the HTTP transport, which re-sends the chain on every poll.

`RemoteSigner` caches the fetched chain internally (`SetChainCacheTTL`, default 60s) so signing does not incur an HTTP chain fetch on every message — caching is purely an implementation detail of the signer, not part of the `Signer` interface.

---

## New error sentinel (server)

```go
// ErrSendBeforeNegotiated is returned from Connection.Send when called
// before the first AgentToServer is received (agent capabilities unknown).
// Push responses from OnMessage callbacks, not OnConnected.
var ErrSendBeforeNegotiated = errors.New("...")
```

---

## Capability bits (auto-managed)

Operators do not set these manually — they are derived from configuration:

| Bit | Auto-set by |
|---|---|
| `AgentCapabilities_RequiresPayloadTrustVerification` | Client, when `PayloadTrustProvider` is set |
| `AgentCapabilities_AcceptsPayloadTrustAnchorTOFU` | Client, when the provider is a `TOFUAnchor` with no persisted anchor yet |
| `ServerCapabilities_OffersPayloadTrustVerification` | Server, when `PayloadSigner != nil` |

---

## Minimal wiring example

**Server:**
```go
signer, _ := signing.LocalSignerFromFiles("server.key", "server-chain.pem")
srv := server.New(logger)
srv.Start(server.StartSettings{
    Settings: server.Settings{
        PayloadSigner: signer,
        Callbacks:     ...,
    },
})
```

**Agent (pre-provisioned CA):**
```go
verifier, _ := signing.VerifierFromFile("ca.pem")
client := client.NewWebSocket(logger)
client.Start(context.Background(), types.StartSettings{
    PayloadTrustProvider: signing.FixedAnchor(verifier),
    ...
})
```

**Agent (TOFU enrollment):**
```go
client.Start(context.Background(), types.StartSettings{
    PayloadTrustProvider: signing.TOFUAnchor(
        signing.NewFileTOFUStore("/var/lib/opamp/trust-anchor.pem"),
    ),
    ...
})
```


client/types/startsettings.go
```diff
@@ -69,6 +58,44 @@ type StartSettings struct {
...
 
+	// PayloadVerifier validates the X.509 trust chain delivered in the
+	// initial SignedServerToAgent.trust_chain_response of a connection
+	// and verifies the detached signature on every subsequent
+	// ServerToAgent message. MUST be set when the Agent's capability
+	// set includes
+	// AgentCapabilities_RequiresPayloadTrustVerification. When nil
+	// (the default), payload trust verification is disabled and the
+	// Server-to-Agent wire format is the standard ServerToAgent
+	// protobuf — identical to upstream OpAMP.
```
tigrannajaryan: Please make sure to remove references to "upstream" and other peculiarities that may be present because the work was initially done on a fork. This is the upstream.

client/types/startsettings.go
```diff
@@ -69,6 +58,44 @@ type StartSettings struct {
...
+	//     persists it via PayloadTOFUStore.Save().
+	//
+	// WARNING: TOFU provides no security on the first connection; a
+	// compromised distribution server can install an attacker-controlled
+	// trust anchor. Disable by default and enable only for environments
+	// where the first connection is considered sufficiently trusted.
+	// Requires persistent storage across restarts — agents running in
+	// stateless container environments without a persistent volume will
+	// repeat TOFU enrollment on every restart.
+	PayloadTOFUStore signing.TOFUStore
```
tigrannajaryan: My preference would be to have one interface that controls all attestation flows, whether it is with or without TOFU. It would hopefully allow adding other attestation modes in the future and extend it by adding optional methods to the interface without changing StartSettings.

Do you think it is possible refactor these settings like that?

server/serverimpl.go
```diff
@@ -434,8 +464,45 @@ func (s *server) handlePlainHTTPRequest(req *http.Request, w http.ResponseWriter
...
+	// with no persistent connection, so the trust handshake happens
+	// per-response: every signed response carries the chain alongside
+	// the signature. The Agent's HTTP receive path is stateful across
+	// polls (see client/internal/attestation.go) but tolerates the
+	// chain being re-sent — it just ignores it after the first.
+	//
+	// TODO(perf): for RPC-backed signers, newConnectionSigningState
+	// re-fetches the chain on every request — at 10⁶ agents polling
+	// every 30s that's ~33k RPS just for ChainDER. A server-level
+	// cache (or a TTL-aware Signer wrapper) would amortise the cost.
```
tigrannajaryan: Presumably ChainDER() can implement the cache. Is this something we need to care about here? If the signer is RPC-backed, it can still be wrapped in a locally caching implementation and only make an RPC call on cache expiry. Is there something I am missing that requires this note here and needs more work before we commit to a particular `Signer` interface?

signing/algorithm.go
```diff
@@ -0,0 +1,169 @@
+package signing
```
tigrannajaryan, line 1: I don't feel qualified to verify these bits. We need to find someone else who can review the algorithm and certs.

tigrannajaryan (commented):
I took a first pass and this looks quite reasonable to me.

My only important comment is about what settings we add to StartSettings. I also need help with reviewing the algorithm and certs parts.

We obviously need to have full test coverage for the added functionality.

tigrannajaryan (commented):
I took a first pass and this looks quite reasonable to me.

My only important comment is about what settings we add to StartSettings. I also need help with reviewing the algorithm and certs parts.

We obviously need to have full test coverage for the added functionality.

client/types/startsettings.go
liustanley: Agree that this refactor makes sense, I took a first pass at this but feel free to let me know if you have opinions on the interface names: https://github.com/open-telemetry/opamp-go/pull/599/commits/c08fd783594bdb2af0ab118359c64541df9fc183

```
type PayloadTrustProvider interface {
    Verifier() (Verifier, error)   // (nil,nil) -> no anchor yet -> TOFU
}
```

Two constructors cover the current modes:
- `signing.FixedAnchor(v)`: fixed, pre-configured anchor
- `signing.TOFUAnchor(store)`: TOFU enrollment

tigrannajaryan:
The public API looks good to me. 

@open-telemetry/opamp-go-approvers any thoughts?

server/serverimpl.go
liustanley: No I don't think this note is needed here and is a bit misleading, I'll remove it. Caching was never an interface concern (`RemoteSigner` already caches internally).

Separately, I'm collapsing `Sign` + `ChainDER` into a single `Sign` returning a small `SignResult{Payload, Signature, ChainDER}`, so the per-message chain fetch is gone entirely and the chain provably anchors the cert that just signed.

signing/chain.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	leaf := certs[len(certs)-1]
+
+	intermediates := x509.NewCertPool()
+	for i := 0; i < len(certs)-1; i++ {
+		intermediates.AddCert(certs[i])
+	}
+
+	opts := x509.VerifyOptions{
```
domodwyer, line 152: I would highly recommend setting the `DNSName` field for this call to verify the certificates are valid for the hostname we are connecting to, otherwise an attacker can present a valid certificate for `attacker.owned.domain.com` (which they can easily obtain) and this function will accept it for `opamp.bigcorp.com`.

I see that the [client does run `cert.VerifyHostname()`](https://github.com/open-telemetry/opamp-go/pull/599/changes#diff-829e9d3881744dd5e783705ad7d6651f6accb146ecfa60e13f013c8d02775f00R200) which is functionally the same as setting `DNSName`, but it's more fragile - it has to be implemented in every caller, spreading the validation logic across multiple blocks of independent code (this function also doesn't say that the caller MUST verify the hostname independently!).

signing/local_signer.go
```diff
@@ -0,0 +1,129 @@
...
+		// cert.Raw is the DER bytes the certificate was parsed from
+		// (or that x509.CreateCertificate produced). Copy to defend
+		// against later mutation of cert.Raw by callers, even though
+		// it's expected to be immutable in practice.
+		raw := make([]byte, len(cert.Raw))
+		copy(raw, cert.Raw)
+		chainDER[i] = raw
+	}
+
+	return &LocalSigner{
```
domodwyer, line 100: It might be good to have the `NewLocalSigner()` also verify the chain it is using prior to returning, so that errors with the chain are caught at startup time instead of manifesting as client connection errors later.

signing/local_signer.go
```diff
@@ -0,0 +1,129 @@
...
+	sig, err := signWithKey(s.key, s.alg, payload)
+	if err != nil {
+		return SignResult{}, err
+	}
+	// Return a shallow copy of the chain slice so callers cannot reorder or
+	// replace the signer's internal slice elements. The DER backing arrays
+	// are immutable after construction and are shared to keep Sign cheap on
+	// the hot path.
+	chain := make([][]byte, len(s.chainDER))
+	copy(chain, s.chainDER)
```
domodwyer: I'm not sure how valuable this is - it doesn't provide immutability of the underlying chain DER, as the caller can modify the backing array content.

In either case, this function should probably say "the caller must not modify the returned chain bytes" in the docs for absolute clarity :+1:

signing/local_signer.go
```diff
@@ -0,0 +1,129 @@
...
+	return s
+}
+
+// TrustAnchorPEM implements [TrustAnchorProvider]. Returns the PEM-encoded
+// root CA set by [WithRootCA]. Returns an error if WithRootCA was not called.
+func (s *LocalSigner) TrustAnchorPEM(_ context.Context) ([]byte, error) {
+	if len(s.rootCADER) == 0 {
+		return nil, errors.New("signing: no root CA configured on LocalSigner (call WithRootCA first)")
+	}
+	return pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: s.rootCADER}), nil
```
domodwyer: I dunno how frequently this will be called, but you could encode this in the setter (`WithRootCA()`) and avoid having to encode it for each call here - the DER appears to be unused otherwise :+1:

signing/types.go
```diff
@@ -0,0 +1,144 @@
...
+	// it to Verify on every subsequent message.
+	ValidateChain(ctx context.Context, chainDER [][]byte, now time.Time) (*x509.Certificate, error)
+
+	// Verify validates signature over payload using the public key of
+	// leaf. The signature algorithm is derived from leaf's public-key
+	// type and (for ECDSA) curve, cross-checked against
+	// leaf.SignatureAlgorithm. The payload bytes are the wire bytes of
+	// SignedServerToAgent.payload — the receiver does not re-marshal
+	// anything.
+	Verify(ctx context.Context, payload, signature []byte, leaf *x509.Certificate) error
```
domodwyer: It might be worth being very explicit here:

> The correctness of this call requires `leaf` to have been previously validated via [`ValidateChain()`].


I would consider adding a type wrapper over the leaf certificate here, so that you can statically guarantee at the type / compiler level that the certificate was verified prior to being passed to `Verify()`, making it impossible to provide a certificate that has not gone through `ValidateChain()`.

That would look like:


```suggestion
	// Returns the validated leaf certificate on success. The Agent
	// stores the leaf for the duration of the connection and passes
	// it to Verify on every subsequent message.
	ValidateChain(ctx context.Context, chainDER [][]byte, now time.Time) (*VerifiedCertificate, error)

	// Verify validates signature over payload using the public key of
	// leaf. The signature algorithm is derived from leaf's public-key
	// type and (for ECDSA) curve, cross-checked against
	// leaf.SignatureAlgorithm. The payload bytes are the wire bytes of
	// SignedServerToAgent.payload — the receiver does not re-marshal
	// anything.
	Verify(ctx context.Context, payload, signature []byte, leaf *VerifiedCertificate) error
```

Where `VerifiedCertificate` is:

```go
type VerifiedCertificate struct {
    inner *x509.Certificate
}
```

and private to this package.

signing/types.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// Verifier validates a delivered trust chain and verifies detached
+// signatures against the resulting leaf certificate.
+//
+// Implementations are expected to perform RFC 5280 §6 X.509 path
+// validation in ValidateChain. The Verify method performs the
+// signature-only check against the leaf returned by a successful
+// ValidateChain call.
+type Verifier interface {
```
domodwyer, line 127: I'm not sure about splitting chain verification and payload validation into two separate steps.

In general you cannot verify a certificate to establish it is trusted, and then reuse that certificate as trusted forever because they expire! As is, I believe `Verify()` would accept an expired leaf certificate (or expired intermediate / root).

It might be worth [writing a quick benchmark](https://pkg.go.dev/testing#hdr-Benchmarks) to measure how long a `ValidateChain()` call takes, and if it is acceptably fast, changing this `Verifier` such that it runs chain validation for every call to `Verify()` instead. 

This would guarantee the chain is valid at time-of-use, relying only on the well-vetted stdlib code.

Alternatively if validation is too slow, we could track the closest expiration time of any cert in the chain, and include that in the `VerifiedCertificate` to check against, but this doesn't account for other checks that can change state through time (revocation).

signing/remote_signer.go
```diff
@@ -0,0 +1,189 @@
...
+		return nil, fmt.Errorf("remote signer: CA returned HTTP %d: %s", resp.StatusCode, body)
+	}
+	return body, nil
+}
+
+// chainDER returns the current signing chain by GET-ing /v1/chain and decoding
+// the returned PEM blob into DER byte slices ordered intermediates-first,
+// leaf-last. It is called by Sign on every outbound message; results are cached
+// for chainTTL to bound the fetch rate.
+func (s *RemoteSigner) chainDER(ctx context.Context) ([][]byte, error) {
```
domodwyer, line 160: Presumably we're expecting concurrent calls to `Sign()`, which would cause concurrent calls to `chainDER()` and therefore hit the backend with as many requests for a chain update request.

It might be better to have the first caller perform the HTTP request, and all other concurrent callers wait for it to complete - this'll save from a burst on the backend signer / resource usage to handle the response in this process.

domodwyer (commented):
_note: I have only reviewed the `signing` package._

In general this looks good! I think the [this comment](https://github.com/open-telemetry/opamp-go/pull/599/changes#r3712600197) about how the `Verifier` handles cert expiry, and [setting the `DNSName` when verifying the chain](https://github.com/open-telemetry/opamp-go/pull/599/changes#r3712258475) are my only blocking concerns.

However we should absolutely add at least some integration tests! Ideally we should verify that static fixture certs can be loaded and used to verify a previously signed payload in order to ensure no regressions are introduced now or in the future. Having test helper functions to generate test certs is awesome, but they're easily updated alongside the code, so a solid fixture test is really helpful to close this kind of gap 👍 

It would be great to have a bunch of tests in general, but I think we should at least cover:

* A "successful validation" integration test as above.
* A negative case "failed validation" integration test to show that a modified payload is rejected.
* An expired certificate is rejected (as currently wrote, we'd need two tests: one when the cert if first presented to the client, one that simulates an expiration after that has occurred).
* A code signing certificate for `itsallbroken.com` isn't accepted for `opamp.bigcorp.com` (hostname verification is enforced).

signing/local_signer.go
liustanley: That's true, I'll keep this copy since it still stops element reassignment and add documentation to clarify that these fields are read-only

signing/types.go
liustanley: That makes sense with me, I added `VerifiedCertificate` with some additional fields that are also needed in `ValidateChain` in the suggestion below
```
type VerifiedCertificate struct {
	leaf    *x509.Certificate
	chain   []*x509.Certificate // ordered intermediates first, leaf last
	roots   *x509.CertPool
	dnsName string
}
```

signing/types.go
liustanley: Good catch, I tested this with a benchmark and found that re-validation logic from `ValidateChain()` takes ~107µs on a realistic 3-cert chain (2 intermediates + leaf). This seems acceptably fast so I added chain validation for every call to `Verify()` through `VerifiedCertificate.ValidAt()`.

signing/types.go
```diff
@@ -0,0 +1,154 @@
...
+	ChainDER [][]byte
+}
+
+// Signer produces a detached signature over payload bytes together with the
+// certificate chain that anchors it (see [SignResult]).
+//
+// Implementations may sign in-process (see [LocalSigner]) or delegate to an
+// external service (HSM, remote signing RPC). Sign takes a context for
+// cancellation, deadlines, and trace propagation.
+type Signer interface {
```
pierreprinetti, line 97: I wonder if we could directly use stdlib's [`crypto#Signer`](https://pkg.go.dev/crypto#Signer) interface instead:

```go
type Signer interface {
        // Public returns the public key corresponding to the opaque,
        // private key.
        Public() PublicKey

        // Sign signs digest with the private key, possibly using entropy from
        // rand. For an RSA key, the resulting signature should be either a
        // PKCS #1 v1.5 or PSS signature (as indicated by opts). For an (EC)DSA
        // key, it should be a DER-serialised, ASN.1 signature structure.
        //
        // Hash implements the SignerOpts interface and, in most cases, one can
        // simply pass in the hash function used as opts. Sign may also attempt
        // to type assert opts to other types in order to obtain algorithm
        // specific values. See the documentation in each package for details.
        //
        // Note that when a signature of a hash of a larger message is needed,
        // the caller is responsible for hashing the larger message and passing
        // the hash (as digest) and the hash function (as opts) to Sign.
        Sign(rand io.Reader, digest []byte, opts SignerOpts) (signature []byte, err error)
}
```

In your implementation you return a `SignResult` to return the signed payload to the caller. The rationale is: you can't be sure that the signing implementation actually signed the bytes you passed. However **the signer should guarantee that the signature corresponds to the passed payload**, byte for byte. Without this guarantee, you leave it to the caller to verify the semantic equivalence of the returned payload, which opens the door to a class of nasty errors.

Your `SignResult`'s `ChainDER`, when using the stdlib's `Signer`, would be replaced by `Public() [PublicKey](https://pkg.go.dev/crypto#PublicKey)', leaving the caller type-switch it and encode it to the desired wire representation (DER?) if needed.

What do you think? My take is that conforming to the stdlib's interfaces helps you avoid common pitfalls, on top of leveraging our Go developer friends' muscle memory.

liustanley: Thanks for the suggestion. I agree that conforming to the stdlib's interfaces is good in theory, and we do use the standard `crypto.Signer` internally in the simple local signer case where it fits, but it might result in some gaps in the message attestation implementation for remote signers.

> the signer should guarantee that the signature corresponds to the passed payload, byte for byte.

Agreed on the goal, and that's exactly why `SignResult` returns the signed bytes. As I understand it, the main issue with the signer in this scenario is that a remote signing service could be implemented to receive a protobuf message (i.e. `ServerToAgent`) instead of a payload in raw bytes. Two systems serializing the same protobuf message can result in different output bytes, since [protobuf is not canonical](https://protobuf.dev/programming-guides/serialization-not-canonical/). In this case, the only way to guarantee that the signature matches up with the payload is to use the exact bytes that the signing service signed.

> Your SignResult's ChainDER, when using the stdlib's Signer, would be replaced by `Public() [PublicKey](https://pkg.go.dev/crypto#PublicKey)', leaving the caller type-switch it and encode it to the desired wire representation (DER?) if needed.

One problem we found with a separate chain accessor (and the reason why we folded `ChainDER()` into `SignResult`) is that having two separate calls for the signature and the chain can result in a race condition in the case that a rotation happens in between the two calls. We also need `context` and `error` in the chain function in order to do I/O that can fail with cancellation and deadlines, which isn't possible with the `Public()` signature.

pierreprinetti: Right, it makes sense that if there's a network roundtrip involved, the method calls are fallible and should carry context for cancellation. Also on a high level I understand the argument that returning the public key from a separate method may open the door to state mismatch.
I am still not convinced by not guaranteeing that the Sign method treats the bytes as an opaque message and actually signs exactly the bytes it's given. I'll have to dig a little bit more into the code to better understand both arguments. Thanks for the answers!

PR #638, opened 2026-08-20
Handle SIGTERM in example server

cijothomas:
## Description

The example server currently waits only for `os.Interrupt`. Container runtimes such as Docker send SIGTERM by default when stopping a container, so the process exits through the default signal behavior and skips `uisrv.Shutdown()` and `opampSrv.Stop()`.

Subscribe to SIGTERM alongside `os.Interrupt` so both interactive and container shutdown reach the existing cleanup path. This was found while integrating the example server into the OpenTelemetry Demo.

## Testing

- `cd internal/examples && go test ./server/...`
- `cd internal/examples && go build -o /tmp/opamp-server ./server`

### DIFF
--- a/internal/examples/server/main.go
+++ b/internal/examples/server/main.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"os"
 	"os/signal"
+	"syscall"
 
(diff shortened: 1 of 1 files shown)


cijothomas:
@tigrannajaryan Could you review merge?

